PR #42, opened 2026-03-18
fix(app): show menu bar icon state for file processing and manual recording

pasrom:
## Problem
The menu bar icon stayed idle (`.inactive`) in two cases:
1. **File processing** via "Process Audio/Video Files..." — no `watchLoop` is created, so `currentBadge` never entered the watchLoop branch
2. **Manual recording** — `watchLoop` is created inside an async `Task`, so there was a brief window where `watchLoop` was set but `loop.isActive` was still false

## Fix
`currentBadge` now checks `pipelineQueue.activeJobs` independently of `watchLoop`, so the icon reflects the pipeline state regardless of how jobs were enqueued.

## Test plan
- [ ] Open a file via "Process Audio/Video Files..." → icon animates during transcribing/diarizing/protocol
- [ ] Start manual recording via "Record App..." → icon shows recording animation immediately

### DIFF
--- a/app/MeetingTranscriber/Sources/MeetingTranscriberApp.swift
+++ b/app/MeetingTranscriber/Sources/MeetingTranscriberApp.swift
@@ -69,11 +69,12 @@ struct MeetingTranscriberApp: App {
             } icon: {
                 Image(nsImage: MenuBarIcon.image(
                     badge: currentBadge,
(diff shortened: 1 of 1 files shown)


pasrom:
Superseded by #43 (squashed)

PR #100, opened 2026-04-11
test(app): add E2E full pipeline test with zero mocks

pasrom:
## Summary
- Full end-to-end pipeline test with **zero mocks**: real detection, real recording, real transcription (Parakeet), real diarization (FluidDiarizer)
- Meeting Simulator launches as subprocess, plays `two_speakers_de.wav` through speakers
- WatchLoop detects via PowerAssertionDetector, CATapDescription records the audio
- Parakeet transcribes (~1.6 KB German text with R_/M_ dual-track speaker labels), FluidDiarizer identifies 2 speakers
- Skipped on regular CI via `shouldSkipE2E` (CI && !E2E_ENABLED) — requires audio hardware + mic permission
- Runs on the self-hosted Mac mini runner via `e2e.yml` workflow (workflow_dispatch + tag push)
- `SimulatorHelper` in TestHelpers.swift: builds + launches meeting-simulator, with pipe-deadlock prevention
- ~67 s runtime locally on M-series; cold first run downloads Parakeet (~50 MB) + FluidAudio models

## Test plan
- [x] `E2E_ENABLED=1 swift test --filter E2EFullPipeline` passes locally (67 s, transcript 1595 chars)
- [x] Lint clean (`./scripts/lint.sh` — 0 errors, 0 serious)
- [x] PR-CI passes (E2E auto-skipped on github-hosted runners, all other tests unaffected)
- [ ] E2E workflow passes on self-hosted Mac mini runner (`gh workflow run e2e.yml --ref test/e2e-full-pipeline`)

### DIFF
--- a/app/MeetingTranscriber/Tests/E2EFullPipelineTests.swift
+++ b/app/MeetingTranscriber/Tests/E2EFullPipelineTests.swift
@@ -1,0 +1,113 @@
+@testable import MeetingTranscriber
+import XCTest
+
+/// End-to-end pipeline test with **zero mocks**.
+///
+/// Launches the meeting-simulator tool (which opens a window, creates a power assertion,
+/// and plays audio). WatchLoop detects the meeting via PowerAssertionDetector,
+/// DualSourceRecorder taps the simulator's audio via CATapDescription, Parakeet
+/// transcribes the recording, FluidDiarizer identifies speakers, and the test verifies
+/// that the transcript output file exists and contains recognized speech.
+///
+/// Skipped in CI — requires audio hardware, mic permission, and model downloads.
+/// Run locally: `swift test --filter E2EFullPipeline`
+@MainActor
+final class E2EFullPipelineTests: XCTestCase {
+    private var tmpDir: URL?
+
+    // MARK: - Setup
+
+    override func setUp() async throws {
+        try XCTSkipIf(
+            shouldSkipE2E,
+            "E2E test requires audio hardware — set E2E_ENABLED=1 on self-hosted runners",
+        )
+
+        let dir = FileManager.default.temporaryDirectory
+            .appendingPathComponent("e2e_pipeline_\(UUID().uuidString)")
+        try FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
+        tmpDir = dir
+    }
+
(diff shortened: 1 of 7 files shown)


pasrom:
Closing this PR — the test concept (full live recording → pipeline → assertion) is right, but the **xctest framework is the wrong frame** for it.

## Lessons learned from this PR

### What we tried
`E2EFullPipelineTests` spawned `meeting-simulator` as a subprocess and used the real `DualSourceRecorder` (CATapDescription + AVAudioEngine) inside an xctest test method. Worked locally on a logged-in dev Mac in 67 s, produced a clean ~1.6 KB transcript with R_/M_ speaker labels.

### What broke on the self-hosted Mac mini runner
The same test crashed reproducibly with `freed pointer was not the last allocation` (libmalloc nano-zone hardening) → SIGABRT in xctest. Three rounds of investigation pinned the cause:
- xctest is an ad-hoc-signed binary whose cdhash changes every build → never inherits stable TCC permissions
- meeting-simulator subprocess raised an `NSXPCSharedListener … ClientCallsAuxiliary Connection interrupted` warning — partial AppKit/XPC subsystem in headless service contexts
- The Mac mini has no built-in mic; `AVAudioEngine.inputNode` against a no-input device is in an unhappy edge-case path that triggers the libmalloc abort
- BlackHole-2ch as virtual mic input fixed the crash but the captured audio ended up silent (CATapDescription bound through xctest doesn't get the right system-audio routing context)

### What we discovered works
Deploying the actual built `.app` bundle to a stable path (`~/Applications/MeetingTranscriber-Dev.app`), granting permissions once, and triggering it via the existing DebugRPCServer **does work end-to-end** — detection fires, dual-source recording captures real audio (1.78 MB per track), Parakeet transcribes correctly. Diarization currently fails because the mic track is silent (BlackHole loopback isn't fed), but that's a configuration thing, not architecture.

### New approach
Pattern C: deploy the production `.app`, drive it via RPC, assert on output. Implementation in branch `feat/e2e-app-driver` (separate workflow `e2e-app.yml`, distinct from PR-blocking CI). CI for PRs stays Pattern A — fixture-based component tests in `WhisperKitE2ETests` / `ParakeetE2ETests` / `WatchLoopE2ETests` etc.

Test fixture (`two_speakers_de.wav`) and the `meeting-simulator` tool stay relevant and will be reused.

PR #183, opened 2026-05-07
fix(app): roll back partial record-only finalize

pasrom:
## Summary

Record-only mode finalises a recording by moving the mix/app/mic WAVs into the user's output dir and writing a `_meta.json` sidecar. When a move mid-flow throws (e.g. mic source vanished between recorder stop and our move), the already-moved WAVs were left in the output dir as orphans. The `do/catch` only logged. To a downstream consumer (e.g. Linux GPU pipeline reading the dir over Syncthing) those orphan WAVs were indistinguishable from a complete recording missing only its sidecar.

## Approach

Track every URL we materialise during finalize in a `[URL]`. On any thrown error in `do/catch`, roll them back. The sidecar is written last and atomically; without it, anything we already moved is provisional.

While here: `_app.wav` / `_mic.wav` literals get joined to the existing `mixFilenameSuffix` constant on `DualSourceRecorder` to avoid hardcoded copies of the same string.

## Why no startup orphan sweep

Initially this PR included a `cleanupOrphanedRecordOnlyOutputs` sweep at `WatchLoop.start()` that would delete `*_(mix|app|mic).wav` files without a matching `_meta.json` sidecar — to recover from app crashes mid-finalize. **Rejected after live testing exposed a destructive bug:** the `recordings/` directory is shared with normal (non-record-only) pipeline output, which `PipelineQueue.copyAudioToOutput` writes via `moveItem` to the same location. Normal pipeline recordings **never** write a sidecar (only record-only does). The sweep would therefore delete every legitimate non-record-only recording on the next launch, and conflict with `PipelineQueue.recoverOrphanedRecordings` which already treats orphan `_mix.wav` files as candidates for re-queuing.

The H7 finding's stated impact ("manual cleanup needed after crash mid-finalize") is acceptable. A proper crash-recovery design would require either a record-only-only subdir layout (breaks the consumer protocol) or a `.partial` rename pattern at finalize — both out of scope for this fix.

## Test plan
- [x] `swift test --parallel --filter WatchLoopTests` — 32 tests pass, including `test_recordOnly_partialMoveFailure_rollsBackOrphanedFiles`.
- [x] `swift test --parallel` — only pre-existing `MenuBarIconSnapshotTests` flake on local machine, unrelated.
- [x] `swift build -Xswiftc -DAPPSTORE` — App Store variant compiles.
- [x] `./scripts/lint.sh` — clean.

### DIFF
--- a/app/MeetingTranscriber/Sources/DualSourceRecorder.swift
+++ b/app/MeetingTranscriber/Sources/DualSourceRecorder.swift
@@ -49,6 +49,8 @@ class DualSourceRecorder: RecordingProvider {
     /// Suffix on the merged-output WAV file, used by downstream code (the
     /// record-only sidecar writer) to recover the recording basename.
     static let mixFilenameSuffix = "_mix.wav"
+    static let appFilenameSuffix = "_app.wav"
+    static let micFilenameSuffix = "_mic.wav"
 
     /// Remove leftover `*_app_raw.tmp` files from a previous crash.
     static func cleanupTempFiles(recordingsDir: URL = AppPaths.recordingsDir) {
@@ -84,7 +86,7 @@ class DualSourceRecorder: RecordingProvider {
 
         // ── AudioTapLib capture session ──
         let appTempURL = recDir.appendingPathComponent("\(ts)_app_raw.tmp")
-        let micURL: URL? = noMic ? nil : recDir.appendingPathComponent("\(ts)_mic.wav")
+        let micURL: URL? = noMic ? nil : recDir.appendingPathComponent("\(ts)\(Self.micFilenameSuffix)")
 
         let session = AudioCaptureSession(
             pid: appPID,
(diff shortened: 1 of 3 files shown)


pasrom:
Closing — H7 not pursued.

After live-testing exposed that the originally-shipped orphan-sweep destroyed legitimate non-record-only recordings (`recordings/` is shared with `PipelineQueue.copyAudioToOutput` which never writes a sidecar), the fix was reduced to just the error-path rollback. On reflection, even that is wrong:

1. Downstream Syncthing-style consumers may already have picked up the partial WAVs mid-flight; deleting them creates a "appeared then vanished" state worse than leaving them.
2. The H7 finding's stated impact ("user has to manually clean up") is already the production baseline. Keeping orphan WAVs is reversible (user deletes via Finder); destroying them is not.
3. Asymmetric recovery: WAV data is the precious bit, sidecar metadata is reconstructable by hand.

Tracking this as "accepted impact" in the side-effects review. A future fix would use a `.partial` marker file written before the move loop and atomically renamed on success — non-destructive, marks half-finalized recordings without removing them. Out of scope for now.

The suffix-constants cleanup (`DualSourceRecorder.appFilenameSuffix` / `micFilenameSuffix`) is small enough to ship separately if useful.

PR #205, opened 2026-05-08
feat(app): apply custom vocabulary as Whisper decoding prompt

pasrom:
## Summary
- Until now the custom-vocabulary file at `AppSettings.customVocabularyPath` was only consumed by `ParakeetEngine` (CTC boosting). When transcribing with WhisperKit, those terms had no effect — domain-specific names like "Klaus-Dieter" or "CATapDescription" were free to be misheard as generic words even though the user had configured them as project vocabulary.
- `WhisperKitEngine` now reads the same newline-separated file, joins non-empty terms into a `"Glossary: t1, t2, …"` prompt phrase, tokenises it via the loaded model's `WhisperTokenizer`, and feeds the result into `DecodingOptions.promptTokens`.
- No settings UI change — existing users with a configured vocab path get biasing automatically the next time they transcribe with WhisperKit.

## Implementation
- Three pure helpers on `WhisperKitEngine` (all `nonisolated static`): `parseVocabulary(from:)` (file → trimmed, non-empty terms), `makeVocabularyPrompt(terms:)` (terms → `"Glossary: …"`), `clampTokens(_:maxLength:)` (head-truncate to fit the 224-token budget).
- `refreshVocabularyTokensIfNeeded(tokenizer:)` runs once per `customVocabularyPath` change — caches the encoded tokens so repeated transcriptions don't re-tokenise.
- `cachedPromptTokens.isEmpty ? nil : cachedPromptTokens` is passed to `DecodingOptions.promptTokens`, preserving the `nil = no prompt` semantics WhisperKit expects.
- `AppState.syncLanguageSettings` propagates the path to the WhisperKit engine in addition to Parakeet.
- Token cap is 224, well under WhisperKit's `Constants.maxTokenContext = 448`, leaving room for the prefill prompt + audio.

## Tests
- **Unit (7):** `parseVocabulary` covers empty path / missing file / multi-line file with whitespace + blanks; `makeVocabularyPrompt` covers empty-terms → nil and terms → `"Glossary: <comma-joined>"`; `clampTokens` covers under-limit pass-through and over-limit head-truncation. Plus a smoke test that `WhisperKitEngine.customVocabularyPath` defaults to `""`.
- **End-to-end (1, skipped on CI):** `testTranscribeSegmentsAppliesCustomVocabularyAsPrompt` loads `openai_whisper-small`, sets a six-term vocab path, transcribes the German fixture, asserts `cachedPromptTokens` is non-empty, ≤ `maxPromptTokens`, and > 5 (six terms tokenise to a handful). Runs in ~6 s locally when the model is cached. `cachedPromptTokens` is `private(set)` to make this assertable.
- Full suite green, `swift build -c release` clean.

## Manual verification
Performed locally on `openai_whisper-small`: vocabulary file is parsed, tokenised through the live model, and the resulting prompt-token array stays under the budget after a real transcription run. Token count for the six-term glossary above lands well below the 224-token cap.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppState.swift
+++ b/app/MeetingTranscriber/Sources/AppState.swift
@@ -425,6 +425,7 @@ final class AppState {
     private func syncLanguageSettings() {
         if settings.transcriptionEngine == .whisperKit {
             whisperKit.language = settings.whisperLanguageOrNil
+            whisperKit.customVocabularyPath = settings.customVocabularyPath
         }
         if settings.transcriptionEngine == .parakeet {
             parakeetEngine.customVocabularyPath = settings.customVocabularyPath
--- a/app/MeetingTranscriber/Sources/WhisperKitEngine.swift
+++ b/app/MeetingTranscriber/Sources/WhisperKitEngine.swift
@@ -55,7 +55,7 @@ final class WhisperKitEngine: TranscribingEngine {
     /// Cached prompt tokens for `DecodingOptions.promptTokens`. Refreshed
     /// when `customVocabularyPath` changes since last transcription. Empty
     /// means "no biasing" — translated to `nil` at the `DecodingOptions` call.
-    private var cachedPromptTokens: [Int] = []
+    private(set) var cachedPromptTokens: [Int] = []
     private var lastTokenizedVocabularyPath: String = ""
 
     /// Whisper's prompt context is bounded; 224 leaves headroom under the
--- a/app/MeetingTranscriber/Sources/WhisperKitEngine.swift
+++ b/app/MeetingTranscriber/Sources/WhisperKitEngine.swift
@@ -42,12 +42,25 @@ extension TimestampedSegment {
 final class WhisperKitEngine: TranscribingEngine {
     var modelVariant = "openai_whisper-large-v3-v20240930_turbo"
     var language: String?
+    /// Path to a newline-separated vocabulary file. Empty disables biasing.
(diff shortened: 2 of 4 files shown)


pasrom:
Closing — biasing breaks WhisperKit's decoding on the production model.

Smoketest against `large-v3-v20240930_turbo` on real meeting audio:
- `vocabulary loaded terms=6 tokens=30`
- `Decoding Temperature: 0.0` initial decode
- `Fallback #1 (firstTokenLogProbThreshold)` ← quality fallback fires
- `Decoding Temperature: 0.2` retry with sampling
- `transcription_complete segments=0`

The same audio sources transcribed correctly (`segments=4–6`) in the same
session before biasing was active. Confirmed: the prompt drives the
first-token logprob below WhisperKit's default `-1.5` threshold, and the
fallback's temperature-0.2 sampling on the prefilled state still produces
no usable output.

The unit-test E2E (`testTranscribeSegmentsAppliesCustomVocabularyAsPrompt`)
ran against `whisper-small` on the clean `two_speakers_de.wav` fixture
and passed — the small model is too permissive to surface the threshold
trip. That gap on the test side is also captured.

Properly shipping this needs an opt-in toggle, threshold tuning when
biasing is active (or empty-result retry-without-prompt), prompt-language
recommendations matching audio language, and a WER regression test
against the production model. Out of scope for "1 Nachmittag" — deferred.

Branch `feat/whisperkit-contextual-biasing` keeps the static helpers
(`parseVocabulary`, `makeVocabularyPrompt`, `clampTokens`) as a starting
point if we revisit.

PR #235, opened 2026-05-11
refactor(app): move openAIAPIKey from Keychain to file-based storage

pasrom:
## Summary

`AppSettings.openAIAPIKey` is the only remaining caller of `KeychainHelper`. Migrates it to a 0600 file at `AppPaths.dataDir/.openai-key` — same pattern the `DebugRPCServer` `.rpc-token` already uses.

Deletes `KeychainHelperTests.swift` and the two `AppSettingsTests` methods that exercised the direct-Keychain path. `KeychainHelper.swift` itself stays as a legacy migration helper (one-shot read on first launch after upgrade), explicitly commented so nobody adds new callers.

## Why now

xctest spawned by the runner's user-level launchd service on the Mac mini gets a non-interactive context where `SecItemAdd` / `SecItemUpdate` on the persistent `login.keychain` silently return `errSecInteractionNotAllowed`. Today's `ci/sanitizer-to-mini` experiment (PR #234) added per-job test-keychain provisioning as a workaround, but it hit per-user state races when both TSan and ASan ran in parallel on the two Mini runners — each job's `security default-keychain -s` clobbered the other's view, because `runner` is the same OS user on both.

File-based storage is per-process / per-path and races nothing. After this lands, PR #234 can drop the test-keychain provisioning steps entirely and the sanitizer matrix on Mini becomes a clean win.

## Risk

The Keychain stored a Bearer token kept locally for an OpenAI-compatible HTTP endpoint. The file stores the same Bearer token. Both rely on the user's OS account for protection; FileVault encrypts at rest. An attacker who can read `~/Library/Application Support/MeetingTranscriber/.openai-key` (mode 0600) could also unlock the user's Keychain. Net risk: no meaningful change.

## Migration

First launch after upgrade reads the legacy Keychain entry once via the kept `KeychainHelper.read`, writes it to the file, then deletes the Keychain entry. Runs successfully in production signed builds; silent no-op anywhere Keychain isn't accessible (dev builds + self-hosted CI). Users with an existing OpenAI key keep it without re-entering.

## Test plan

- [x] Local `swift test --filter AppSettingsTests --parallel` — green.
- [x] Local `./scripts/lint.sh` — clean (after one auto-fix pass for `wrapPropertyBodies` + `blankLinesAtEndOfScope`).
- [ ] CI green on the PR.
- [ ] After merge: rebase #234 onto main, remove the test-keychain provisioning steps from `quality-and-safety.yml`, re-trigger the sanitizer matrix on Mini, expect both legs green.

---
<a href="https://app.blacksmith.sh/pasrom/codesmith/meeting-transcriber/pr/235"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews


pasrom:
Reconsidered — keeping Keychain in production preserves the per-app code-signing ACL protection (Threat 2: other apps running as the same user can't read the secret). Going with a serialized sanitizer matrix in PR #234 instead, so the per-job ephemeral test keychain works without per-user state races between parallel matrix legs.

PR #308, opened 2026-05-20
fix(app): detect zero-signal mic callbacks as broken microphone

pasrom:
## Summary

- Cherry-picks the zero-buffer mic detection fix from `leibniztheoctopus/meeting-transcriber@498db95` with review applied on top, plus regression tests.
- Closes the gap where macOS delivers live audio callbacks filled with zeros (unplugged USB mic, BlackHole misrouted, post-sleep driver glitch) and the existing probe greenlit it as healthy.
- The probe runs proactively on app launch and on `NSApplication.didBecomeActiveNotification` — so the red `!` badge now surfaces *before* the user starts a recording, complementing the reactive `ChannelHealthMonitor` that catches mid-recording drift.

## What changed

**Commit 1 — `fix(app):` (Leibniz, co-authored)**
- `inputFormat(forBus:)` instead of `outputFormat(forBus:)` — hardware-authoritative
- `BufferCounter` now tracks `maxPeak` alongside `count`
- Probe returns `.broken` when peak stays below threshold even though buffers are flowing
- Review additions:
  - Extracted `peakAmplitude(of:)` as a pure static helper (testable; folds Float32/Int16 duplication)
  - Unknown sample formats fall back to `peak = 1.0` (preserves pre-cherry-pick "any buffer = live" semantics)
  - Named constant `silentMicPeakThreshold` (~-80 dBFS) replaces the magic `0.0001` literal
  - **Race fix**: polling loop previously exited on the first buffer (often a warm-up zero-buffer from CoreAudio) and returned `.broken`. Now polls until `peak > threshold` or deadline.

**Commit 2 — `test(app):`**
- 8 unit tests covering `peakAmplitude(of:)`: Float32 (zero / full-scale / below-threshold / exact-threshold), Int16 (zero / full-scale), empty buffer, and Int32 unknown-format fallback. Strict `>` comparison is pinned against accidental `>=` drift.

## Test plan

- [x] `swift build` clean
- [x] `swift test --filter PermissionHealthCheckTests` — 42/42 pass (34 existing + 8 new)
- [x] `./scripts/lint.sh` — 0 violations
- [ ] Manual smoke: unplug USB mic, launch app, expect red `!` badge before recording
- [ ] CI: e2e + e2e-app + sanitizer matrix

## Notes

- `probeMicrophone()` itself remains hardware-bound and is not directly unit-tested. The race-fix is structurally trivial (loop condition `count > 0` → `maxPeak > threshold`) and correctness now hinges on `peakAmplitude(of:)`, which is covered.
- SwiftFormat's `preferIsEmpty` and SwiftLint's `empty_count` both misfire on the `(count: Int, maxPeak: Float)` tuple field — suppressed inline with disable comments.

### DIFF
--- a/app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
+++ b/app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
@@ -171,22 +171,59 @@ enum PermissionHealthCheck {
         }
     }
 
-    /// Thread-safe counter for mic probe buffer arrival (tap callback runs on audio thread).
+    /// Peak amplitude (linear, 0…1) below which a mic stream is treated as silence even
+    /// if buffers are flowing — a working mic in any real environment exceeds the
+    /// noise-floor by orders of magnitude. ~−80 dBFS.
+    static let silentMicPeakThreshold: Float = 0.0001
+
+    /// Returns the maximum absolute sample amplitude in `buffer` on channel 0, normalized
+    /// to 0…1. Unknown sample formats fall back to `1.0` so they pass the silence check —
+    /// preserves pre-cherry-pick behavior (any buffer = live).
+    static func peakAmplitude(of buffer: AVAudioPCMBuffer) -> Float {
+        let frames = Int(buffer.frameLength)
+        guard frames > 0 else { return 0 }
+        if let channelData = buffer.floatChannelData {
+            var peak: Float = 0
+            for sample in UnsafeBufferPointer(start: channelData[0], count: frames) {
+                let abs = Swift.abs(sample)
+                if abs > peak { peak = abs }
+            }
+            return peak
+        }
+        if let channelData = buffer.int16ChannelData {
+            var peak: Float = 0
+            let scale = 1.0 / Float(Int16.max)
+            for sample in UnsafeBufferPointer(start: channelData[0], count: frames) {
+                let abs = Swift.abs(Float(sample)) * scale
+                if abs > peak { peak = abs }
+            }
+            return peak
+        }
+        return 1.0
+    }
+
+    /// Thread-safe mic probe stats (tap callback runs on audio thread).
     private final class BufferCounter: @unchecked Sendable {
         private let lock = NSLock()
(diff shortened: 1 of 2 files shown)


pasrom:
## Coverage rationale

Patch coverage is 41.46% (24 uncovered lines) — that's expected for this PR and not a quality gap. Breakdown of what's covered vs not:

**Covered (`peakAmplitude(of:)` — pure helper):**
- Float32 / Int16 / empty / Int32-unknown-format / exact-threshold edge → 100% by 8 unit tests
- The threshold comparison operator (`>` vs `>=`) is pinned by `testPeakAmplitudeAtExactThresholdIsRejected`

**Not covered (uncoverable without audio-hardware mock framework):**
- `probeMicrophone()` body — `AVAudioEngine` setup + `installTap` + `engine.start()` + polling loop + cleanup. Needs a real input device + TCC + AVCaptureDevice authorization. This is the exact pattern documented in [`xctest-coverage-no-go-zones`](docs/plans/.local/research/) — AVAudioEngine paths are an established no-go zone (alongside NSOpenPanel / NSWorkspace / TCC probes / Socket binds).
- `BufferCounter` (`record` + `snapshot`) — private nested helper. Only invoked from inside the tap closure during a live probe. Stays `private` deliberately; exposing it as `internal` to inflate coverage would be the "tests-only API" anti-pattern that [`feedback-production-bugs-need-injection-tests`](#) explicitly warns against — the injection seam is at the *pure-helper* level (`peakAmplitude`), not at the audio-pipeline-internals level.

**Race-fix correctness traceability:** the polling-loop fix (`count > 0` → `maxPeak > threshold`) is structurally trivial once `peakAmplitude` is correct, and the threshold-edge test guards against `> ` drifting to `>=`. The remaining hardware-engine surface gets exercised by the live `e2e-app.yml` workflow on the self-hosted runner.

Net effect on project coverage: -0.1% (75.4% → 75.3%). The new code is 41% covered, but the *coverable surface* (the pure helper) is 100% covered.

PR #327, opened 2026-05-23
feat(app): show configured mic speaker name in live caption overlay

pasrom:
## Summary

Slice 1 of the live-speaker-labels follow-up to PR #318. Replaces the hard-coded mic prefix `Du` in the live caption overlay with `settings.micName`, so the bar reads e.g. `Roman:` instead of `Du:`. App-channel default stays `Remote` — slice 2 (known-speaker matching on the app track) will replace that.

- New per-line `LiveCaptionLine.speaker` field on the RPC wire format, pinned in `LiveCaptionWireFormatTests` so a future rename fails locally before it bricks `mt-cli` / `e2e-live-captions.sh`.
- Label source moves from `LiveCaptionChannel.label` (compile-time enum) into `LiveCaptionsState.micLabel` / `.appLabel` (Observable, settings-driven).
- `LiveTranscriptionController` reads the label at every final, so a rename takes effect on the next final without rebuilding the controller — already-committed finals keep their per-line snapshot.
- Privacy: `StreamingTranscriber`'s `.public` log prefix uses the raw channel id (`mic` / `app`) instead of the user-visible name, so the unified log never sees the display name.

## Test plan

- [x] `swift build` (debug + release)
- [x] `swift test --parallel` — full suite green (1791 tests)
- [x] `./scripts/lint.sh` — clean (only pre-existing untracked Sortformer A/B-experiment violations, not part of this PR)
- [x] `LiveTranscriptionE2ETests` asserts mic finals carry `Roman`, app finals stay `Remote`
- [x] `AppStateTests/testLiveCaptionsMicLabelTracksSettingsRename` pins the live rename path
- [ ] CI green

### DIFF
--- a/app/MeetingTranscriber/Sources/AppState.swift
+++ b/app/MeetingTranscriber/Sources/AppState.swift
@@ -90,8 +90,11 @@ final class AppState { // swiftlint:disable:this type_body_length
 
     /// Observable state for the live caption overlay. Always present (the
     /// `LiveCaptionsOverlay` window observes this); content is only populated
-    /// when live transcription is on AND a recording is active.
-    let liveCaptions: LiveCaptionsState = .init()
+    /// when live transcription is on AND a recording is active. Constructed
+    /// in `init` so the mic label reflects `settings.micName` from the
+    /// first overlay render. Kept in sync with later settings changes via
+    /// `setupLiveTranscriptionPrewarm`.
+    let liveCaptions: LiveCaptionsState
 
     #if !APPSTORE
         /// Lazy-started debug RPC server. Only constructed if the env var is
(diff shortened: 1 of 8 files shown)


pasrom:
Closing in favour of an extended slice 2 — slice 1's premise (mic = single local speaker) breaks for shared-mic scenarios (conference rooms, multi-person at one mic) where non-user voices would be mislabelled as the configured micName. Replacing with speaker-matching on **both** channels via SpeakerMatcher so the label is driven by enrolled identity, not by channel assumption.

Wire-format work (LiveCaptionLine.speaker field, applyFinalized signature, AppState ↔ state sync) carries over to the follow-up branch — only the 'mic label defaults to settings.micName' bit gets reverted.

PR #337, opened 2026-05-25
feat(app): re-open speaker naming on finished jobs (late re-naming recovery)

pasrom:
## Summary

Adds a **recovery path** for the speaker-naming dialog auto-confirm bug: the dialog sometimes confirms before the user names speakers (a stray Enter/Cmd-W while another app holds focus). PR #310 shipped *prevention* (a 750 ms keyboard-shortcut grace); this PR adds the *recovery* that prevention can't fully cover.

A finished (`.done`) meeting that retains its naming data now shows an **"Edit Speaker Names"** button in the menu-bar job row. Clicking it re-opens the naming dialog so the user can name speakers — or re-run diarization — after the fact.

## How it works

- **Retain the naming sidecar** — Confirm/Skip now free only the in-RAM caches (`clearNamingCaches`); the persisted `_naming.json` + audio sidecars stay on disk. Full deletion (`removeNamingData`) happens only on explicit **Dismiss** or cancel. This mirrors what the app already does with every recording: the original `_mix.wav`/`_app.wav`/`_mic.wav` are kept in `recordings/` indefinitely, so the (small) naming data simply lives alongside them rather than being reaped on its own.
- **Don't auto-remove retained jobs** — a `.done` job that retains a sidecar (signalled by a non-nil `namingSlug`) is no longer auto-removed from the menu after `completedJobLifetime`, so the affordance stays reachable. Jobs that never produced naming data keep the original ~60 s auto-clear.
- **Re-open affordance** — `reopenSpeakerNaming(jobID:)` reloads the retained data (RAM, else from disk) and flips `.done → .speakerNamingPending`, reusing the existing `.showSpeakerNaming` path. Confirming folds the applied names back into the sidecar (`persistAppliedNames`) so a *second* re-name anchors on the current transcript labels instead of the original auto-names.
- **Log recognition once per job** — a re-opened + re-confirmed job no longer writes a second recognition-stats row whose `suggested` baseline is the already-applied names (a phantom perfect match).

## Scope / limitations

- Recovery is **in-session**: `.done` jobs are dropped from the snapshot on relaunch, so the menu affordance does not survive an app restart. The retained sidecar stays on disk (alongside the recordings the app already keeps), so a future "recent recordings" browser could surface it — intentionally out of scope here.
- Finished diarized meetings now **persist in the menu** until Dismiss, where previously they auto-cleared after ~60 s. This is the deliberate tradeoff that keeps the re-name affordance reachable.

## Tests

- 9 new tests across `PipelineQueueTests` + `MenuBarViewTests`: deferred-delete, Dismiss cleanup, auto-remove gating (retained vs not), reopen transition / no-op / RAM-vs-disk, applied-name persistence for repeat re-name, and the "Edit Speaker Names" button visibility + callback.
- Full `PipelineQueueTests` and `MenuBarViewTests` green; SwiftLint clean; release-build parity check passes.

## Follow-up (noted, not in this PR)

The retained `_16k.wav` is a downsampled duplicate of the `_mix.wav` the app already keeps forever; `lateDiarization` could regenerate it on demand from the original instead of retaining a separate copy, removing the duplicate entirely.

### DIFF
--- a/app/MeetingTranscriber/Sources/MeetingTranscriberApp.swift
+++ b/app/MeetingTranscriber/Sources/MeetingTranscriberApp.swift
@@ -93,6 +93,7 @@ struct MeetingTranscriberApp: App {
                 onProcessFiles: processAudioFiles,
                 onDismissJob: { id in appState.pipelineQueue.removeJob(id: id) },
                 onQuit: quit,
+                onReopenNaming: reopenNaming,
             )
         } label: { // swiftlint:disable:this closure_body_length
             Label {
@@ -333,6 +334,10 @@ struct MeetingTranscriberApp: App {
     private func quit() {
         appState.watchLoop?.stop()
         NSApplication.shared.terminate(nil)
+    }
+
+    private func reopenNaming(_ id: UUID) {
+        appState.pipelineQueue.reopenSpeakerNaming(jobID: id)
     }
 
     // MARK: - Pure Helpers (testable without @main)
--- a/app/MeetingTranscriber/Sources/MenuBarView.swift
+++ b/app/MeetingTranscriber/Sources/MenuBarView.swift
@@ -16,6 +16,10 @@ struct MenuBarView: View {
     let onProcessFiles: () -> Void
     let onDismissJob: (UUID) -> Void
     let onQuit: () -> Void
+    /// Re-open the speaker-naming dialog for a finished job that still has a
+    /// retained naming sidecar. Defaulted + last so the many test/preview
+    /// construction sites stay terse.
+    var onReopenNaming: (UUID) -> Void = { _ in }
 
     private var state: TranscriberState {
         status?.state ?? .idle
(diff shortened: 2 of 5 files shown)


pasrom:
Closing this without merging.

The gap it addresses is still real: once Confirm or Skip has been pressed there is no way back into speaker naming, and the keyboard grace period from #310 prevents the common case rather than recovering from it.

What changed is the ground underneath the branch. Since it was cut, `PipelineQueue.swift` has taken 33 commits (+308 / -1418) and has been split into `PipelineQueue+Recovery.swift` and `PipelineQueue+Stages.swift`. The 178 lines this branch changes in that file target code that now lives elsewhere, so resolving the conflicts would mean rewriting the change rather than replaying it. Redoing it against the current structure is the honest path.

The branch `worktree-late-speaker-renaming-recovery` is kept, not deleted, and now also carries two follow-up commits from 2026-05-27 that had never been pushed. They remove both limitations listed in the description above: finished meetings no longer stay pinned in the menu (the sidecar simply stays on disk instead), and naming can be re-opened from disk through a file picker, which also works after an app restart. Those commits and the tests here are the material to reuse.

PR #398, opened 2026-06-07
perf(app): halve live-caption re-transcription cadence

pasrom:
## Problem

Live captions cost ~32 % CPU while recording sits at ~3.5 % (cpu-load lane reference). The dominant cost is structural: every partial re-transcribes the whole accumulated utterance, and the offline encoder zero-pads any input ≤ 15 s to a constant 15 s window — so each ~0.4 s tick pays a full 15 s encoder pass per speaking channel. Window-capping can't help against constant padding; call frequency is the only lever on the multilingual path.

## Fix

Partial interval 0.4 s → 0.8 s: half the encoder passes per speaking channel, captions still update at a conversational rate. Finals (VAD speech-end, 5 s force-flush, stop-flush) are unchanged — only the intermediate-hypothesis refresh slows.

No test pins the constant: a wall-clock throttle assertion needs real sleeps and is CI-flaky by construction. The effect is measured on the cpu-load lane instead — before/after numbers follow as a comment once both runs complete.

## Expected

Captions window ~32 % → ~18–20 % (half the inference cadence; VAD/resampler/final overhead stays).

### DIFF
--- a/app/MeetingTranscriber/Sources/StreamingTranscriber.swift
+++ b/app/MeetingTranscriber/Sources/StreamingTranscriber.swift
@@ -8,7 +8,7 @@ private let logger = Logger(subsystem: AppPaths.logSubsystem, category: "Streami
 ///
 /// PoC scope: ingests 16 kHz mono Float32 buffers (matches `MicCaptureHandler`
 /// post-resample output), feeds them through `FluidVAD` in streaming mode,
-/// and calls `engine.transcribeSamples` once per ~400 ms while speaking
+/// and calls `engine.transcribeSamples` once per ~800 ms while speaking
 /// (partial) and once per detected speech end / 5 s force-flush (final).
 ///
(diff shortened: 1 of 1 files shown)


pasrom:
Measured on the cpu-load lane (same runner, same day, both windows uncontaminated):

| window | main (baseline) | this PR | Δ |
|---|---|---|---|
| live captions CPU | 33.81 % | **22.55 %** | **−33 %** |
| live captions instructions/s | 1 393 M | 845 M | −39 % |
| recording | 3.68 % | 3.73 % | noise |
| idle | 0.19 % | 0.20 % | noise |

Slightly above the naive halving estimate because finals (VAD speech-end, 5 s force-flush) and the VAD itself are unchanged — only the intermediate-hypothesis cadence drops.

Baseline: run 27085228875 (main, post-#397) · this PR: run 27085497252

PR #403, opened 2026-06-08
fix(app): avoid splitting Parakeet tokens mid-word at the token cap

devdotbo:
## Problem

`ParakeetTokenGrouping.groupIntoSegments` turns Parakeet's token-level timings
into transcript segments, ending a segment on sentence punctuation (`.`, `!`,
`?`) or once it reaches a 20-token cap. The cap was a hard split: the segment
ended at exactly the 20th token no matter where that token sat inside a word.

Parakeet can emit a single word as several sub-word tokens (for example `norm` +
`ally`). Consecutive same-speaker segments are later concatenated with a space
during diarization (`DiarizationProcess.mergeConsecutiveSpeakers`, joined as
`"\(current.text) \(seg.text)"`), so a cap split that fell between `norm` and
`ally` surfaced in the final transcript as `norm ally` instead of `normally`.

## Fix

The 20-token cap becomes a soft target. A segment still ends immediately on
sentence punctuation, but a cap-triggered split now waits for the next word
boundary before flushing.

Boundaries are inferred from the whitespace the tokenizer already carries: a
separating whitespace-only token, a trailing space on the current token, a
leading space on the next token, or the end of the stream. Whitespace-only
tokens are skipped from the joined text (they never count toward the cap and
never appear in output) but are now honored as boundary evidence. Blank-token
detection also widens from `CharacterSet.whitespaces` to
`.whitespacesAndNewlines`, so newline-only tokens count as blank too.

The logic is factored into small helpers on the existing pure
`ParakeetTokenGrouping` type (`nextTokenBoundary`, `isWordBoundary`, `isBlank`),
so it stays unit-testable without loading a model. Grouping remains linear:
`nextTokenBoundary` only scans the blank run that immediately follows a non-blank
token, and each blank run is visited once.

## Tests

`ParakeetTokenGroupingTests`: 15 pass. New coverage for the contract:

- **Mid-word cap split.** 19 whole-word tokens then `norm` + `ally` + ` next`
  keeps `normally` intact and defers the split to the following boundary.
- **Standalone whitespace token at the cap** is honored as a boundary.
- **Unbroken run with no boundary.** 25 contiguous non-blank tokens (no
  whitespace, no punctuation) stay in one segment. This pins the new contract
  that the cap is a target, not a hard limit, and it fails against the previous
  hard-cap code (which split at the 20th token into two segments), so it guards
  the behavior rather than restating it.
- The existing whole-word cap test still splits at 20 when every token ends at a
  safe boundary, so the common case does not regress.

## Verification

- Manually verified end-to-end on real meetings: transcripts come out without
  the mid-word spacing artifacts.
- Release build of the app (`swift build -c release`) clean; test target builds
  clean.

The change is a pure function with no model dependency, so the model-heavy
suites are unaffected; CI runs the full suite and lint.

## Notes

The cap is intentionally soft with no absolute ceiling. Forcing a split after a
fixed token count would partially reintroduce the mid-word splitting this fix
removes. If a hard upper bound on segment length is ever needed for pathological
input, it should be a separate, clearly named escape hatch (for example
`hardMaxTokensPerSegment`) rather than re-hardening this cap.

### DIFF
--- a/app/MeetingTranscriber/Sources/ParakeetTokenGrouping.swift
+++ b/app/MeetingTranscriber/Sources/ParakeetTokenGrouping.swift
@@ -4,29 +4,40 @@ import Foundation
 /// Pure token-grouping logic for Parakeet ASR output.
 ///
 /// Extracted from `ParakeetEngine` so the segmentation rules (sentence-end
-/// punctuation, 20-token cap) can be unit-tested without loading a model.
+/// punctuation, soft 20-token cap) can be unit-tested without loading a model.
 enum ParakeetTokenGrouping {
-    /// Maximum tokens per segment before forcing a split — keeps segments
-    /// short enough that they remain useful for the protocol generator
-    /// even when the model emits long runs without terminal punctuation.
+    /// Target maximum tokens per segment before splitting at the next word boundary.
+    ///
+    /// The cap is soft because Parakeet can emit partial-word tokens. Splitting
+    /// immediately at the token limit can turn `normally` into `norm ally` when
+    /// segment texts are joined later.
     static let maxTokensPerSegment = 20
+
+    private static let whitespace = CharacterSet.whitespacesAndNewlines
 
     /// Group token-level timings into sentence-level `TimestampedSegment`s.
     ///
     /// Ends a segment at sentence-terminating punctuation (`. ! ?`) or
-    /// after `maxTokensPerSegment` tokens, whichever comes first. Tokens
-    /// that are blank after whitespace-trimming are skipped entirely.
+    /// after `maxTokensPerSegment` tokens once a word boundary is reached.
+    /// Tokens that are blank after whitespace-trimming are skipped entirely.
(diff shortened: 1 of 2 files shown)


pasrom (approved):

pasrom:
Thanks @devdotbo — merged.

Nice catch on the mid-word split, and the writeup made it easy to follow: the
`norm`/`ally` → `norm ally` path through `mergeConsecutiveSpeakers` is exactly
the failure mode. I like that the cap stays soft and defers to the next word
boundary instead of adding a hard ceiling, and that the new "unbroken run"
test pins the contract by actually failing against the old hard-cap code
rather than just restating it.

Heads-up, unrelated to your change: you may have seen the `label` check go red
on this PR. That's a known limitation of our auto-label workflow on PRs from
forks — GitHub hands fork PRs a read-only token, so the labeling API call
403s. Nothing to do with your code; the fix is in #404.

Thanks again! 🙏

PR #456, opened 2026-06-27
fix(audiotap): default to ScreenCaptureKit so Teams/Zoom audio is captured

### LINKED ISSUE #79: Bug: CATapDescription captures no audio from Microsoft Teams (app.wav always silent)
## Summary

The app audio track (`app.wav`) is **always completely silent** (peak=0, RMS=0) when recording Microsoft Teams calls, regardless of the audio output device used. The CATapDescription / `AudioHardwareCreateProcessTap()` does not capture any audio from the Teams process.

The transcript may appear to work when using MacBook built-in speakers, but this is a false positive — the microphone picks up the speaker output acoustically (feedback). With headphones (AirPods Max), only the local user's voice is transcribed.

## Environment

- **MeetingTranscriber version**: 0.5.1-rc1 (016b50d) — Homebrew
- **macOS**: Tahoe 26.4 (Build 25E246)
- **Hardware**: MacBook Pro 14" (Mac14,5), Apple M2 Max, 32 GB RAM
- **Meeting app**: Microsoft Teams
- **Transcription engine**: Parakeet TDT v3

## Steps to Reproduce

1. Launch MeetingTranscriber
2. Join a **Microsoft Teams test call** (Teams → Settings → Devices → "Make a test call")
   - The test call plays a robotic voice — this is the remote audio to capture
3. Speak into the microphone when prompted
4. Hang up and wait for processing
5. Inspect the generated `app.wav` file with any audio tool

### Test with AirPods Max (Bluetooth)

- `mic.wav`: peak=4735, RMS=139 ✅ (local voice captured)
- `app.wav`: **peak=0, RMS=0** ❌ (silence)
- `mix.wav`: peak=2368, RMS=69 (mic only at half gain)
- Transcript: only local user's speech

### Test with MacBook built-in speakers + mic

- `mic.wav`: peak=32768, RMS=3182 ✅ (local voice + speaker bleed)
- `app.wav`: **peak=0, RMS=0** ❌ (still silence!)
- `mix.wav`: peak=16384, RMS=1589 (mic only at half gain)
- Transcript: includes robot voice — but **only because the mic picks up the speakers acoustically**

### Proof that app.wav silence is the root cause

During the built-in speaker test, the robot speaks at 00:04 while the user is silent. The mic track shows peak=32768 during this segment — the microphone is hearing the speakers, not the CATapDescription tap.

### Real meeting confirmation

A real 16-minute Teams meeting with multiple participants confirmed the same behavior:
- `mic.wav`: peak=30854, RMS=954, duration=981s ✅
- `app.wav`: **peak=0, RMS=0**, duration=981s ❌
- Transcript: only the local user's speech, other participants absent

### With AirPods Max
```
[ERROR] AggregateDevice.mm:442  Input: index 2 >= originalLayout size 2
[ERROR] AudioHardware-mac-imp.cpp:1512  AudioObjectRemovePropertyListener: no object with given ID 204 (repeated 4x)
[ERROR] [com.meetingtranscriber:DualSourceRecorder] App audio rate differs: actual=24000, expected=48000
```

### With MacBook built-in speakers
```
AggregateDevice.mm:887   Creating DefaultDeviceAggregate
AggregateDevice.mm:914   fetched default output device, ID = 93
AggregateDevice.mm:914   fetched default input device, ID = 100
AggregateDevice.mm:688   Built valid aggregate 207
```
No errors logged — aggregate builds successfully, but `app.wav` is still all zeros.

## Additional Context

- **Jamie** (commercial meeting transcription app) captures Teams audio correctly with the same hardware and permissions — likely uses ScreenCaptureKit.
- **Muesli** (open-source, uses ScreenCaptureKit) also captures Teams app audio correctly.
- All macOS permissions granted: Screen Recording ✅, Microphone ✅, Accessibility ✅.

## Hypothesis

The CATapDescription tap is created and the aggregate device is built, but the IOProc callback receives zero-filled buffers from the Teams process. Possible causes:

1. Teams may use a non-standard audio pipeline that CATapDescription cannot intercept (e.g., WebRTC audio routing)
2. The tap may be attached to the wrong audio unit or sub-process within Teams
3. Teams' 24kHz output rate (observed in AirPods logs) may cause format negotiation to produce silence

## Suggested Investigation

1. Log raw buffer amplitude inside the IOProc callback immediately after `AudioHardwareCreateProcessTap()` — before any resampling — to confirm the tap itself delivers silence
2. Verify that the tap targets the correct Teams process/audio unit (Teams may spawn multiple processes)
3. Compare with a ScreenCaptureKit-based capture of the same Teams process to confirm the audio is accessible via SCK but not via CATapDescription
4. Test with other meeting apps (Zoom, FaceTime) to determine if this is Teams-specific

KanaanSalman:
## Problem

The app track is silent when recording **Microsoft Teams** and **Zoom** calls (resolves #79). CoreAudio process taps (`CATapDescription`) deliver zero-filled buffers for these apps — their call/downlink audio renders through WebRTC / Voice-Processing pipelines that the process-tap mixdown excludes. Apps with a conventional output path (e.g. Webex) are captured fine, which is why the failure looked app-specific.

## Fix

Add a **ScreenCaptureKit** capture backend (`SCStream` with `capturesAudio`) that taps the system audio mix — including conferencing apps — and make it the **default**. The CoreAudio process tap stays as an opt-in fallback via `MEETINGTRANSCRIBER_AUDIO_BACKEND=catap`.

- `SCKAudioCapture` — SCStream-based capture, planar→interleaved float32, same level metering as the tap path. Requires Screen Recording permission (already requested by the app).
- `AppAudioCapturing` — protocol both backends satisfy; `AudioCaptureSession` selects the backend.
- Both deliver identical interleaved float32 PCM, so the downstream mix/resample pipeline is unchanged.

## Regression guard

The record-only e2e previously only checked the **mix** WAV's size — but a fully silent app track is byte-for-byte the same size as a good one, and mic speaker-bleed masks it in the mix. Added:

- `scripts/wav_rms.py` — stdlib WAV mean-dBFS meter.
- `scripts/e2e-app.sh` — asserts the recorded **app track** carries energy (`> -60 dBFS`), so a silent-capture regression fails instead of passing on file size alone.

## Notes

- No new permissions (Screen Recording + Microphone already in entitlements/Info.plist).
- Verified: AudioTapLib + full app build clean; 70 audiotap tests pass.

### DIFF
--- a/docs/architecture-macos.md
+++ b/docs/architecture-macos.md
@@ -134,7 +134,7 @@ State writes to `AppPaths.dataDir`; IPC + queue snapshots to `ipcDir`.
 | `AudioConstants.swift` | Shared audio pipeline constants (target sample rate) |
 | `MicRecorder.swift` | Microphone recording via AVAudioEngine |
 | `FluidVAD.swift` | VAD preprocessing via FluidAudio Silero v6 — silence trimming + `VadSegmentMap` timeline remapping |
-| `tools/audiotap/Sources/` | AudioTapLib — CATapDescription-based app audio capture (SPM library) |
+| `tools/audiotap/Sources/` | AudioTapLib — app audio capture (SPM library); ScreenCaptureKit default + CATapDescription fallback |
 
 ### Support
 
@@ -234,14 +234,24 @@ Configurable in **Settings → Audio → Per-Channel Indicator**: master toggle 
 ### Capture
 
 ```
-AudioTapLib (CATapDescription)
-├─ Input: App PID → CoreAudio process tap → aggregate device
+AudioTapLib (default: ScreenCaptureKit; fallback: CATapDescription)
+├─ Input: system audio mix (all apps − self) → SCStream(capturesAudio)
 ├─ Output: Interleaved float32 (mono or stereo) → FileHandle (raw PCM)
 ├─ Mic: AVAudioEngine → mono WAV file (MicCaptureHandler)
 └─ Metadata: micDelay, actualSampleRate, actualChannels via AudioCaptureResult
 ```
 
-**Key:** CATapDescription requires NO Screen Recording permission (purple dot indicator only). Handles output device changes by recreating tap automatically.
+**Backends** (`AppAudioCapturing`, selected by `AudioCaptureSession`):
+
+- **`SCKAudioCapture` (default)** — ScreenCaptureKit. Captures conferencing apps
+  (Microsoft Teams, Zoom) whose call audio is rendered through WebRTC /
+  Voice-Processing pipelines a process tap reads back as zero-filled silence.
+  Follows output-device changes automatically. Requires Screen Recording
+  permission.
+- **`AppAudioCapture`** — CoreAudio `CATapDescription` global process tap. Lower
+  overhead but silent for Teams/Zoom; opt back in with
+  `MEETINGTRANSCRIBER_AUDIO_BACKEND=catap`. Recreates the tap on output-device
+  changes.
 
 ### Processing (DualSourceRecorder.stop())
 
--- a/scripts/e2e-app.sh
+++ b/scripts/e2e-app.sh
@@ -195,6 +195,7 @@ require_command curl
 require_command jq
 require_command swift
 require_command codesign
+require_command python3  # scripts/wav_rms.py — app-track silence guard
 
 [ -f "$SIMULATOR_FIXTURE" ] || fail "simulator fixture not found: $SIMULATOR_FIXTURE"
 
@@ -493,6 +494,29 @@ run_one_record_only_meeting() {
     # → expect > 64 KB even after worst-case truncation.
     [ "$mix_size" -gt 65536 ] || fail "$label: mix WAV suspiciously small: $mix_size bytes (expected > 64 KB)"
     log "$label: mix WAV $mix_path ($mix_size bytes)"
+
+    # The mix-size check above only proves bytes were written — a fully silent
+    # app track (the Teams/Zoom "all-zeros capture" failure mode) is byte-for-
+    # byte the same size as a good one, and the mic can mask it in the mix via
+    # speaker bleed. So assert the *app* track itself carries energy: the
+    # simulator plays the fixture for the whole meeting, so a silent app track
+    # means system-audio capture is broken (e.g. wrong/regressed backend).
+    local app_filename app_path app_dbfs app_min
+    app_min="${E2E_APP_TRACK_MIN_DBFS:--60}"
+    app_filename="$(jq -r '.files.app // empty' "$sidecar")"
+    if [ -n "$app_filename" ]; then
+        app_path="$sidecar_dir/$app_filename"
+        [ -f "$app_path" ] || fail "$label: app track WAV not found: $app_path"
+        app_dbfs="$(python3 "$SCRIPT_DIR/wav_rms.py" "$app_path")" \
+            || fail "$label: could not measure app track RMS ($app_path)"
+        if awk -v v="$app_dbfs" -v t="$app_min" 'BEGIN { exit !(v > t) }'; then
+            log "$label: app track $app_filename energy ${app_dbfs} dBFS (> ${app_min}) — capture OK"
+        else
+            fail "$label: app track $app_filename is SILENT (${app_dbfs} dBFS ≤ ${app_min} dBFS) — system-audio capture produced no signal. Conferencing apps (Teams/Zoom) are silent under the CATapDescription process tap; the SCK backend is required. Check AudioCaptureSession.selectedBackend / MEETINGTRANSCRIBER_AUDIO_BACKEND."
+        fi
+    else
+        log "$label: sidecar has no app track (.files.app null) — skipping app-track energy check"
+    fi
 
     # Negative: record-only short-circuits before VAD/transcription/protocol.
     # No `.txt`/`.md` files from THIS meeting should exist in recordings/.
--- a/scripts/wav_rms.py
+++ b/scripts/wav_rms.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+"""Print the mean RMS level of a WAV file in dBFS.
+
+Used by the live-recording E2E (scripts/e2e-app.sh) to assert that a captured
+*track* actually carries audio — a silent-but-full-size WAV (the Teams/Zoom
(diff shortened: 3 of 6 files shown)


pasrom:
Thanks for digging into this. The genuinely clear win here is the e2e silence guard: today the e2e only checks the mix WAV size, and a silent app track is the same size with mic bleed hiding it. Your `wav_rms.py` + app-track RMS assertion closes exactly that gap, and I would happily take it as its own small PR right away.

On defaulting to ScreenCaptureKit, two things I would like to settle first:

**1. Is the silence still reproducible on current `main`?** `main` now taps the whole `.app` bundle PID tree, not just the root PID. Teams is Electron and renders call audio in a child process, so tapping only the main PID gave silence, but tapping the tree picks it up, and my Teams recordings have audio on a recent `main` build. So Teams is likely already covered. Could you share a repro on current `main` where the app track is still silent, ideally Zoom (native, single PID), since that isolates it from the Electron case? I am not aware of a reproduced Zoom-via-tap silence report. Note also that #79 was actually closed via a TCC permission reset, and its log shows a 24 kHz vs 48 kHz sample-rate mismatch, so the WebRTC/voice-processing explanation is plausible but not confirmed.

**2. SCK as the default has real costs on current `main`.** SCK audio requires the Screen Recording permission (even audio-only), while the CATap path needs only the microphone entitlement, so defaulting to SCK escalates permissions for everyone, with no `#if APPSTORE` guard for the sandboxed build. This branch also predates the capture-time resampling, live-sink, and crash-recovery rework, so making SCK the default would stop app-track live captions, reintroduce the audio drift we fixed, and break crash recovery format-wise.

Suggested path: land the e2e guard separately (I will merge that), and if a concrete app still goes silent on current `main`, add SCK as an opt-in fallback reimplemented against today's capture path (resampler + timeline gap-fill + live-sink + `#if APPSTORE`). The branch is several weeks and hundreds of commits behind, so that is closer to a fresh implementation than a rebase. Really appreciate the work.

drkpxl:
Looking forward to getting this merged in, app doesn't early work otherwise.

pasrom:
Thanks @drkpxl. Two people hitting this makes it worth pinning down, so here is a concrete lead.

We think the silence correlates with **how Teams/Zoom is installed**, not only which app it is. On our side the **native Microsoft Teams desktop app** captures fine on current `main` (the tap walks the whole `Microsoft Teams.app` process tree, so the Electron helper that renders call audio is included). A **Teams web app / PWA** (Teams in a browser tab, or "installed as an app" via Chrome/Edge) is different:

- Meeting **detection** matches on the owning application name (`Microsoft Teams` / `zoom.us`). A browser-hosted Teams is owned by the browser (Chrome/Edge/Safari) or a Chrome-Apps wrapper, so it is never recognized as a Teams meeting and auto-record does not start.
- **Capture** taps the selected app's own `.app` bundle process tree. For a browser tab or PWA the meeting audio lives in a browser process that is not under a `Microsoft Teams.app` bundle, so unless it is pointed at the exact browser process tree it taps the wrong target and records silence.

So "native app works, web/PWA does not" is expected, and it is a different root cause than the WebRTC/voice-processing theory: an app-targeting gap for browser-hosted meetings. This is also exactly where ScreenCaptureKit genuinely helps, since it captures the system audio mix regardless of which process produced it. That is a strong argument for offering SCK, but as an opt-in path for the browser/PWA case rather than the global default (the default swap has the native-path regressions noted above).

Could you both confirm the decisive detail: are you running Teams/Zoom as the **native desktop app**, or as a **web app / browser tab / PWA**, and if a browser, which one? That single answer tells us whether this is a browser-targeting gap or something else.

If it is the browser/PWA case (or anything still silent on current `main`), a short issue would make it actionable, ideally with:

- macOS version and the app version (from the About screen).
- For each app that is silent: native desktop app vs web/PWA, and if a browser, which browser and whether it was "installed as an app."
- A verbose audio log: Settings → Diagnostics → **Verbose Audio Logging**, then start a recording and share the `[debug] Tap target: pid=… exe=… bundle=…` line plus a `[debug] App audio RMS (5s): … dBFS` line. That shows exactly which process/bundle got tapped, which confirms or rules out the browser-targeting theory immediately.

Thanks again, this is genuinely useful.

pasrom:
Thanks for this @KanaanSalman. Where it landed after checking on current `main`:

**The e2e silence guard shipped** (merged in #574). Your idea, assert the captured *app track* itself carries energy rather than just the mix size, was the clear win. `main` already had an equivalent analyzer used by the browser-meeting e2e lane (`mt-cli wav-verdict`), so I wired the record-only leg to that. Full credit for the idea is in that PR.

**Switching the default backend to ScreenCaptureKit I am declining.** For audio-only capture on macOS 14.2+, Core Audio process taps are Apple's purpose-built API and need only the audio-capture grant; SCK audio requires Screen Recording with no audio-only exemption (it is a screen-capture framework, and forces a dummy video target). Making SCK the default would put a Screen Recording prompt in front of every user and capture un-isolated system audio, against this app's whole "no Screen Recording" positioning. There is also no `#if APPSTORE` guard, so the sandbox build would default to the backend it can least grant.

**That said, you did identify a real gap, and I don't want to wave it away.** Voice-processed call audio (Teams/Zoom via `AUVoiceProcessingIO`) can reach the tap as zero/near-silent buffers; that is a documented failure mode. But a default swap is the wrong fix for it: the same voice-processing path *ducks other sources for SCK too*, so SCK is not a guaranteed capture, and in this project that exact symptom has also traced to a missing audio-capture grant (#524, fixed in v0.6.0) rather than the codec. So the right response is a targeted fallback, keeping taps as default: detect a running-but-silent tap and rebuild it, and/or an opt-in per-channel SCK fallback behind the Screen Recording grant a user has already given.

For the native and browser cases specifically, capture already works on `main`: native Teams/Zoom via the whole-`.app`-bundle PID-tree tap, and browser meetings via issue #503 (opt-in WebRTC power-assertion detection, no Screen Recording, the same tap captures the audio, e2e-proven non-silent).

Closing this PR since the useful half is merged and the default-backend change is declined above. If you can still reproduce a silent app track on current `main`, please open an issue with Verbose Audio Logging on and the `[debug] Tap target: ... bundle=...` line (plus native vs browser) — that is what tells the zero-buffer case apart from a TCC gate, and it is the starting point for the fallback above. Thanks again for pushing on this.

PR #497, opened 2026-07-10
Fix custom output directory not persisting in non-sandboxed builds

igorlenskiii-web:
## Symptom

Settings → Output → choosing a custom output folder appears to work (the UI shows the selected path), but nothing persists: every new recording writes to the default `~/Downloads/MeetingTranscriber/` again, and the folder is silently recreated there. Observed on the Homebrew/direct build v0.6.0-rc6, macOS 26.5.

## Root cause

`AppSettings.setCustomOutputDir` stores the folder as a **security-scoped** bookmark:

```swift
guard let data = try? url.bookmarkData(options: .withSecurityScope, ...) else { return }
```

Security-scoped bookmarks require the App Sandbox. The direct (non-App-Store) build is not sandboxed (it can't be — CATap capture), so `bookmarkData(options: .withSecurityScope)` throws, the `guard try?` swallows the error, and the setting is never written. `effectiveOutputDir` then falls back to `AppPaths.downloadsProtocolsDir` forever, while the Settings UI keeps displaying the user's chosen path.

## Fix

- Extract bookmark creation/resolution into `makeOutputDirBookmark` / `resolveOutputDirBookmark` helpers.
- Creation: try `.withSecurityScope` first (APPSTORE build keeps its behavior), fall back to a regular bookmark for the non-sandboxed build, and log an error instead of failing silently.
- Resolution: mirror the fallback (`.withSecurityScope` → plain), so both bookmark kinds resolve regardless of which build wrote them.
- Stale-bookmark re-creation goes through the same helper.

`AppSettings+RPC.rpcOutputDirPath` needs no change — it already resolves with `[.withoutUI, .withoutMounting]`, which works for both bookmark kinds.

## Testing

Reproduced the symptom on macOS 26.5 with the rc6 direct build (chosen folder shows in UI, files keep landing in `~/Downloads/MeetingTranscriber`). The patch was reviewed against the existing style but not compiled locally — relying on CI (unit tests + E2E) to exercise it; happy to adjust if anything trips.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -1,4 +1,7 @@
+import os.log
 import SwiftUI
+
+private let logger = Logger(subsystem: AppPaths.logSubsystem, category: "AppSettings")
 
 // SwiftFormat strips redundant raw values matching their case name, which then
 // trips SwiftLint's `raw_value_for_camel_cased_codable_enum`; the rawValues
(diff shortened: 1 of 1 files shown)


pasrom:
Thanks for digging into this and for the clean, well-documented change. The symptom you describe (custom output folder shows in the UI but recordings keep landing in `~/Downloads/MeetingTranscriber`) is real and worth fixing, so I want to get the root cause pinned down before merging, because I could not reproduce the specific mechanism this PR targets.

## I could not reproduce the scoped-bookmark throw

The stated root cause is that `bookmarkData(options: .withSecurityScope)` throws in the non-sandboxed direct build, so `guard try?` swallows it and the folder never persists. I tried to reproduce that throw and could not, in any configuration I tested on the direct (non-App-Store) build:

- a bare non-sandboxed process,
- a real signed `.app` bundle with the exact Homebrew entitlements (mic only, no `com.apple.security.app-sandbox`), both ad-hoc and Developer-ID signed,
- create in one process / resolve in a relaunched process,
- TCC-protected targets (Documents / Desktop / Downloads).

In all of them `url.bookmarkData(options: .withSecurityScope)` succeeds and `URL(resolvingBookmarkData:options: .withSecurityScope)` resolves it back; `startAccessingSecurityScopedResource()` even returns `true`. Apple's docs say scoped bookmarks are "used in sandboxed apps", but a non-sandboxed app creating and resolving them does not appear to throw. If you have a build/OS where the scoped `bookmarkData` call actually throws, a captured error (the `NSError` domain/code) would be very helpful, because that is the crux.

## Why that matters for this diff

Because the scoped-create branch succeeds on the direct build, `makeOutputDirBookmark` returns from its first `if let` and the new `options: []` create fallback is never reached there. So in the non-sandboxed build the bytes stored by `setCustomOutputDir` are the same scoped bookmark as before, and `customOutputDir` resolves them the same way as before. That means this change would not alter the stored-or-resolved behavior on the affected build, which is what makes me think the persistence failure is coming from somewhere other than bookmark creation.

## A likelier cause to check

`customOutputDir` feeds `effectiveOutputDir`, and the pipeline captures `effectiveOutputDir` **by value** when the processing queue is built; that queue is only rebuilt at an idle watch-start. So if you pick a new folder while watching is active or a job is in flight, the Settings UI updates immediately (it reads `effectiveOutputDir` live) but the running queue keeps writing to the previously captured directory. That reproduces your exact symptom with the bookmark working correctly. Could you check:

- Does the chosen folder stick after a full quit + relaunch, or after toggling watching off and back on while idle? If it starts working then, this is the queue-capture path, not bookmarks.
- After picking a folder, is `customOutputDirBookmark` actually present in the app's defaults?

## What I would happily take from this PR regardless

- The **resolve-side tolerance** (accepting a plain bookmark as well as a scoped one) is a genuine robustness win: resolving a *plain* bookmark with `.withSecurityScope` does throw `Cocoa 259`, so once any non-scoped bookmark exists, the fallback is the right call.
- The `logger.error` on creation failure is a good idea too.

The one thing that needs to change before this could merge is the doc comment on `makeOutputDirBookmark`: it states the "`.withSecurityScope` throws in the non-sandboxed build" mechanism as established fact, and I have not been able to reproduce that, so it should not be baked into the source as the rationale.

Appreciate the contribution; let's confirm the repro and I think we land the useful parts of this quickly.

pasrom:
Thank you for this, and sorry it sat so long. The symptom you reported is real, and it now has a fix in #691. The mechanism turned out to be a different one, and since none of this diff survives that finding, I am closing this PR. The report is what led to the fix, so it did its job even though the patch does not land.

## What was checked

The stated root cause is that `bookmarkData(options: .withSecurityScope)` throws in the non-sandboxed build. That could not be reproduced in any configuration, as noted above. What settles it is the history: the custom output folder arrived in v0.5.1, and every released version that creates a bookmark at all creates a scoped one.

```
v0.2.0 - v0.4.1   no custom output folder yet
v0.5.1            AppSettings.swift                     options: .withSecurityScope
v0.6.0            AppSettings.swift                     options: .withSecurityScope
v0.7.0            AppSettings.swift                     options: .withSecurityScope
v0.8.0            AppSettings+OutputDirectory.swift     options: .withSecurityScope
```

So no shipped build has ever written a plain bookmark. That makes the create-side `options: []` fallback a guard against a state this codebase cannot produce, and it makes the resolve-side tolerance, which I called a robustness win in my earlier comment, a defence against bytes nobody ever wrote. I was too quick there, and the history is what corrects me.

## Where the symptom actually comes from

Two separate things produced what you saw.

The display half was its own bug: `customOutputDirBookmark` was a computed passthrough to `UserDefaults`, so `@Observable` had no stored property to track and Settings kept showing the previous folder until relaunch. That was fixed in #626.

The half that matches your description was still open until now. `resolveCustomOutputDir` is a `try?`, so when the bookmark stops resolving, `effectiveOutputDir` quietly yields `~/Downloads/MeetingTranscriber`. Settings still shows the chosen folder, recordings go elsewhere, and nothing says so. It is reachable whenever the folder is not there at that moment: an external drive unplugged, a network share unmounted, the folder renamed or deleted. If your output folder lived on any of those, that is the path you hit.

#691 keeps the fallback, because a recording must never be blocked or lost over a missing folder, and adds the telling: one notification per unavailability episode, naming the folder it is standing in for and where output is going instead.

## What of your patch is in it

The idea behind your `logger.error` is. `setCustomOutputDir` used to be `guard let data = makeBookmark(for: url) else { return }`, so a creation failure meant the user picked a folder and nothing happened, silently. It now keeps the previous choice and logs why.

Not the create-side fallback and not the resolve-side tolerance, for the reason above. And not the doc comment, since it stated the unreproduced mechanism as established fact.

Thanks again for the report and for the careful write-up. It pointed at something that was genuinely broken.

Issue #503, opened 2026-07-15
Feature Request: Web Browser Meeting Detection & Recording


BrodeOne:
# Feature Request: Web Browser Meeting Detection & Recording

Hey there,

Just a heads up before jumping in: I'm pretty new to this and I don't want to pretend I know exactly how the codebase handles all the deep macOS APIs. So the technical "how-to" part below is more of a rough logic concept. The "why" is the main thing!

### The Problem
Right now, the app is awesome at catching native desktop apps. But we are missing a huge chunk of meetings that happen directly in the browser. 
* Platforms like **Google Meet** and **Whereby** don't even have native Mac apps.
* A lot of people use the web versions of Teams, Zoom, or Webex because they either don't want to install the heavy desktop apps or their company IT strictly blocks them.

Since the app can't catch these right now, users have to manually click to start recording, which kinda defeats the whole purpose of the auto-detection magic.

### How it could work (the basic logic)

Since we can't hook into web apps as easily as desktop apps, I was thinking about a flow like this:

**1. Watch the window titles**
The app could occasionally (maybe every 3 seconds?) check the open window titles of popular browsers like Chrome, Safari, Edge, Firefox. It would just look for patterns like `"Zoom Meeting"` or `"Standup - Google Meet"`. 

**2. Double-check so it doesn't glitch**
Because web pages can load slow or people quickly flip through tabs, it probably shouldn't trigger instantly. Maybe it needs to see the meeting title twice in a row (over ~6 seconds) before doing anything. This way it ignores random loading screens.

**3. Ask first (No auto-start)**
Browser tabs can be messy. Instead of just auto-recording (like it does for desktop apps), it should pop up a simple alert: *"Do you want to record this meeting?"* 
* If they hit **Record**, it grabs the audio and starts.
* If they hit **Ignore**, it goes on a 60-second cooldown so it doesn't spam the user with popups.

**4. Grabbing the audio**
I know browsers like Chrome use a bunch of hidden background processes for rendering audio. When the user hits record, the app would need to target the browser and pull in all its sub-processes to make sure it actually captures the sound, and then mix it the way it normally does.

### Settings Menu Updates
To make it user-friendly, we could update the `Settings -> Apps to Watch` menu:
* Add normal checkboxes for **Google Meet** and **Whereby**.
* Add a master toggle for **Browser Web Meetings**. If a user turns this off, the app completely ignores browser tabs to save resources.

[...]

PR #527, opened 2026-07-22
feat(app): detect Google Meet calls in Chromium browsers

ohernandezdev:
## Summary

Adds auto-detection for Google Meet calls running in a Chromium browser (Chrome, Brave, Edge, Chromium), alongside the existing Teams/Zoom/Webex detection.

Meet has no dedicated app, so no app-specific power assertion exists. Detection therefore requires **two signals** before the browser is recorded:

1. The browser process holds a WebRTC power assertion (`WebRTC has active PeerConnections`). Any site using camera/mic holds the same one, so the assertion alone must never fire — implemented as `AssertionPattern.requiresWindowConfirmation`.
2. An in-call Meet tab title (`Meet – <code or name>`) confirms it. Title matching for browser-hosted patterns is **strict** (`AppMeetingPattern.strictTitleMatch`): `selectWindowTitle` gets no first-non-idle fallback, since any unrelated tab title would otherwise pass as "the meeting".

Recording the browser's root PID captures Meet audio because `resolveTapPIDs` already expands to every PID under the app bundle (same mechanism as Electron apps). Mid-call tab switches keep the meeting alive: `isMeetingActive` checks the assertion only, so the recording survives the Meet tab losing focus.

## Changes

- `AppMeetingPattern`: new `strictTitleMatch` field + `.meet` pattern (in-call regex anchored at start only, so profile/browser title suffixes still match; `^Google Meet\b` landing page is idle)
- `MeetingTitleMatcher`: honors `strictTitleMatch` in `selectWindowTitle`
- `PowerAssertionDetector`: new `requiresWindowConfirmation` on `AssertionPattern` + Google Meet pattern (keyword `webrtc`)
- `AppSettings`: `watchMeet` toggle (default on) wired into `watchApps`
- General settings: new "Google Meet (Chrome, Brave, Edge)" toggle in Apps to Watch
- RPC settings snapshot: `detection.watchMeet`

## Limitations

- Needs Screen Recording permission (window titles). Without it the Meet pattern simply never fires — the two-signal gate fails closed rather than recording the whole browser.
- Initial detection requires the in-call Meet tab to be the active tab in some browser window during the confirmation polls (window titles only expose the active tab). In practice you just joined the call, so it is.

## Testing

- 12 new unit tests in `GoogleMeetDetectionTests.swift`: detection in Brave/Chrome, browser-suffix titles, WebRTC assertion without a Meet tab (must not fire), Meet landing page (idle, must not fire), non-WebRTC browser assertions with a Meet tab open (must not fire), strict-vs-non-strict title selection, mid-call tab switch keeps the meeting active, Apps-to-Watch filtering
- Updated the pattern-count/allowlist pins (`AppMeetingPattern.all`, RPC settings key allowlist, `watchApps` expectations)
- Full suite green locally: 2222 tests, 0 failures (skipping model-download E2E + snapshot lanes)
- Verified end-to-end against a real Google Meet call in Chrome on macOS 26: auto-detect → dual-source recording → transcription/diarization → protocol

### DIFF
--- a/app/MeetingTranscriber/Sources/AppSettings+RPC.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings+RPC.swift
@@ -34,6 +34,7 @@
                 watchTeams: watchTeams,
                 watchZoom: watchZoom,
                 watchWebex: watchWebex,
+                watchMeet: watchMeet,
                 autoWatch: autoWatch,
                 pollIntervalSeconds: pollInterval,
             )
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -112,6 +112,13 @@ final class AppSettings {
         didSet { defaults.set(watchWebex, forKey: "watchWebex") }
     }
 
+    /// Google Meet in a Chromium browser (Chrome, Brave, Edge). Detection needs
+    /// Screen Recording permission: the browser's WebRTC power assertion is
+    /// only trusted when an in-call Meet tab title confirms it.
+    var watchMeet: Bool {
+        didSet { defaults.set(watchMeet, forKey: "watchMeet") }
+    }
+
     /// Auto-start watching on app launch.
     var autoWatch: Bool {
         didSet { defaults.set(autoWatch, forKey: "autoWatch") }
@@ -457,6 +464,7 @@ final class AppSettings {
         if watchTeams { apps.append("Microsoft Teams") }
         if watchZoom { apps.append("Zoom") }
         if watchWebex { apps.append("Webex") }
+        if watchMeet { apps.append("Google Meet") }
         return apps
     }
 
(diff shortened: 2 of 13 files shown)


pasrom:
Thanks for this, and sorry for the overlap. Really nice work (the two-signal design, 12 tests, real end-to-end verification).

Browser meeting detection actually landed just after you opened this, via #523 (issue #503, merged 2026-07-25). It uses the WebRTC power assertion directly instead of window titles, so it needs no Screen Recording and isn't tab-limited, and it's consent-gated (opt-in) rather than default-on. Same goal, incompatible mechanism, so I'll close this as superseded (it's also now conflicting against the files #503 rewrote).

The one thing yours adds that we don't have is multi-browser support (Brave, Edge, Chromium) since ours matches Chrome only. I'll track that as a follow-up on the assertion-based path.

Thanks again, and please don't let this discourage future PRs. 🙏

PR #532, opened 2026-07-26
test(app): prove posted key events drive a SwiftUI TextField binding

pasrom:
## Summary

Thin PoC that closes one of the E2E "last mile" gaps: **typing into fields**, currently listed as manual-QA-only.

The GUI Testing guidance treats typing as un-automatable because AX set-value does not fire the SwiftUI binding (hence no `/ui/setValue`). That is only true for the **AX value attribute**. A real **posted key event** goes through the `keyDown:` -> `insertText:` responder chain that the binding observes, so it **does** update the binding.

This proves the mechanism:
- Host a `TextField` (bound to a reference box) in a key `NSWindow`.
- Focus the field, post per-character `NSEvent`s via `NSWindow.sendEvent`.
- Assert the bound value updated.

It runs in the **plain xctest process** (no `NSApp`, no TCC, no self-hosted runner) because `sendEvent` routes to the first responder directly.

## Why it is non-vacuous

Verified by stubbing the injector to a no-op: the assertion goes red (bound value stays empty). Only the posted events turn it green.

## Scope

- Test-only. The injector is deliberately test-local (no production visibility widening).
- One positive test; a fragile AX-negative test was intentionally left out.

## Follow-up (not in this PR)

Wire the same `NSEvent` injection into the in-process `/ui` debug surface as a `/ui/type` endpoint, then a live test types into a real Settings field and asserts the `AppSettings` write-back via `/state`. That turns the mechanism into end-to-end coverage of the manual "typing" acceptance. Drag/focus-order and the `NSOpenPanel`/`NSAlert` seams are the natural next steps after that.

### DIFF
--- a/app/MeetingTranscriber/Tests/KeyboardInjectionPoCTests.swift
+++ b/app/MeetingTranscriber/Tests/KeyboardInjectionPoCTests.swift
@@ -1,0 +1,110 @@
+#if !APPSTORE
+    import AppKit
+    @testable import MeetingTranscriber
+    import SwiftUI
+    import XCTest
+
+    /// PoC for the e2e "last mile": prove that a *posted key event* updates a
+    /// SwiftUI `TextField`'s binding, unlike an AX set-value (which the `/ui`
+    /// harness deliberately does not expose, because AX-set does not fire the
+    /// SwiftUI binding). If this is green, the same mechanism is what a future
+    /// in-process `/ui/type` endpoint would use to automate the "typing into
+    /// fields" acceptance that is currently manual-QA-only.
+    ///
+    /// The injector is intentionally test-local (not production visibility
+    /// widening): slice 1 only establishes the mechanism.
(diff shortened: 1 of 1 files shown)


pasrom:
Closing in favour of folding this into the /ui/type slice.

The mechanism this proves is sound (verified: green, and non-vacuous - stubbing the injector to a no-op turns it red). But as a standalone PR it tests SwiftUI/AppKit framework behaviour rather than any production code, which the GUI Testing guidance explicitly lists under "Don't". Nothing depends on the assumption yet either, so guarding it in isolation is premature.

Once /ui/type exists, the same test stops being a framework test in a vacuum and becomes the layer-1 unit test of the mechanism that endpoint is built on - which is exactly what the guidance asks for. It will land there alongside a real integration test that types into an actual Settings field and asserts the AppSettings write-back via /state.

PR #536, opened 2026-07-27
test(ci): add a long-session soak lane

pasrom:
Closes the biggest gap a blind-spot review of the testing strategy found: **long sessions are entirely unobserved**.

Every existing lane records for seconds. Real meetings run 30-120 minutes and the menu-bar app stays resident for weeks. A steady leak in the recording buffers, the caption ring buffer, or the pipeline snapshots passes `e2e-app` (which asserts a transcript appears) and `e2e-cpu-load` (which averages CPU over a 25-second window) without either noticing.

## What it does

Holds one meeting open for an hour, samples `GET /metrics` throughout, reports how the process footprint moved — and then asserts the pipeline **still produces a job**. Leaking and quietly stopping working are different failures; a lane that only measured memory would pass while the app had stopped recording.

## The measurement choice is load-bearing

Growth is a **least-squares slope over all samples**, not last-minus-first. Checked against six synthetic series:

| series | delta | slope |
|---|---|---|
| flat | 0 | 0 MB/h |
| +1 MB/min | 30 | 60 MB/h |
| spike in the middle, ends equal | 0 | 0 MB/h (peak still reported) |
| **leak ending on a low sample** | **50** | **300 MB/h** |
| shrinking | -100 | -600 MB/h |
| one sample | — | errors, no silent verdict |

The fourth row is why: last-minus-first would have understated that leak six-fold. The arithmetic is pure over its input, so it is verifiable without a running app — which matters, because it carries the lane's entire verdict.

## Gating and cadence

Follows the CPU-load lane's policy: log everything, gate only a generous catastrophe bound, emit a `RESULT` line with the full sample series so a real threshold can come from observed variance rather than a guess.

**Weekly, not nightly.** The run occupies the shared `audio` runner for over an hour; every PR-triggered lane would queue behind a nightly one. Leaks are slow-moving, so weekly detection is soon enough — raise it if the trend data earns it.

## Honest limitation

I could not execute a full soak before opening this: it needs the self-hosted runner, takes over an hour, and would have meant driving the app on a machine someone is working on. Verified locally: shell syntax, the growth arithmetic against the six series above, the preflight guards (non-integer, zero, and a sampling interval that would yield fewer than two samples all refused), and `--help`. The first real run is the scheduled one, which is also why it starts life gating nothing but a catastrophe bound.

### DIFF
--- a/.github/workflows/e2e-soak.yml
+++ b/.github/workflows/e2e-soak.yml
@@ -1,0 +1,108 @@
+name: E2E (Soak)
+
+# Long-session memory soak of the deployed production app: hold one meeting
+# open for an hour, sample RPC /metrics throughout, and report how the process
+# footprint moved (scripts/e2e-soak.sh).
+#
+# Why this exists: every other lane records for seconds, while real meetings run
+# 30-120 minutes and the menu-bar app stays resident for weeks. A steady leak in
+# the recording buffers or the pipeline snapshots is invisible to all of them.
+# The lane also asserts the pipeline still produces a job afterwards — leaking
+# and quietly stopping are different failures.
+#
+# Weekly, not nightly, on purpose: the run occupies the shared `audio` runner for
+# well over an hour, which would delay every PR-triggered lane behind it. Leaks
+# are slow-moving, so weekly detection is soon enough; raise the cadence if the
+# trend data turns out to be worth it.
+#
+# Measure-and-log phase: gates only a generous catastrophe bound, and collects a
+# RESULT artifact per run so a real threshold can be derived from observed
+# run-to-run variance instead of guessed.
+
+on:
+  workflow_dispatch:
+    inputs:
+      minutes:
+        description: "Soak duration in minutes"
+        required: false
+        default: "60"
+  schedule:
+    # Sundays 06:30 UTC — after every nightly lane (quality 03:00, appstore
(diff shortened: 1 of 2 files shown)


pasrom:
Marking this draft: a real short run on the self-hosted runner (6 min, then 2 min with real audio) found two defects that the arithmetic-only verification could not.

**1. The closing assertion never fires.** Record-only mode writes to `<effectiveOutputDir>/recordings/`, i.e. `~/Downloads/MeetingTranscriber/recordings/`, and after a completed meeting that directory stayed empty — so the lane failed on an otherwise healthy run. `e2e-app.sh --record-only` already validates exactly this (sidecar + mix WAV), so the fix is to reuse its proven wait-and-assert rather than re-derive one.

**2. Short runs report a false leak.** With real audio the footprint went 71 -> 103 -> 103 MB: an allocation ramp when recording starts, then flat. Fitted over two minutes that reads as 960 MB/hour and would trip the catastrophe bound. The settle window does not cover it, because the ramp happens when RECORDING starts, not at launch. The baseline has to be taken after the recording ramp, not at t=0.

What did hold up: detection, recording, sampling, the least-squares fit, the RESULT line, and the argument guards all behaved exactly as intended on the runner.

Both defects would have shipped and then failed on the first scheduled run, where they would have looked like a broken app rather than a broken lane.

PR #548, opened 2026-07-30
feat(app): MicInputDetector — detect WeChat / Tencent Meeting / FaceTime / WhatsApp calls via Core Audio mic-input state

jianjettfu-oss:
## What

A third meeting-detection channel beside window-title polling and power assertions: **mic-input state**. `MicInputDetector` polls Core Audio process objects (`kAudioHardwarePropertyProcessObjectList`, macOS 14.2+ — the same API family the recorder's tap uses) and reads `kAudioProcessPropertyIsRunningInput` per process, matching against watched bundle IDs. A `CompositeMeetingDetector` runs the assertion detector and the mic detector behind the existing `MeetingDetecting` protocol, so `WatchLoop` is unchanged.

Apps enabled by this channel: **WeChat, Tencent Meeting, FaceTime, WhatsApp** — four toggles in Settings → General, **all off by default** (additive opt-in; existing installs see no new auto-recording behavior).

## Why

These call apps can't be detected by the existing channels:

- their in-call power assertions are unnamed, absent, or indistinguishable from idle behavior (nothing like Teams' "call in progress"),
- their call-window titles are localized and unstable (微信/腾讯会议 UI strings), so title patterns don't hold.

A call, however, always captures the microphone — `IsRunningInput` flipping true on the app's bundle is the reliable signal. For hundreds of millions of users WeChat is *the* calling app, and it's precisely the app no bot-based tool can ever join — local capture is the only architecture that covers it, which makes it a natural fit for this project.

## Design notes

- Same confirmation-count (2 consecutive polls) and cooldown (5 s) semantics as `PowerAssertionDetector`, so brief mic grabs (dictation, a short voice message) don't trigger.
- Title lookup reuses `MeetingTitleMatcher`; these apps ship empty `meetingPatterns`, so titles fall back to the clean `"<App> Call"` placeholder.
- **Unmatched-bundle diagnostic**: any process capturing mic input that matches no watched app is logged once per session (`bundle`, `.public`). This is the discovery channel for apps whose audio is owned by a different daemon than the app bundle — on first live test it immediately named `com.apple.CoreSpeech` (dictation) as a non-call mic holder, and it will name FaceTime's real audio-owning bundle if `com.apple.FaceTime`/`com.apple.avconferenced` turn out not to be it.
- `isMeetingActive` is per-strategy: each detector answers only for apps in its own pattern set, so the composite's OR is exact.

## Testing

- `MicInputDetectorTests` (detection, confirmation threshold, counter reset on gaps, liveness, cooldown, pattern filtering, drift guard mirroring the assertion detector's) and `CompositeMeetingDetectorTests` (first-hit-wins, fall-through, per-strategy liveness) are included.
  **Correction:** an earlier revision of this description claimed these were green locally — they were not run. This machine has Command Line Tools only (no Xcode), so `swift test` cannot resolve `XCTest` and the test target never compiled; the exit code I trusted came from a pipe, not from the test run. The tests are written against the same seams the existing `PowerAssertionDetectorTests` use, but **please treat them as unverified until CI runs them** — I'd appreciate the maintainer approving the workflow run on this PR.
- Live on macOS 26.1: detector polls real Core Audio state inside the running app (unmatched-bundle diagnostic observed firing); end-to-end recording of a simulator meeting through the composite detector verified.
- Caveat flagged in code: FaceTime's audio-owning process is listed empirically (`FaceTime` + `avconferenced`); the unmatched diagnostic will surface the real bundle if neither matches during a live FaceTime call.

Works best together with #547 (without `NSAudioCaptureUsageDescription` the recording these detections trigger has a silent app track).


pasrom:
Reviewed this including an independent security pass, and verified the mechanism on a real machine before touching it. Summary: the approach is sound and the architecture fit is right, so I pushed four commits rather than sending review round-trips. One design question is left for the maintainer, below.

**Mechanism, verified here (macOS 26.5.2).** `kAudioHardwarePropertyProcessObjectList` returns the full process list with bundle IDs, and a helper holding the mic flips `kAudioProcessPropertyIsRunningInput` true with correct bundle attribution and back to false the moment it exits. Run from a freshly signed bundle with zero TCC rows, launched as its own responsible process: full list, no prompt, no TCC row created. So this channel adds no permission surface and keeps the no-Screen-Recording default path intact.

**Self-trigger and termination are clear.** Our own recorder shows up as a running-input process but can never match the fixed four-app allowlist, the unmatched-bundle log excludes `Bundle.main.bundleIdentifier`, and `isMeetingActive` is pattern-scoped, so our recording cannot keep a foreign meeting alive.

**Security review: nothing found.** Diffed against the merge base rather than the file list, so nothing could hide in an unexpected file. No network path of any kind, no file I/O at all (state is four UserDefaults booleans plus os.log), no `Process`/`NSTask`/`dlopen`, no entitlement, Info.plist, Package.swift, script or workflow changes, no user content logged, nothing executing at build time. The watched list is hardcoded and the toggles only filter it, so there is no user-controlled bundle-ID injection.

**Tests are meaningful.** I neutralised the confirmation threshold (`hits >= 1` instead of `>= confirmationCount`) and 3 of 12 tests failed, so they are not vacuous. They also all pass here, which resolves your "unverified, no Xcode" caveat positively.

**What I pushed**

1. `fix(app)` skip the round when no app is watched. All four toggles ship off, so the common case was a detector with an empty pattern set that still enumerated every Core Audio process object on every poll, and every mic-using bundle fell into the unmatched-bundle diagnostic, which logs it `.public` once per session. Installs that never opted in were paying for the feature and getting the bundle ID of every app that touched their microphone into any diagnostics export. Pinned with a test asserting the process provider is not called at all.
2. `docs(app)` correct the comment about false positives. It said longer non-call mic grabs are "filtered downstream by the consumer (min-duration policy)", but there is no such policy: no minimum-duration filter exists in WatchLoop, WatchLoopEndPolicy or PipelineQueue. A WeChat or WhatsApp voice message past roughly `pollInterval * confirmationCount` (about 6 s at the defaults) starts a real recording and a pipeline job. The comment now names the actual mitigations.
3. `test(app)` split the composite cases into their own file: `swiftlint --strict` fails on two XCTestCase classes in one file, which blocks CI.
4. `chore(app)` exempt the AppSettings initializer from `function_body_length`: the four new toggles pushed it to 64 lines against the 60-line cap, another CI blocker. Kept to a single line because that file is at 599 of its own 600-line cap.

Also rebased onto main, so the branch was force-updated. Gates on the result: SwiftFormat (the pinned 0.61.1) 0/395, `swiftlint --strict` clean, 176 tests across the touched and adjacent suites green.

**Left for the maintainer.** The strongest lever against the voice-message false positive is `AppMeetingPattern.requiresRecordingConsent`, which already exists and is what the browser channel uses. These are personal 1:1 call apps, so a confirm-before-recording prompt is arguably right for them anyway. I deliberately did not enable it, for two reasons: it is a product decision, and the consent flow currently depends on a delivered notification, which has an open failure mode when notifications are suppressed. Worth deciding explicitly rather than by default.

Two things remain unverified: the WeChat and Tencent Meeting bundle IDs (neither app installed here), and which process actually owns FaceTime audio during a live call. The diagnostic log you ship is the right way to find the latter out.

pasrom:
Follow-up on the two bundle IDs I could not check locally (neither app is installed here). The Homebrew casks are explicit about app identity, so they settle it without needing the apps:

- **WeChat** `com.tencent.xinWeChat` confirmed: the `wechat` cask quits exactly that ID. Its sandboxed helpers `.IPCHelper`, `.MiniProgram` and `.WeChatMacShare` also exist, and I deliberately did not add them, since a mini program recording audio is not a call.
- **Tencent Meeting** `com.tencent.meeting` confirmed: the `tencent-meeting` cask quits that ID.
- **The international build was missing.** VooV Meeting, which the cask names "Tencent Meeting International Version", ships as `com.tencent.tencentmeeting` and zaps `~/Library/Containers/com.tencent.tencentmeeting`. Anyone on that build would never have been detected. Added, with a test.
- **`com.tencent.wemeet` I could not corroborate.** The only wemeet-namespaced artifact anywhere is a `.FileDelta` helper container in the mainland cask, so it may never match a main app process. I left it in as insurance, since an ID that never appears costs nothing, but the comment now says it is unverified rather than asserting it.

Also fixed the four pre-existing tests that this change breaks and CI caught: two pinned `AppMeetingPattern.all.count == 5`, and two cast `defaultDetector(settings:)` to `PowerAssertionDetector`, which is now a `CompositeMeetingDetector`. To keep the toggle-wiring assertions honest without reaching into the composite, `defaultDetectors(settings:)` is now split out of `defaultDetector`, which becomes a one-line composition over that list.

Full app suite run locally this time, not filtered: green apart from `MenuBarIconSnapshotTests`, which fails on local rendering and passes in CI.

Still open, and only answerable in the field: which process actually owns FaceTime audio during a connected call. I watched Core Audio for two minutes with FaceTime.app running and saw no input-running edge at all, which is expected since the app has to be in a call, not merely open. Your once-per-session diagnostic for unmatched running-input bundles is the right way to capture it.

pasrom:
FaceTime is settled. Measured on macOS 26.5.2 with a real call, polling Core Audio at 0.25 s:

```
t= 41.69s  com.apple.avconferenced   idle -> RUNNING    call connects
t=155.47s  com.apple.avconferenced   RUNNING -> idle    hung up, seen within one poll
           (muted and unmuted in between: no edge at all)
```

- **The owner is `com.apple.avconferenced`.** `com.apple.FaceTime` never reports input. Your list has both, so the pattern fires correctly; I left the app bundle in for other macOS versions and replaced the guess in the comment with the measurement.
- **Muting does not release the input device.** That was the risk worth checking: if it did, the detector would report the meeting as over mid-conversation and cut the recording. It does not.
- **Hang-up is seen immediately**, so end detection is prompt and the normal grace path takes over.
- **The owner being a daemon has consequences** now written into the comment: the reported PID is the daemon's, so the recorder taps `avconferenced`, which is right because the call audio lives there, but the title matcher finds no window and falls back to the placeholder, and ParticipantReader has nothing to read. Also worth knowing that `avconferenced` serves AV conferencing generally, not FaceTime alone.
- **A FaceTime *link* joined in a browser is a different animal entirely**: that call is owned by `com.apple.WebKit.GPU`, so this channel does not see it, and it looks exactly like a browser meeting. It belongs to the browser detection channel, not here.

One incidental observation while measuring: a process without a bundle ID (plain `afplay`) shows up as running audio, and `systemAudioProcesses()` drops those. Not an issue for the four watched apps, but an app whose call audio runs in a bundle-less helper would be invisible to this channel by construction.

Full app suite green locally (minus `MenuBarIconSnapshotTests`, which fails on local rendering and passes in CI).

Issue #379, opened 2026-06-04
Crash (SIGABRT): uncaught NSException from installTapOnBus on mic device-change during recording (v0.6.0-rc1)


jhavez:
## Summary

MeetingTranscriber **crashed mid-recording** (SIGABRT) when an audio **input-device change fired during an active capture session**. The crash is an **uncaught Objective-C `NSException` thrown by `AVAudioNode.installTapOnBus`** while the mic engine was being restarted in response to a CoreAudio device-change notification. Because it's an ObjC exception (not a Swift `Error`), it cannot be caught with Swift `do/try/catch` → it propagates to `std::terminate` → `abort()`.

There is also a **secondary data-durability bug**: the in-progress recording's WAV header was never finalized, so ~22 minutes of captured audio appeared as a 0-byte/0-duration file until the RIFF/`data` chunk sizes were manually repaired (see below).

## Environment

| | |
|---|---|
| App version | **0.6.0-rc1** (beta cask) |
| macOS | **26.5 (25F71)** |
| Arch | Apple Silicon |
| Crash signal | `SIGABRT` / `EXC_CRASH` / "Abort trap: 6" |
| Faulting thread | `0` (`com.apple.main-thread`) |
| ASI | `abort() called` (`libsystem_c.dylib`) |
| Trigger | Input audio device change **during** an ~22-minute recording |

## Root cause (from crash report — both crashed-thread stack and lastExceptionBacktrace agree)

A CoreAudio property listener fires on device change → the mic engine is torn down and restarted → `installTapOnBus` throws because the tap is (re)installed with a format that no longer matches the new hardware node's output format. The throw is an ObjC `NSException`, which Swift cannot catch.

```
+[NSException raise:format:]                        (CoreFoundation)
AVAudioEngineImpl::InstallTapOnNode(...)            (AVFAudio)   ← throws here
-[AVAudioNode installTapOnBus:bufferSize:format:block:]
MicCaptureHandler.startEngine(deviceUID:)           (MeetingTranscriber)
MicCaptureHandler.executeRestart(deviceUID:)
MicCaptureHandler.handleDeviceChange()
MicCaptureHandler.handleEngineConfigChange()
closure #1 in MicCaptureHandler.installDeviceChangeListener()
thunk for @escaping (UInt32, UnsafePointer<AudioObjectPropertyAddress>) -> ()
invocation function for block in HALPropertyListener::Call(...)   (CoreAudio)
[...]

PR #550, opened 2026-07-30
fix(app): mic track is silent whenever another app activates voice processing (discrete channel layout downmix)

jianjettfu-oss:
## Symptom

The **mic track of a real call is digital silence**, while the app track records normally. Nothing is logged as an error — the only downstream sign is a transcript full of Whisper's silence hallucinations ("Thank you." repeated).

Observed on a 50-minute WeChat video call (macOS 26.1, MacBook Air, built-in mic):

```
21:29:18  Mic hardware format: 48000.0 Hz, 3ch      ← call starts, device flips 1ch → 3ch
21:29:24  Mic: filled 1 silent frames for a device-restart gap
   …      App audio RMS: -24…-33 dBFS    (fine)
   …      Mic RMS:       -48…-62 dBFS    (nothing)
22:19:23  Mic hardware format: 48000.0 Hz, 1ch      ← call ends, device back to 1ch
```

`ffmpeg silencedetect -70dB` on the resulting `_mic.wav`: **one unbroken silent region, 0 → 3005.1 s** of a 3022 s recording. It records audio again only in the last ~17 seconds, i.e. after the call released the device.

## Root cause

`AVAudioConverter`'s **implicit** downmix matrix is all-zero when the input format carries a *discrete* channel layout. `convert(to:error:)` returns success, sets no error, and writes silence.

A discrete layout is exactly what a microphone array reports — and the built-in mic switches from 1ch to a **3ch discrete array** as soon as any other app activates voice processing (any WeChat / FaceTime / Teams / Zoom call). `MicCaptureHandler` taps at the hardware format (correct, since #379) and relies on the converter to fold it to the 16 kHz mono WAV, so from the moment a call starts every sample written is zero.

Measured directly (48 kHz → 16 kHz mono, 0.5-amplitude sine on **every** input channel, so a correct downmix must yield ≈ 0.5):

| input format | peak out | |
|---|---|---|
| 1ch | 0.500 | ok |
| standard stereo | 0.500 | ok |
| discrete 2ch | **0.000** | silent |
| discrete 3ch | **0.000** | silent |
| discrete 4ch | **0.000** | silent |
| discrete 3ch, `channelMap = [0]` | 0.500 | ok |
| discrete 4ch, `channelMap = [0]` | 0.500 | ok |

## Fix

`MicChannelMap.downmixMap(for:)` decides when an explicit `converter.channelMap` is required:

- **mono** → no map (nothing to fold),
- **standard stereo** → no map, the implicit L+R fold is correct and keeps both channels,
- **discrete / unknown layout, or >2ch with no meaningful mono fold** → `[0]`, one real channel instead of silence.

For a mic array, channel 0 is the primary element, so selecting it is the sensible choice; averaging array elements would risk comb filtering from the inter-element delay.

`MicChannelMapTests` pins both halves: the policy, and the underlying converter behaviour itself (implicit → silence, explicit map → signal). If a future macOS fixes the implicit path, the pinning assertion fails and this workaround can be revisited deliberately rather than silently.

## A note on verification

The measurements above were produced by running the exact conversion path in a standalone Swift program on the affected machine — that evidence is solid. **The XCTest target itself I could not run**: this machine has Command Line Tools only (no Xcode), so `swift test` cannot resolve `XCTest`. Please treat the test file as unverified until CI runs it; I'd appreciate the workflow being approved on this PR.

Related: #547 (a second silent-capture failure, on the app-audio side, from a missing usage-description key).

### DIFF
--- a/tools/audiotap/Sources/MicCaptureHandler.swift
+++ b/tools/audiotap/Sources/MicCaptureHandler.swift
@@ -220,6 +220,19 @@ public class MicCaptureHandler: @unchecked Sendable {
                 standardFormatWithSampleRate: fileSampleRate, channels: 1,
             )! // swiftlint:disable:this force_unwrapping
             converter = AVAudioConverter(from: tapFormat, to: outputFormat)
+            // AVAudioConverter's IMPLICIT downmix writes SILENCE — with no
+            // error — when the input carries a discrete channel layout, which
+            // is exactly what a mic array reports. The built-in mic switches to
+            // a 3ch discrete array the moment another app activates voice
+            // processing (any WeChat / FaceTime / Teams call), so the mic track
+            // of a real call recorded 50 minutes of digital silence while the
+            // app track was fine. Selecting a channel explicitly restores it.
+            if let map = MicChannelMap.downmixMap(for: tapFormat) {
+                converter?.channelMap = map
+                logger.info(
+                    "Mic: discrete/multichannel layout — selecting channel 0 explicitly (implicit downmix yields silence)",
+                )
+            }
             logger.info(
                 "Mic: converting \(Int(tapFormat.sampleRate))Hz/\(tapFormat.channelCount)ch → \(Int(self.fileSampleRate))Hz/1ch",
             )
--- a/tools/audiotap/Sources/MicCaptureHandler.swift
+++ b/tools/audiotap/Sources/MicCaptureHandler.swift
@@ -211,26 +211,11 @@ public class MicCaptureHandler: @unchecked Sendable {
 
         converter = nil
         resampleRatio = fileSampleRate / tapFormat.sampleRate
-        // Convert to 16 kHz mono whenever the tap isn't already there — this
-        // covers resampling AND downmixing a multi-channel input device. Since
-        // the tap now matches the node's real channel count (issue #379), a
-        // 2ch device is captured as 2ch and folded to the mono WAV here.
-        if tapFormat.sampleRate != fileSampleRate || tapFormat.channelCount != 1 {
-            let outputFormat = AVAudioFormat(
-                standardFormatWithSampleRate: fileSampleRate, channels: 1,
-            )! // swiftlint:disable:this force_unwrapping
-            converter = AVAudioConverter(from: tapFormat, to: outputFormat)
-            // AVAudioConverter's IMPLICIT downmix writes SILENCE — with no
-            // error — when the input carries a discrete channel layout, which
-            // is exactly what a mic array reports. The built-in mic switches to
-            // a 3ch discrete array the moment another app activates voice
-            // processing (any WeChat / FaceTime / Teams call), so the mic track
-            // of a real call recorded 50 minutes of digital silence while the
-            // app track was fine. Selecting a channel explicitly restores it.
-            if let map = MicChannelMap.downmixMap(for: tapFormat) {
-                converter?.channelMap = map
+        if let setup = MicConverterFactory.make(tapFormat: tapFormat, fileSampleRate: fileSampleRate) {
+            converter = setup.converter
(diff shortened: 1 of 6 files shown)


jianjettfu-oss:
## End-to-end verification on the real device

Reproduced the failing condition without needing a live call: a helper process holds the built-in mic with `setVoiceProcessingEnabled(true)`, which flips it to the same 3ch discrete array a call produces (`system_profiler`: `Input Channels: 3`). Recorded a meeting through the app in that state, before and after the fix, same machine, same room, same silence:

| build | mic WAV mean | mic WAV max | |
|---|---|---|---|
| before | -91.0 dB | -91.0 dB | digital silence |
| after | -72.1 dB | -47.4 dB | real signal |

`-72 dB` matches the tap-side RMS the app logs for the same period (`-68.9…-70.4 dBFS`), i.e. the input now survives the conversion instead of being zeroed. The new branch is confirmed taken in the log:

```
Mic hardware format: 48000.0 Hz, 3ch
Mic: discrete/multichannel layout — selecting channel 0 explicitly (implicit downmix yields silence)
Mic: converting 48000Hz/3ch → 16000Hz/1ch
```

(The room was quiet and nobody was speaking, so the absolute level is low — the point is that it is no longer the -91 dB digital floor.)

pasrom:
Thanks, this one holds up. I reproduced both halves independently on macOS 26.5.2 before touching it:

- `AVAudioConverter`'s implicit downmix, 48 kHz to 16 kHz mono, 0.5 sine on every input channel: 1ch and plain stereo 0.500; DiscreteInOrder 2/3/4ch 0.000 with `convert` reporting success and no error; `channelMap [0]` restores 0.500.
- The trigger, without needing a call: a helper process holding the built-in mic with `setVoiceProcessingEnabled(true)` flips the device to 48 kHz 3ch `0x930003`, and the implicit downmix of that exact live format is 0.000. Back to 1ch when the helper exits.
- Your tests run green here (you mentioned you cannot run them): `MicChannelMapTests` 5/5, full `tools/audiotap` suite exit 0.

I pushed four commits on top of yours instead of sending review round-trips, since most of them need local test and lint runs to verify.

1. **Widen the policy.** The tag-family check only rescued Discrete and Unknown layouts, but these 2ch layouts are silent too, measured on 26.5.2: StereoHeadphones, MatrixStereo, MidSide, XY, Binaural, all 0.000 and all restored by `[0]`. MidSide and XY are microphone-pair configurations, so a stereo mic would have kept recording silence. The rule is now an allowlist: implicit path only for mono, layout-less stereo, and the plain `Stereo` tag.
2. **Extract `MicConverterFactory`.** The channel-map decision was only reachable through a live `AVAudioEngine`, and `MicCaptureHandler.swift` had grown to 612 lines against the 600-line cap that `swiftlint --strict` fails the build on.
3. **A test at that seam.** Worth spelling out: with the `converter.channelMap = map` assignment neutralised, `MicChannelMapTests` still passes 8/8 while the new tests fail 2/4 (map `[-1]` instead of `[0]`, converted peak 0.0 instead of 0.5). The policy tests alone could not have caught this bug.
4. **Fix the `TapFormatResolver` doc comment** that claimed the converter folds the tap format to mono. That assumption is what produced the silent track.

Also rebased onto main, so the branch was force-updated. Gates on the result: SwiftFormat (the pinned 0.61.1) 0/390, `swiftlint --strict` clean, audiotap suite exit 0, app package builds.

Two open points, both minor. "Channel 0 is the primary element" is unverified: per-channel RMS on the 3ch array in a quiet room was -79.1 / -77.4 / -75.1 dBFS, so all three elements carry comparable signal and `[0]` is a safe pick, but the "primary" claim is not established. And named surround layouts do fold correctly yet now get `[0]` anyway; no mic input reports one, so I left it that way and said so in the code comment.

PR #570, opened 2026-08-01
fix(build): drop the deprecated string form from the cask requirement

pasrom:
Homebrew 6.0.x deprecates `depends_on macos: ">= :sonoma"` and prints a warning on every operation that touches the cask, telling users to report it to this tap. It fires today on a plain `brew upgrade`.

The bare symbol means the same thing on modern brew. Verified against the live cask before changing anything here: with `depends_on macos: :sonoma`, `brew info --cask meeting-transcriber` still reports `Required: macOS >= 14` and the warning is gone.

Note these two files are only templates. The casks users actually install live in the tap repo, and the release workflow rewrites just the version and sha256 lines there, so the same one-line change has to land in the tap to silence the warning. This keeps the template from re-introducing it when it seeds a cask file the tap does not have yet.

### DIFF
--- a/Casks/meeting-transcriber.rb
+++ b/Casks/meeting-transcriber.rb
@@ -7,7 +7,7 @@ cask "meeting-transcriber" do
   desc "Auto-transcribe and summarize meetings"
   homepage "https://github.com/pasrom/meeting-transcriber"
 
-  depends_on macos: ">= :sonoma"
+  depends_on macos: :sonoma
 
   conflicts_with cask: "meeting-transcriber@beta"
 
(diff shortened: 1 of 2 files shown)


pasrom:
Closing unmerged: this change is already tracked as deliberately deferred, and I opened this without checking that.

Homebrew 5.1.11 introduced the deprecation and changed the meaning of the bare symbol in the same release: from "exactly Sonoma" to "Sonoma or newer". On brews older than that, `depends_on macos: :sonoma` still means `==`, so this one-line change would block fresh installs on Sequoia and Tahoe for those users. The current string form parses as `>=` everywhere.

Checked today against the 30-day Homebrew analytics: 9.3% of clients are still older than 5.1.11 (1.98M of 21.26M events), essentially unchanged from a week ago. The deprecation warning is cosmetic and brew exits 0; a blocked install would not be. Worth revisiting once that share is near zero.

PR #573, opened 2026-08-04
test(ci): run the browser-meeting E2E lane against Brave, Edge and Chromium

pasrom:
## Summary

Follow-up to #572 (multi-browser detection). The #503 browser-meeting E2E lane (`scripts/e2e-browser.sh`) only exercised Chrome. Since Brave, Edge and Chromium now all detect under the shared "Google Chrome" identity via the same `WebRTC has active PeerConnections` assertion, the same fixture lane proves each fork end to end (detection → RPC consent → non-silent CATap capture) with only the launched `.app` differing.

## Changes

- `scripts/e2e-browser.sh`: new `--browser chrome|brave|edge|chromium` flag that resolves the app path and diagnostic label. The `pendingConsentApp` assertion stays `"Google Chrome"` (the shared identity, correct for every fork). Per-run `/tmp` artifacts are browser-suffixed for post-mortem uploads.
- `.github/workflows/e2e-browser.yml`: Chrome stays the primary (builds + runs on every event); Brave, Edge and Chromium run nightly/dispatch as best-effort steps that reuse the Chrome-built bundle via `--no-build`. They stay off labeled PRs (one primary browser keeps PR runs fast) and are `continue-on-error`: a fork not installed on the runner, or a fork-specific flake, is a signal to check, not a merge blocker on this already non-gating canary.

## Runner prerequisite

Brave, Microsoft Edge and Chromium are installed under `~/Applications` on the self-hosted mini (no sudo needed; a missing one just fails its best-effort step). Their `CFBundleExecutable` names (`Brave Browser`, `Microsoft Edge`, `Chromium`) match the detection process names #572 added.

## Validation

Dispatched against this branch on the mini to prove each fork live before merge (results referenced in the PR thread). This is the field measurement the #572 caveat pointed at.

### DIFF
--- a/.github/workflows/e2e-browser.yml
+++ b/.github/workflows/e2e-browser.yml
@@ -1,8 +1,10 @@
 name: E2E (Browser Meeting)
 
-# Live browser-meeting detection + capture E2E (issue #503). Drives Chrome +
-# a local WebRTC-tone fixture through the power-assertion detector, answers the
-# consent prompt over RPC, and asserts the record-only _app.wav is non-silent.
+# Live browser-meeting detection + capture E2E (issue #503). Drives a Chromium
+# browser + a local WebRTC-tone fixture through the power-assertion detector,
+# answers the consent prompt over RPC, and asserts the record-only _app.wav is
+# non-silent. Chrome is the primary (builds + runs on every event); Brave, Edge
+# and Chromium run nightly/dispatch as best-effort forks reusing that bundle.
 # Background: CLAUDE.md > E2E Architecture. Driver: scripts/e2e-browser.sh.
 #
 # NON-GATING CANARY: not referenced by any required check and not listed in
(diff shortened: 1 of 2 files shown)


pasrom:
Validated live on the self-hosted Apple Silicon mini — all four Chromium forks pass the full chain (launch → WebRTC assertion held under the browser's own process name → detection under the shared "Google Chrome" identity → RPC consent → non-silent CATap capture):

| Browser | Version | Result |
|---|---|---|
| Google Chrome | 150.0.7871.188 | PASS |
| Brave Browser | 151.1.93.129 | PASS |
| Microsoft Edge | 151.0.4129.59 | PASS |
| Chromium | 153.0.7991.0 | PASS |

Run: https://github.com/pasrom/meeting-transcriber/actions/runs/30919770223 (overall success; the fork steps are not continue-on-error, so these are honest conclusions).

Two things the first mini runs surfaced (both now fixed in this PR / runner setup):
- The forks need `--disable-features=WebRtcHideLocalIpsWithMdns`: without it the in-page loopback PeerConnection never reaches `connected` on a fresh profile (the browser holds only a "Playing audio" assertion, never "WebRTC has active PeerConnections"), so detection correctly does not fire. Chrome happened to connect without it; Brave/Chromium did not.
- The three fork steps are independent (`!cancelled()`), so one failing no longer skips the rest.

This closes out the field-verification caveat from the multi-browser detection change: the Brave/Edge/Chromium process names and their WebRTC assertion are now measured end to end, not inferred.

Issue #524, opened 2026-07-20
[Bug] App Sound not recording (only my mic)


joshstrange:
### App version

0.6.0-rc6 (6cc5204) · Homebrew

### Build variant

Homebrew (beta / RC)

### macOS version

15.7.3 (24G419)

### What happened?

Every meeting I've recorded so far records me but the other track is silent. I get the notification that it doesn't hear anything but there is nothing I can do about it.

The only thing I can think of is I use SoundSource but I don't do anything special with Zoom or redirect. I just don't know if it installs something that causes Meeting Transcriber to not be able to grab the audio.

I love having my side of the conversation but of course it'd be better to have both/all sides. Thank you!

### Steps to reproduce

1. Start a zoom meeting with 2+ participants
2. Both people talk
3. Only the host's voice is recorded

### Diagnostic log

[MeetingTranscriber-diagnostics-1784582035.log](https://github.com/user-attachments/files/30204503/MeetingTranscriber-diagnostics-1784582035.log)

pasrom:
Thanks again for the log. It let me pin this down further, and I think it is very likely a permission issue rather than SoundSource.

Your log shows the process tap on Zoom being created successfully and audio callbacks running, but every buffer is silent for the whole meeting (all three sessions). On macOS, the Core Audio process tap that captures the other participants is gated by a permission, and here is the catch: the build you are on (0.6.0-rc6) shipped **without** the "Audio Recording" usage string, which means macOS silently denies the tap and falls back to your Screen Recording grant. If Screen Recording is not enabled for Meeting Transcriber, the tap then returns success but delivers pure silence, with no error and nothing in the log. That matches your log exactly, and it does not require SoundSource to explain it. (Meetings still get auto-detected because detection uses a separate mechanism that does not need Screen Recording.)

The next build fixes this on two fronts: it adds the proper Audio Recording permission prompt (one you can actually grant), and it now asks macOS for Screen Recording at record time so the app shows up in that list with a switch you can flip. When the next release is out, please update and, on your first recording after updating, allow the Audio Recording prompt and enable Meeting Transcriber under System Settings, Privacy & Security, Screen & System Audio Recording. That should bring the other side of the call back.

If it is still silent after granting both, then we look at SoundSource: quit it completely (Quit from its menu bar icon) and do a short test recording. I will ping this issue when the build with the fix is available.

pasrom:
[...]

Issue #84, opened 2026-04-02
New Microsoft Teams (com.microsoft.teams2) not detected


amit-math:
## Summary
MeetingTranscriber does not detect or auto-record meetings when 
using the new Microsoft Teams app (2026 version).

## Environment
- macOS 15.x on Apple M5
- MeetingTranscriber installed via Homebrew (brew tap pasrom/meeting-transcriber)
- Microsoft Teams Version 26072.605.4546.510
- Teams toggle is enabled in MeetingTranscriber settings
- All permissions granted (Screen Recording, Microphone, Accessibility)

## Problem
The menu bar icon stays idle when joining a Teams call — no animation, 
no recording starts. Auto-detection does not trigger.

## Root Cause (likely)
The new Microsoft Teams uses a different bundle ID than the classic version:

- **New Teams:** `com.microsoft.teams2` (process name: `MSTeams`)
- **Old Teams:** `com.microsoft.teams` (process name: `Teams`)

The app name reported by the system is "Microsoft Teams" and the 
executable is `MSTeams` located at:
/Applications/Microsoft Teams.app/Contents/MacOS/MSTeams

Output from `lsappinfo`:
"Microsoft Teams" ASN:0x0-0x70070:
bundleID="com.microsoft.teams2"
executable path="/Applications/Microsoft Teams.app/Contents/MacOS/MSTeams"
pid = 3444 type="Foreground" flavor=3 Arch=ARM64

## Expected Behavior
MeetingTranscriber should detect active calls in the new Teams app 
and auto-start recording, just as it does for Zoom and Webex.

## Workaround
Manual file import via ⌘P works for transcription and diarization.
Only the auto-detection is affected.

pasrom:
Thanks for the report! A couple of questions:

1. **Which MeetingTranscriber version are you using?** (Check via the menu bar icon → About, or `brew info meeting-transcriber`)
2. **Did you click "Start Watching" in the menu bar dropdown?** Auto-detection only runs while watching is active — without it, the menu bar icon stays idle regardless of active meetings.

If watching was active, could you also try a **full TCC permission reset**? We've seen cases ([#79](https://github.com/pasrom/meeting-transcriber/issues/79)) where permissions appear granted but aren't properly wired:

1. **Screen Recording:** Remove via "−" button
2. **Microphone:** Toggle off
3. **Accessibility:** Remove via "−" button

Then quit the app, manually re-add it in each section via "+", and relaunch.

[...]

PR #579, opened 2026-08-04
fix(app): tap Safari's WebKit helpers so app audio is captured (#524)

duncanwood:
Fixes the Safari case of #524. Root-cause detail in [this comment](https://github.com/pasrom/meeting-transcriber/issues/524#issuecomment-5185500571).

## The bug

Safari app-audio capture records digital silence. The tap is created successfully and runs for the whole recording — it is simply pointed at a process that never produces sound.

```
before:  App audio tap: 1 PID(s) [Safari(692)]
after:   App audio tap: 2 PID(s) [Safari(692), com.apple.WebKit.GPU(21403)]
```

`resolveTapPIDs` builds its set with `ProcessTreeEnumerator.pidsRooted(in:)`, which matches executables by bundle-path prefix. Correct for Chrome and Electron, whose helpers live inside the app bundle. It cannot work for Safari — WebKit runs its GPU and WebContent services as XPC outside the app:

```
/System/Volumes/Preboot/Cryptexes/Incoming/OS/System/Library/Frameworks/WebKit.framework/
  Versions/A/XPCServices/com.apple.WebKit.GPU.xpc/Contents/MacOS/com.apple.WebKit.GPU
```

So rooting at `Safari.app` yields Safari's main process alone, while `kAudioProcessPropertyIsRunningOutput` shows the audio coming from `com.apple.WebKit.GPU`.

This also explains reports of *"it recorded for a moment, then went silent"*: early in a call audio still flows through the main process, then WebKit hands it to the GPU service and the tap is left behind. In my case that was 40 seconds of audio followed by 59 minutes of silence.

## The fix

`ProcessResponsibility.tapPIDs` unions in the processes macOS attributes to the same responsible app — `responsibility_get_pid_responsible_for_pid`, the relationship TCC consults when a WebKit service touches the microphone — filtered to those CoreAudio has a process object for.

The filter is load-bearing. Unfiltered, Safari's responsibility tree is ~28 PIDs including a Metal shader compiler and a speech-synthesis service, and tapping that set **also** produced zero bytes. Filtered, it is exactly the audio clients (2 here).

**Non-browser apps are unaffected:** when responsibility is unavailable or adds nothing, the bundle-derived set is returned unchanged. The private symbol is resolved via `dlsym`, so a missing symbol degrades to today's behaviour rather than failing.

## Verification

Live, unmuting mid-recording:

```
muted:    App audio RMS -110.5 dBFS
unmuted:  App audio RMS -42.4, -43.3, -41.9, -33.2 dBFS
```

Sustained capture confirmed over several minutes on macOS 26.5.2 / Safari 26.x, well past the point where it previously went silent.

Tests: 194 audiotap, 200 app. `swiftlint --strict` clean. `DualSourceRecorder.swift` stays within the 600-line limit (the logic lives in `ProcessResponsibility`).

## One trap worth flagging

`proc_listallpids` returns a PID **count**, not a byte count. Dividing by `sizeof(pid_t)` truncates the list to a quarter and silently drops whichever processes fall past the cut — which cost me an hour, since the dropped ones were Safari and the audio helper. `LiveRunningPIDsTests` exercises the real enumeration for that reason.

## Not included

I also have a change that re-resolves the tap PID set periodically and rebuilds when it changes, for the case where a helper is recycled mid-call. I left it out: I could not reproduce a recycle, so it would be unjustified surface area. Happy to open it separately if you think it is worth having.

Related: #79 looks like the same shape for Teams, though I have not tested it.

### DIFF
--- a/tools/audiotap/Sources/ProcessResponsibility.swift
+++ b/tools/audiotap/Sources/ProcessResponsibility.swift
@@ -1,0 +1,144 @@
+import CoreAudio
+import Darwin
+import Foundation
+
+/// Groups a helper process with the application macOS holds *responsible* for it.
+///
+/// `ProcessTreeEnumerator` groups processes by executable-path prefix, which is
+/// right for apps whose helpers live inside their own bundle (Chrome, Electron)
+/// and cannot work for Safari: WebKit runs its GPU and WebContent services as
+/// XPC under `/System/Library/Frameworks/WebKit.framework`, so rooting at
+/// `Safari.app` returns Safari's own processes and never the one producing the
+/// call audio.
+///
+/// Responsibility is the relationship macOS itself uses to attribute a helper's
+/// behaviour to an app — it is what TCC consults when a WebKit service touches
+/// the microphone. Measured on macOS 26.5:
+///
+///     pid 2032 [com.apple.WebKit.GPU] -> responsible 707 [com.apple.Safari]
+///
+/// `responsibility_get_pid_responsible_for_pid` is not in a public header, so it
+/// is resolved with `dlsym`; a missing symbol degrades to "no opinion" and the
+/// caller keeps its bundle-derived set.
+public enum ProcessResponsibility {
+    private typealias ResponsibleFn = @convention(c) (pid_t) -> pid_t
+
+    /// Resolved lazily, and deliberately absent from the App Store build: the
+    /// symbol is private API, and shipping the string invites an automated
+    /// review flag. Every caller already degrades to the bundle-derived set
+    /// when this is nil, so the sandboxed variant simply does not get the
+    /// Safari fix — a conscious tradeoff.
+    private static let responsibleFn: ResponsibleFn? = {
+        #if APPSTORE
+            nil
+        #else
+            guard let handle = dlopen(nil, RTLD_NOW),
+                  let sym = dlsym(handle, "responsibility_get_pid_responsible_for_pid")
+            else { return nil }
+            return unsafeBitCast(sym, to: ResponsibleFn.self)
+        #endif
+    }()
+
+    /// The app responsible for `pid`, or 0 when unavailable. Identity for a
+    /// process that is its own responsible party (a normal app).
+    public static func responsiblePID(of pid: pid_t) -> pid_t {
+        guard let fn = responsibleFn else { return 0 }
+        let owner = fn(pid)
+        return owner > 0 ? owner : 0
+    }
+
+    /// PIDs that CoreAudio has a process object for — every process that could
+    /// meaningfully be tapped, typically a few dozen.
+    public static func audioCapablePIDs() -> Set<pid_t> {
+        var address = AudioObjectPropertyAddress(
+            mSelector: kAudioHardwarePropertyProcessObjectList,
+            mScope: kAudioObjectPropertyScopeGlobal,
+            mElement: kAudioObjectPropertyElementMain,
+        )
+        var size: UInt32 = 0
+        guard AudioObjectGetPropertyDataSize(
+            AudioObjectID(kAudioObjectSystemObject), &address, 0, nil, &size,
+        ) == noErr, size > 0 else { return [] }
+
+        var objects = [AudioObjectID](
+            repeating: 0, count: Int(size) / MemoryLayout<AudioObjectID>.size,
+        )
+        guard AudioObjectGetPropertyData(
+            AudioObjectID(kAudioObjectSystemObject), &address, 0, nil, &size, &objects,
+        ) == noErr else { return [] }
+
+        var pids = Set<pid_t>()
+        for object in objects {
+            var pidAddress = AudioObjectPropertyAddress(
+                mSelector: kAudioProcessPropertyPID,
+                mScope: kAudioObjectPropertyScopeGlobal,
+                mElement: kAudioObjectPropertyElementMain,
+            )
+            var pid: pid_t = -1
+            var pidSize = UInt32(MemoryLayout<pid_t>.size)
+            if AudioObjectGetPropertyData(object, &pidAddress, 0, nil, &pidSize, &pid) == noErr,
+               pid > 0 {
+                pids.insert(pid)
+            }
+        }
+        return pids
+    }
+
+    /// Audio-capable processes that macOS attributes to `owner`, including
+    /// `owner` itself.
+    ///
+    /// Iterates the audio-capable set (a few dozen) rather than every live
+    /// process (~1900 here), because the intersection is the answer either way
+    /// and this asks the kernel for responsibility a few dozen times instead of
+    /// a few thousand per recording start.
+    ///
+    /// `audioCapable` and `responsibleFor` are injection seams so the grouping
+    /// is verifiable without real processes.
+    public static func audioPIDsResponsible(
+        to owner: pid_t,
+        audioCapable: () -> Set<pid_t> = audioCapablePIDs,
+        responsibleFor: ((pid_t) -> pid_t)? = nil,
+    ) -> [pid_t] {
+        guard owner > 0 else { return [] }
+        // No symbol and no injected lookup means no opinion to offer; returning
+        // empty keeps the caller on its bundle-derived set rather than silently
+        // widening the tap.
+        if responsibleFor == nil, responsibleFn == nil {
+            return []
+        }
+        let lookup = responsibleFor ?? { responsiblePID(of: $0) }
+        return audioCapable().filter { $0 == owner || lookup($0) == owner }.sorted()
+    }
+
+    /// The PID set an app-audio tap should cover, given the bundle-derived set
+    /// the caller already computed.
+    ///
+    /// Safari's audio is produced by WebKit XPC services that live outside
+    /// `Safari.app`, so the bundle-derived set is Safari's main process alone —
+    /// which never emits sound. Measured: `App audio tap: 1 PID(s)
+    /// [Safari(692)]` while the audio came from `com.apple.WebKit.GPU(21403)`.
+    /// The tap was valid and recorded an hour of silence.
+    ///
+    /// Only widens when `rootPID` is its OWN responsible party, i.e. a normally
+    /// launched app. A process started from a shell or script is attributed to
+    /// the launcher, and widening on that would tap whatever audio-capable
+    /// processes happen to share it — a terminal that played a system sound, a
+    /// media tool from the same shell. GUI-launched meeting apps are always
+    /// their own responsible party, so the guard costs nothing there.
+    public static func tapPIDs(
+        rootPID: pid_t,
+        bundleDerived: [pid_t],
+        responsibleFor: ((pid_t) -> pid_t)? = nil,
+        audioCapable: () -> Set<pid_t> = audioCapablePIDs,
+    ) -> [pid_t] {
+        let lookup = responsibleFor ?? { responsiblePID(of: $0) }
+        let owner = lookup(rootPID)
+        guard owner != 0, owner == rootPID else { return bundleDerived }
+        let byOwner = audioPIDsResponsible(
+            to: owner, audioCapable: audioCapable, responsibleFor: responsibleFor,
+        )
+        guard !byOwner.isEmpty else { return bundleDerived }
+        var seen = Set<pid_t>()
+        return (bundleDerived + byOwner).filter { seen.insert($0).inserted }
+    }
+}
--- a/tools/audiotap/Tests/ProcessResponsibilityTests.swift
+++ b/tools/audiotap/Tests/ProcessResponsibilityTests.swift
@@ -1,0 +1,122 @@
+@testable import AudioTapLib
+import XCTest
+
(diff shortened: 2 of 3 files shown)


tools/audiotap/Sources/ProcessResponsibility.swift
```diff
@@ -0,0 +1,152 @@
...
+    /// So union in the processes macOS attributes to the same responsible app,
+    /// filtered to those CoreAudio has a process object for. The filter is
+    /// load-bearing: unfiltered, Safari's responsibility tree is ~28 PIDs
+    /// including a Metal shader compiler and a speech-synthesis service, and
+    /// tapping that set also produced zero bytes.
+    ///
+    /// Returns `bundleDerived` unchanged when responsibility is unavailable or
+    /// adds nothing, so non-browser apps keep exactly today's behaviour.
+    public static func tapPIDs(rootPID: pid_t, bundleDerived: [pid_t]) -> [pid_t] {
+        let owner = responsiblePID(of: rootPID)
```
pasrom: **Correctness, worth fixing before merge.** `tapPIDs` widens unconditionally, using `responsiblePID(of: rootPID)` even when that is not the app itself. For a process launched from a terminal or a script the responsible party is the *launcher*: measured here, a shell-launched process attributes to the launcher, with 207 live PIDs sharing that owner. Real meeting apps are GUI-launched and are their own responsible party, so this is a no-op for them; but a script-launched root (dev runs, the e2e recording lane) would widen to whatever audio-capable processes share that launcher. On my machine that set happened to be empty right now, so this is latent rather than always-on, but any audio-capable sibling of the launcher (a terminal that has played a system sound, a media tool started from the same shell) would then be tapped instead of just the target app. Safari itself is its own responsible party (your `2032 -> 707` implies `707 -> 707`), so a one-line guard keeps the fix and closes the failure mode:

```swift
let owner = responsiblePID(of: rootPID)
guard owner == rootPID else { return bundleDerived }
```

tools/audiotap/Sources/ProcessResponsibility.swift
```diff
@@ -0,0 +1,152 @@
...
+    /// Live PID snapshot via `proc_listallpids`.
+    ///
+    /// NOTE: `proc_listallpids` returns the number of **PIDs**, not the number
+    /// of bytes — unlike most `proc_*` calls, and unlike what the parameter
+    /// name `buffersize` suggests. Dividing the result by `sizeof(pid_t)` here
+    /// truncated the list to a quarter of the running processes, which silently
+    /// dropped Safari and the very helper carrying the call audio, so the tap
+    /// was built without the process making sound. Measured: 819 returned,
+    /// misread as 204.
+    public static func liveRunningPIDs() -> [pid_t] {
```
pasrom: This re-implements `ProcessTreeEnumerator.liveRunningPIDs()` (same module), which lists the same PIDs via `proc_listpids`. The two use different libproc calls with opposite return-unit conventions (bytes vs. count), which is exactly the footgun the note above documents. Cleanest fix: don't enumerate all PIDs at all. `tapPIDs` immediately intersects with `audioCapablePIDs()` (a few dozen entries), so invert the filter:

```swift
let byOwner = audioCapablePIDs().filter {
    $0 == owner || responsiblePID(of: $0) == owner
}
```

Same result set, deletes `liveRunningPIDs` and its tests, and turns ~1900 responsibility syscalls per recording start into a few dozen.

tools/audiotap/Sources/ProcessResponsibility.swift
```diff
@@ -0,0 +1,152 @@
...
+///
+/// `responsibility_get_pid_responsible_for_pid` is not in a public header, so it
+/// is resolved with `dlsym`; a missing symbol degrades to "no opinion" and the
+/// caller keeps its bundle-derived set.
+public enum ProcessResponsibility {
+    private typealias ResponsibleFn = @convention(c) (pid_t) -> pid_t
+
+    private static let responsibleFn: ResponsibleFn? = {
+        guard let handle = dlopen(nil, RTLD_NOW),
+              let sym = dlsym(handle, "responsibility_get_pid_responsible_for_pid")
```
pasrom: `responsibility_get_pid_responsible_for_pid` via `dlsym` is a private symbol, and AudioTapLib links into the sandboxed build with no `#if` guard, so the symbol string ships and runs in the App Store variant, which is a plausible automated-review flag. The code already degrades cleanly when the symbol is absent, so `#if APPSTORE { responsibleFn = nil }` covers it (the Safari fix then does not ship in that variant, a conscious tradeoff).

tools/audiotap/Sources/ProcessResponsibility.swift
```diff
@@ -0,0 +1,152 @@
...
+    /// Bundle-path enumeration cannot describe Safari: its audio is produced by
+    /// WebKit XPC services under `/System/Library/Frameworks`, so rooting at
+    /// `Safari.app` yields Safari's main process ALONE — which never emits
+    /// sound. Measured: `App audio tap: 1 PID(s) [Safari(692)]` while the audio
+    /// came from `com.apple.WebKit.GPU(21403)`. The tap was valid and recorded
+    /// an hour of silence.
+    ///
+    /// So union in the processes macOS attributes to the same responsible app,
+    /// filtered to those CoreAudio has a process object for. The filter is
+    /// load-bearing: unfiltered, Safari's responsibility tree is ~28 PIDs
```
pasrom: Worth re-verifying this rationale. The tap-creation path already drops any PID with no CoreAudio process object, so the filtered and unfiltered sets should translate to the same tapped objects. The filter is harmless to keep, but the "unfiltered ~28-PID set produced zero bytes" result looks like it may have been confounded (TCC/timing) rather than caused by the extra PIDs.

tools/audiotap/Sources/ProcessResponsibility.swift
```diff
@@ -0,0 +1,152 @@
...
+    /// including a Metal shader compiler and a speech-synthesis service, and
+    /// tapping that set also produced zero bytes.
+    ///
+    /// Returns `bundleDerived` unchanged when responsibility is unavailable or
+    /// adds nothing, so non-browser apps keep exactly today's behaviour.
+    public static func tapPIDs(rootPID: pid_t, bundleDerived: [pid_t]) -> [pid_t] {
```
pasrom: `tapPIDs` (the function `resolveTapPIDs` actually calls) has no test seam; only the pure `pidsResponsible` helper is covered. I wrote a live regression test grounded on CoreAudio's `IsRunningOutput` (so it targets the process actually producing Safari's audio) plus a deterministic composition test, and added an injection seam here. Happy to send those over.

tools/audiotap/Tests/ProcessResponsibilityTests.swift
```diff
@@ -0,0 +1,74 @@
+@testable import AudioTapLib
+import XCTest
+
+/// Covers the responsible-PID grouping that keeps a browser's call audio tapped
+/// after WebKit recycles the helper carrying it (issue #524).
+///
+/// The regression these guard: on 2026-08-04 a 62-minute interview captured 40
+/// seconds of the other participant and then digital silence, because the tapped
+/// PID set was resolved once at start and the helper was later replaced.
```
pasrom: Shipped code resolves the tap PID set once in `start()`, with no re-resolution, so `testPicksUpAHelperThatAppearsAfterRecordingStarted` exercises a loop that does not exist in production and passes without guarding anything shipped. Leaving the periodic re-resolve out is the right call (no repro, and recording Safari is manual today so the helper is already an audio client at start). Suggest scoping this wording to start-time coverage and tracking the recycle case as an explicit follow-up.

tools/audiotap/Tests/LiveRunningPIDsTests.swift
```diff
@@ -0,0 +1,32 @@
...
+/// and where the truncation bug lived: `proc_listallpids` returns a PID count,
+/// not a byte count, and dividing by `sizeof(pid_t)` silently dropped three
+/// quarters of the process list (819 read as 204), taking Safari and the
+/// audio-carrying helper with it.
+final class LiveRunningPIDsTests: XCTestCase {
+    func testEnumeratesEveryLiveProcessNotAFraction() {
+        let pids = ProcessResponsibility.liveRunningPIDs()
+        // A live macOS session runs many hundreds of processes; a truncated
+        // read lands around a quarter of that.
+        XCTAssertGreaterThan(pids.count, 300, "process list looks truncated")
```
pasrom: This passes even against the exact bug it documents: `proc_listallpids` returns newest-first, so a divide-by-4 truncation keeps the freshly spawned test process, and on a busy machine a quartered count is still > 300 (I saw ~1900 live processes here, so ~480 after truncation). Assert against the API's own sizing call instead of a fixed number. If you invert the filter in `tapPIDs`, `liveRunningPIDs` and these tests go away entirely.

pasrom (changes requested):
Thanks for this. The diagnosis is right and well-evidenced: bundle-path enumeration cannot see WebKit's GPU/WebContent XPC services under `/System/Library/Frameworks`, so rooting at `Safari.app` only ever returns Safari's own processes. I reproduced the mechanism live: with Safari playing audio, the only process reporting `IsRunningOutput=true` is a `com.apple.WebKit.GPU` that macOS attributes to Safari and lives outside the bundle, so the old set records silence and the responsibility union recovers it.

One correctness fix worth making before merge (the `owner == rootPID` guard) plus a few things worth tightening, left inline. With the guard and the reuse cleanup this is a good fix for the from-start Safari-silence case.

duncanwood:
Thanks — this was a genuinely useful review. All seven points addressed in `eb0fb4b`.

**The launcher guard.** You are right and this was a real bug I introduced. Added `guard owner == rootPID`. Verified Safari is unaffected: `rootPID 692 -> responsible 692`, guard passes, tap set still `[692 Safari, 21403 com.apple.WebKit.GPU]`.

**The duplicate enumerator.** Took your inversion — iterate `audioCapablePIDs()` and ask responsibility for those, instead of enumerating every live PID and intersecting. Same result, and it deletes `liveRunningPIDs` along with the libproc return-unit footgun entirely rather than documenting it. ~1900 syscalls per recording start become a few dozen.

**Private symbol.** `#if APPSTORE` -> `nil`, so the string does not ship in the sandboxed variant.

**You were right to challenge the filter rationale, and my claim was wrong.** The "unfiltered ~28-PID set produced zero bytes" measurement was confounded: I took it while the since-fixed `proc_listallpids` truncation bug was still live, so that set was missing the audio helper for an entirely unrelated reason. The tap path does drop objectless PIDs, exactly as you say. I have removed the claim; the filter stays for cost (it is what makes the inversion cheap), and the comment now says that instead of asserting a causal result I cannot support.

**Tests.** `tapPIDs` now has injection seams and direct coverage: the launcher guard, unavailable-responsibility fallback, dedupe, and root-first ordering (#84). Both tests you flagged are gone — the recycle test guarded a loop deliberately not in this PR, and `LiveRunningPIDsTests` did pass against its own bug (newest-first ordering keeps the freshly spawned test process, and a quartered count still clears a fixed >300 bound on a busy machine). That one was a good catch; it was worse than no test.

Yes please on the live regression test grounded on `IsRunningOutput` — that targets the process actually producing audio, which is the property that matters and the one my composition tests only approximate. Happy to take it as a commit here or fold it in myself, whichever suits.

`swiftlint --strict` clean repo-wide, swiftformat clean (using the pinned 0.61.1 rules — my local 0.62 flags pre-existing files, so I left those alone). 195 audiotap + 78 app recorder tests pass.

On the recycle case: still no reproduction here, so it stays out. Tracking it as a follow-up rather than speculative surface area, as you suggested.

pasrom (approved):
All seven points are addressed, and I verified the result rather than the description.

- **Launcher guard.** `owner == rootPID` is in place, and I confirmed live it does not re-break Safari: Safari is its own responsible party (rootPID resolves to itself), so the guard passes and the tap set still recovers the out-of-bundle `com.apple.WebKit.GPU` helper. GUI-launched apps are always their own responsible party, so the guard is a no-op for real targets.
- **Duplicate enumerator removed.** `liveRunningPIDs` is gone and the filter is inverted over `audioCapablePIDs()`, so a recording start asks the kernel for responsibility a few dozen times instead of a few thousand.
- **Private symbol.** `responsibleFn` is nil under `#if APPSTORE`, so the string does not ship in the sandboxed variant; that build compiles and simply does not get the Safari fix.
- **Corrected claim.** The unverified "unfiltered set produced zero bytes" comment is removed; the audio-capable filter is now documented as a cost optimization, not correctness.
- **Tests.** `tapPIDs` gained injection seams and is covered directly, including the launcher guard, the unavailable-responsibility fallback, dedupe, and root-first ordering. The two misleading tests (the recycle loop that does not exist in this PR, and the false-green enumeration test) are gone.

Verified locally at this head: 195 audiotap tests pass, `swiftlint --strict` is clean, and the APPSTORE variant compiles.

One note for merge, not a review blocker: the branch is behind `main` and required checks are strict, so it needs to be brought up to date before the required CI (lint, analyze, test and build for both variants) can gate it green.

Thanks for the thorough turnaround.

PR #583, opened 2026-08-06
fix(app): negotiate max_completion_tokens when max_tokens is rejected

### LINKED ISSUE #582: [Bug] Protocol generation fails with HTTP 400 on GPT-5-family models (max_tokens vs max_completion_tokens)
### App version

0.6.0 · Homebrew

### Build variant

Homebrew (stable)

### macOS version

26.5.2 (25F84)

### What happened?

Protocol generation fails with HTTP 400 for every recording when the
OpenAI-compatible provider points at a current OpenAI model. Transcription and
diarization work fine, so `.txt` transcripts are produced but no `.md` protocol is
ever written.

`OpenAIProtocolGenerator` sends the output cap as `max_tokens`. The OpenAI Chat
Completions API renamed that parameter to `max_completion_tokens`, and GPT-5-family
models now reject the old name outright rather than accepting it as an alias:

```
[OpenAIProtocolGenerator] openai_http_error status=400
endpoint=https://<resource>.services.ai.azure.com/openai/v1/chat/completions
body={"error":{"message":"Unsupported parameter: 'max_tokens' is not supported with
this model. Use 'max_completion_tokens' instead.","type":"invalid_request_error",
"param":"max_tokens","code":"unsupported_parameter"}}

[PipelineQueue] [ce9472e5] protocol_generation_failed error=HTTP 400: ...
```

Reproduced against Azure AI Foundry with a `gpt-5.4` deployment, on four consecutive
recordings (three auto-detected meetings and one batch import).

Note this cannot simply be renamed: LM Studio and Ollama accept only `max_tokens`, so
both spellings need to be supported.

Workaround for anyone hitting this: switch to a model that still accepts `max_tokens`
(e.g. Grok, or an older GPT deployment).

### Steps to reproduce

1. Settings → Output → LLM provider: OpenAI-Compatible API
2. Endpoint: `https://<resource>.services.ai.azure.com/openai/v1`, model: `gpt-5.4`
   (any GPT-5-family model reproduces it, OpenAI directly or via Azure Foundry)
3. Record any meeting, or batch-import an audio file with ⌘P
4. Transcription succeeds; protocol generation fails with the 400 above and no `.md`
   is written

### Diagnostic log

Log lines quoted above, from
`log show --predicate 'subsystem BEGINSWITH "com.meetingtranscriber"'`.

alexpfau:
Fixes #582.

## Problem

`OpenAIProtocolGenerator` sends the output cap as `max_tokens`. The OpenAI Chat
Completions API renamed that to `max_completion_tokens`, and GPT-5-family models now
reject the old name with HTTP 400 instead of accepting it as an alias. Every protocol
generation fails, so users get `.txt` transcripts but never an `.md` protocol.

It cannot simply be renamed: LM Studio and Ollama accept only `max_tokens`.

## Approach

Send `max_tokens` as today. Only when the endpoint answers **400 with a body naming
both parameters** does it retry once with `max_completion_tokens`. Requiring both names
to appear keeps unrelated 400s (bad key, unknown model, malformed request) from
triggering a pointless second request.

No new setting, no model-name sniffing, and existing local-server users are unaffected —
they never see the retry path.

The retry costs one extra round-trip on the first generation against a strict endpoint;
the 400 comes back immediately. Caching the negotiated parameter per endpoint would
avoid even that, but it needs somewhere to persist the preference, so I left it out —
happy to add it if you'd prefer.

## Testing

- 141 tests green in `OpenAIProtocolGeneratorTests` + `ProtocolGeneratorTests`
- Four new tests: the real Azure 400 body matches; unrelated 400s don't; a rejection
  produces exactly `["max_tokens", "max_completion_tokens"]` in order; an unrelated 400
  produces exactly one attempt
- **Verified end-to-end** against Azure AI Foundry with a `gpt-5.4` deployment, on both
  a live auto-detected meeting and a batch import (⌘P). Protocols generate correctly
  where every attempt previously failed with the 400.
- Also re-checked against `grok-4.3`, which accepts `max_tokens`, to confirm the
  non-retry path still works unchanged.

Note: `swift test` for the full suite hangs locally on
`AppSettingsTests.testOpenAIAPIKeyViaKeychainHelper` (blocks on a macOS Keychain
prompt), so I ran the affected suites rather than the whole thing. Unrelated to this
change, but flagging it.

### DIFF
--- a/app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
+++ b/app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
@@ -13,16 +13,25 @@ struct OpenAIProtocolGenerator: ProtocolGenerating {
     /// bytes. Catches a fully-stalled connection, but does NOT bound total time —
     /// a slow-but-trickling stream resets it on every byte. See `maxTotalSeconds`.
     let timeoutSeconds: TimeInterval
-    /// Hard wall-clock cap for the whole generation. `timeoutSeconds` alone lets
+    /// Wall-clock cap for a single request attempt. `timeoutSeconds` alone lets
     /// a struggling LLM that trickles tokens run for hours (each byte resets the
     /// idle timer); this deadline ends it regardless. Generous by default so a
-    /// legitimately long protocol isn't cut off.
+    /// legitimately long protocol isn't cut off. Note the deadline is applied per
+    /// attempt, so the `max_completion_tokens` fallback in `generate()` can span
+    /// up to twice it — only reachable via an endpoint that trickles a 400 body.
     let maxTotalSeconds: TimeInterval
     /// Output-token cap sent as `max_tokens`, or `max_completion_tokens` when the
     /// endpoint rejects the former. Bounds a runaway/verbose
     /// generation (the 131k-context case that ground for ~2h) by length rather
     /// than time. Generous — a meeting protocol is well under this — and a hit
     /// is surfaced as `protocolTruncated`, never silently presented as complete.
+    ///
+    /// The two names are not interchangeable: `max_completion_tokens` bounds
+    /// visible output tokens *plus* reasoning tokens, and reasoning tokens never
+    /// arrive as content deltas. A long transcript can therefore end the stream on
+    /// `finish_reason: length` with little or no content, surfaced here as
+    /// `protocolTruncated`. The old name never had to account for that, since the
+    /// models that accept it emit no reasoning tokens.
     let maxOutputTokens: Int
     let session: URLSession
 
--- a/app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
+++ b/app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
@@ -18,7 +18,8 @@ struct OpenAIProtocolGenerator: ProtocolGenerating {
     /// idle timer); this deadline ends it regardless. Generous by default so a
     /// legitimately long protocol isn't cut off.
     let maxTotalSeconds: TimeInterval
-    /// Output-token cap sent as `max_tokens`. Bounds a runaway/verbose
+    /// Output-token cap sent as `max_tokens`, or `max_completion_tokens` when the
+    /// endpoint rejects the former. Bounds a runaway/verbose
     /// generation (the 131k-context case that ground for ~2h) by length rather
     /// than time. Generous — a meeting protocol is well under this — and a hit
     /// is surfaced as `protocolTruncated`, never silently presented as complete.
@@ -53,11 +54,42 @@ struct OpenAIProtocolGenerator: ProtocolGenerating {
             ["role": "user", "content": transcript],
         ]
 
+        do {
+            return try await send(messages: messages, tokenLimit: .maxTokens)
+        } catch let ProtocolError.httpError(status, body)
+            where status == 400 && Self.rejectsMaxTokens(body) {
+            logger.info("Endpoint rejected 'max_tokens'; retrying with 'max_completion_tokens'")
+            return try await send(messages: messages, tokenLimit: .maxCompletionTokens)
+        }
+    }
+
+    /// Name of the output-token cap parameter in the request body.
+    ///
+    /// OpenAI renamed `max_tokens` to `max_completion_tokens`; current models
+    /// (GPT-5 family and newer) reject the old name with HTTP 400, while local
+    /// servers such as LM Studio and Ollama only understand the old one. So the
+    /// old name is sent first and the new one used only on an explicit refusal.
+    enum TokenLimitParameter: String {
+        case maxTokens = "max_tokens"
+        case maxCompletionTokens = "max_completion_tokens"
+    }
+
+    /// True when a 400 body is the API specifically refusing `max_tokens` and
+    /// naming `max_completion_tokens` as the replacement. Both names must appear
+    /// so unrelated 400s (bad model, bad key, malformed request) don't retry.
+    static func rejectsMaxTokens(_ errorBody: String) -> Bool {
+        errorBody.contains(TokenLimitParameter.maxCompletionTokens.rawValue)
+            && errorBody.contains(TokenLimitParameter.maxTokens.rawValue)
+    }
+
+    private func send(
+        messages: [[String: Any]], tokenLimit: TokenLimitParameter,
+    ) async throws -> String {
         let body: [String: Any] = [
             "model": model,
             "messages": messages,
             "stream": true,
-            "max_tokens": maxOutputTokens,
+            tokenLimit.rawValue: maxOutputTokens,
         ]
 
         let bodyData = try JSONSerialization.data(withJSONObject: body)
--- a/app/MeetingTranscriber/Tests/OpenAIProtocolGeneratorTests.swift
+++ b/app/MeetingTranscriber/Tests/OpenAIProtocolGeneratorTests.swift
@@ -336,6 +336,79 @@ final class OpenAIProtocolGeneratorTests: XCTestCase { // swiftlint:disable:this
         _ = try await gen.generate(transcript: "Test", title: "Test", diarized: false)
     }
 
+    // MARK: - max_tokens / max_completion_tokens negotiation
+
+    /// Verbatim 400 body returned by the OpenAI Chat Completions API (and Azure
+    /// Foundry) for GPT-5-family models when `max_tokens` is sent.
+    private static let unsupportedMaxTokensBody = """
+    {"error":{"message":"Unsupported parameter: 'max_tokens' is not supported \
+    with this model. Use 'max_completion_tokens' instead.",\
+    "type":"invalid_request_error","param":"max_tokens",\
+    "code":"unsupported_parameter"}}
+    """
+
+    func testRejectsMaxTokensMatchesUnsupportedParameterBody() {
+        XCTAssertTrue(OpenAIProtocolGenerator.rejectsMaxTokens(Self.unsupportedMaxTokensBody))
+    }
+
+    func testRejectsMaxTokensIgnoresUnrelatedErrors() {
+        XCTAssertFalse(
+            OpenAIProtocolGenerator.rejectsMaxTokens(
+                #"{"error":{"message":"Incorrect API key provided","code":"invalid_api_key"}}"#,
+            ),
+        )
+        XCTAssertFalse(
+            OpenAIProtocolGenerator.rejectsMaxTokens(
+                #"{"error":{"message":"The model does not exist","param":"model"}}"#,
+            ),
+        )
+        XCTAssertFalse(OpenAIProtocolGenerator.rejectsMaxTokens(""))
+    }
+
+    func testGenerateRetriesWithMaxCompletionTokensWhenRejected() async throws {
+        final class Recorder: @unchecked Sendable { var keys: [String] = [] }
+        let recorder = Recorder()
+
+        MockURLProtocol.handler = { request in
+            if self.requestBody(request)["max_tokens"] != nil {
+                recorder.keys.append("max_tokens")
+                return self.mockResponse(
+                    request, status: 400, body: Data(Self.unsupportedMaxTokensBody.utf8),
+                )
+            }
+            recorder.keys.append("max_completion_tokens")
+            return self.mockResponse(request, body: Data(Self.okSSE.utf8))
+        }
+
+        let gen = makeGenerator(session: makeMockSession())
+        let result = try await gen.generate(transcript: "Test", title: "Test", diarized: false)
+
+        XCTAssertEqual(result, "OK")
+        XCTAssertEqual(recorder.keys, ["max_tokens", "max_completion_tokens"])
+    }
+
+    func testGenerateDoesNotRetryOnUnrelated400() async throws {
+        final class Recorder: @unchecked Sendable { var count = 0 }
+        let recorder = Recorder()
+
+        MockURLProtocol.handler = { request in
+            recorder.count += 1
+            return self.mockResponse(
+                request, status: 400,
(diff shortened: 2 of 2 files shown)


app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
```diff
@@ -53,11 +54,42 @@ struct OpenAIProtocolGenerator: ProtocolGenerating {
...
+        case maxTokens = "max_tokens"
+        case maxCompletionTokens = "max_completion_tokens"
+    }
+
+    /// True when a 400 body is the API specifically refusing `max_tokens` and
+    /// naming `max_completion_tokens` as the replacement. Both names must appear
+    /// so unrelated 400s (bad model, bad key, malformed request) don't retry.
+    static func rejectsMaxTokens(_ errorBody: String) -> Bool {
+        errorBody.contains(TokenLimitParameter.maxCompletionTokens.rawValue)
+            && errorBody.contains(TokenLimitParameter.maxTokens.rawValue)
```
pasrom, line 91: The "both names must appear" rule that the doc comment states is not pinned by any test. I removed this second `contains` and the suite stayed at 48 of 48 green, because both negative fixtures in `testRejectsMaxTokensIgnoresUnrelatedErrors` omit both names and therefore pass whichever half survives. Suggestion over there.

Smaller point, worth at most a comment: this reads an error body that `streamProtocol` assembles line by line under a size guard, so the detector quietly assumes both names arrive on one of the first lines. They do for OpenAI and Azure, whether minified or pretty-printed, so nothing to change today.

app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
```diff
@@ -18,7 +18,8 @@ struct OpenAIProtocolGenerator: ProtocolGenerating {
     /// idle timer); this deadline ends it regardless. Generous by default so a
     /// legitimately long protocol isn't cut off.
     let maxTotalSeconds: TimeInterval
-    /// Output-token cap sent as `max_tokens`. Bounds a runaway/verbose
+    /// Output-token cap sent as `max_tokens`, or `max_completion_tokens` when the
+    /// endpoint rejects the former. Bounds a runaway/verbose
```
pasrom, line 24: Since the cap now also goes out as `max_completion_tokens`, it is worth a line here that the two are not interchangeable: the new parameter is defined as an upper bound on visible output tokens plus reasoning tokens, and reasoning tokens never arrive as content deltas. A long transcript can therefore end the stream on `finish_reason: length` with little or no content, which this class turns into `protocolTruncated`. The old name never had to account for that, since the models that accept it emit no reasoning tokens.

Your end-to-end run shows the common case is fine, so nothing to change beyond this comment if you agree with the wording. Whether the default cap should be higher on the fallback path is my problem rather than yours, and I will look at it separately.

app/MeetingTranscriber/Sources/OpenAIProtocolGenerator.swift
```diff
@@ -18,7 +18,8 @@ struct OpenAIProtocolGenerator: ProtocolGenerating {
     /// idle timer); this deadline ends it regardless. Generous by default so a
     /// legitimately long protocol isn't cut off.
     let maxTotalSeconds: TimeInterval
```
pasrom, line 22: Minor, and it is this comment rather than the code that is now inaccurate: the cap is documented as covering the whole generation, but the deadline race lives inside `send()`, so the retry path allows up to twice that. It takes a broken proxy that trickles the 400 body to actually get there, since every byte resets the idle timer, so the realistic cost is nil. Either compute the deadline once in `generate()` and pass the remaining budget down, or soften the wording.

app/MeetingTranscriber/Tests/OpenAIProtocolGeneratorTests.swift
```diff
@@ -336,6 +336,79 @@ final class OpenAIProtocolGeneratorTests: XCTestCase { // swiftlint:disable:this
...
+            OpenAIProtocolGenerator.rejectsMaxTokens(
+                #"{"error":{"message":"Incorrect API key provided","code":"invalid_api_key"}}"#,
+            ),
+        )
+        XCTAssertFalse(
+            OpenAIProtocolGenerator.rejectsMaxTokens(
+                #"{"error":{"message":"The model does not exist","param":"model"}}"#,
+            ),
+        )
+        XCTAssertFalse(OpenAIProtocolGenerator.rejectsMaxTokens(""))
```
pasrom, line 365: Both fixtures above omit both parameter names, so they hold no matter which half of the conjunction in `rejectsMaxTokens` survives. One body per name pins the rule the production doc comment states. Verified: green as your code stands, and each fixture goes red under one of the two halves.

```suggestion
        XCTAssertFalse(OpenAIProtocolGenerator.rejectsMaxTokens(""))
        // Only one of the two names present: a value error, not the rename.
        XCTAssertFalse(OpenAIProtocolGenerator.rejectsMaxTokens(
            #"{"error":{"message":"max_tokens is too large: 16000","param":"max_tokens"}}"#,
        ))
        XCTAssertFalse(OpenAIProtocolGenerator.rejectsMaxTokens(
            #"{"error":{"message":"Use max_completion_tokens for this model"}}"#,
        ))
```

app/MeetingTranscriber/Tests/OpenAIProtocolGeneratorTests.swift
```diff
@@ -336,6 +336,79 @@ final class OpenAIProtocolGeneratorTests: XCTestCase { // swiftlint:disable:this
...
+        let recorder = Recorder()
+
+        MockURLProtocol.handler = { request in
+            if self.requestBody(request)["max_tokens"] != nil {
+                recorder.keys.append("max_tokens")
+                return self.mockResponse(
+                    request, status: 400, body: Data(Self.unsupportedMaxTokensBody.utf8),
+                )
+            }
+            recorder.keys.append("max_completion_tokens")
```
pasrom, line 389: This branch records the parameter name without checking that the retry actually carries it. I mutated `send` to drop the cap entirely on the fallback path and the suite stayed at 48 of 48 green, so the request can silently lose the runaway bound on exactly the path this PR adds. Asserting the body here also pins the value, which nothing currently does, since the existing happy-path test only checks that the key is non-nil.

Verified: green as your code stands, red once the retry stops sending the cap.

```suggestion
            let retryBody = self.requestBody(request)
            XCTAssertEqual(retryBody["max_completion_tokens"] as? Int, 16000)
            XCTAssertNil(retryBody["max_tokens"], "the retry must replace the cap, not add to it")
            recorder.keys.append("max_completion_tokens")
```

app/MeetingTranscriber/Tests/OpenAIProtocolGeneratorTests.swift
```diff
@@ -336,6 +336,79 @@ final class OpenAIProtocolGeneratorTests: XCTestCase { // swiftlint:disable:this
...
+
+        let gen = makeGenerator(session: makeMockSession())
+        do {
+            _ = try await gen.generate(transcript: "Test", title: "Test", diarized: false)
+            XCTFail("expected the 400 to propagate")
+        } catch let ProtocolError.httpError(status, _) {
+            XCTAssertEqual(status, 400)
+        }
+        XCTAssertEqual(recorder.count, 1, "an unrelated 400 must not trigger a retry")
+    }
```
pasrom, line 420: One more mutant survives: removing `status == 400` from the `where` clause keeps all 48 tests green, because nothing sends a non-400 whose body names both parameters. A gateway that wraps the upstream message into a 502 or a 422 would then trigger a retry against a server that is failing for an unrelated reason. A near-copy of this test closes it, verified green as your code stands and red without the status guard. Note that the load-bearing assertion is the request count, not the status: without the guard the retried 502 still propagates as a 502.

```swift
    func testGenerateDoesNotRetryOnNon400NamingBothParameters() async throws {
        final class Recorder: @unchecked Sendable { var count = 0 }
        let recorder = Recorder()

        MockURLProtocol.handler = { request in
            recorder.count += 1
            return self.mockResponse(
                request, status: 502, body: Data(Self.unsupportedMaxTokensBody.utf8),
            )
        }

        let gen = makeGenerator(session: makeMockSession())
        do {
            _ = try await gen.generate(transcript: "Test", title: "Test", diarized: false)
            XCTFail("expected the 502 to propagate")
        } catch let ProtocolError.httpError(status, _) {
            XCTAssertEqual(status, 502)
        }
        XCTAssertEqual(recorder.count, 1, "only an explicit 400 refusal may retry")
    }
```

pasrom (commented):
Thanks for the report and for the fix, and especially for the verbatim error body plus the end-to-end verification against a real deployment. That made this reviewable in one pass.

The approach is right, and it matches what this codebase already does elsewhere: `FluidDiarizer` retries once with the alternative parameter after a failed run. Yours is the stricter version, since it gates on the specific 400 rather than on any error.

I also agree with leaving the per-endpoint cache out. One more reason to like that call, on top of the ones you give: the negotiated answer is a property of the model behind the endpoint rather than of the endpoint itself, and a wrong cache entry would fail silently. A local server that ignores an unknown parameter would simply drop the output cap, with no 400 to recover from.

On formatting, since your amended push crossed this review: I had written it against the previous commit, where the pinned SwiftFormat rejected the brace in `generate()` plus four multi-line calls in the new tests, which is what turned the lint check red. Your push fixes exactly those five spots, and you found them from the red check alone, without seeing any of this. I checked the new commit locally against SwiftFormat 0.61.1 and `swiftlint --strict` and both are clean. The workflows on the new commit are queued behind a maintainer approval, so CI has not re-checked it yet.

That leaves no blocker: three test gaps and two optional notes.

The gaps are mutants I ran against the suite on this commit. Each of these leaves all 48 tests in the file green:

- dropping `status == 400` from the `where` clause
- dropping either half of the conjunction in `rejectsMaxTokens`
- dropping the token cap from the retry request entirely

So the suite pins that a retry happens, but neither the conditions that decide when nor what the retry actually sends. The three suggestions below close that in about 15 lines, and each is verified rather than sketched: they are green against your code as it stands and each goes red under exactly the mutation it is meant to catch. They are also pre-formatted for the pinned SwiftFormat, so applying them adds no new lint work: with all three applied on top of your commit, SwiftFormat is clean, `swiftlint --strict` is clean, and the file runs 49 of 49 green, the extra one being the test the third suggestion adds.

I would like those three in before merge, since the bar in this repo is that a fix pins its own effect, but they are ready-made suggestions rather than a change request. The two inline notes are optional.

What I went looking for and did not find: no retry loop is structurally possible, cancellation and timeout errors cannot match the catch pattern, the double send has no billing or state side effects, and the new tests are safe under the parallel runner, which forks per test method and resets the handlers in `tearDown`.

On the local suite hang you flagged: your reading looks right to me. The keychain item that test reads is per-user rather than per-process, so a leftover from a differently signed test binary prompts. Unrelated to this change, and both test jobs were green on the previous run.

alexpfau:
Thanks — the mutation results were the useful part of that review. Each of the three was a real gap, and I would not have found them from the suite being green.

Pushed 3895f6f with all three suggestions applied as written, plus both optional notes:

- Two fixtures naming only one of the parameters, so each half of the conjunction is now pinned separately.
- The retry body is asserted to carry `max_completion_tokens` at the configured value and no longer carry `max_tokens`, so the cap is replaced rather than dropped or duplicated.
- `testGenerateDoesNotRetryOnNon400NamingBothParameters` for the wrapped-upstream-message case, asserting the request count rather than the status.
- The `maxOutputTokens` doc now records that the two names are not interchangeable, in your wording — reasoning tokens count against the new cap and never arrive as content deltas, so a long transcript can end on `finish_reason: length` with little content.
- For `maxTotalSeconds` I took the cheap option and softened the wording to say the deadline is per attempt, rather than threading a remaining budget through `send()`. Restructuring it for a path that needs a proxy to trickle a 400 body seemed like the wrong trade, but say the word if you would rather have the real fix.

I re-ran your four mutants rather than take the claim on trust, and they behave as you describe: dropping the status guard fails only the new 502 test (2 requests, not 1), dropping either half of the conjunction fails one of the two new fixtures, and dropping the cap from the retry fails the new body assertion with `nil` against `Optional(16000)`. Unmutated, the file is 49 of 49 green. SwiftFormat 0.61.1 pinned from the workflow reports 0 of 407 files needing changes, and `swiftlint --strict` is clean.

One unrelated thing I ran into while verifying, since it cost me a confusing red run. `ProtocolGeneratorTests.testBuildSystemPromptSubstitutesLanguage` also fails on my machine, and for the same class of reason as the keychain hang: it reads real per-user state. `buildSystemPrompt` calls `loadPrompt()` with the default argument, so it picks up `~/Library/Application Support/MeetingTranscriber/protocol_prompt.md`, and mine is customised without the `{LANGUAGE}` placeholder. Moving the file aside takes that suite to 142 of 142. `loadPrompt` already takes an injectable `url` for exactly this reason, so the gap is only that `buildSystemPrompt` has no matching parameter to forward.

Nothing to do with this PR, and the same two tests pass in CI because the runner has neither the file nor the keychain item. But it does mean the suite cannot pass on a machine that actually uses the app, which seemed worth reporting rather than working around locally. Filed separately as #584 — happy to send the `buildSystemPrompt` half as a small PR if you want it, since it is the same pattern you already have on `loadPrompt`.

pasrom (approved):
All three suggestions applied verbatim and both optional notes taken, in a separate commit rather than amended into the fix. Appreciated.

Verified on 3895f6fd rather than taken on trust: pinned SwiftFormat 0.61.1 and `swiftlint --strict` clean, 142 tests green across the two affected suites, and removing the status guard now turns the new 502 test red at the request count. The gap is genuinely closed.

On #584, that is a real find and it lands closer to home than you could have known: the prompt file exists on my machine too, and the test passes here only because my customised prompt still happens to contain the placeholder. One edit away from the same red run. Please do send the `buildSystemPrompt` half as a PR, I would take it.

Issue #584, opened 2026-08-07
[Bug] Two tests fail or hang on a machine that actually runs the app (real per-user state)


alexpfau:
### App version

0.6.0, built from source at `f9c8b04` (`main`)

### Build variant

Built from source

### macOS version

26.5.2 (Swift 6.3.3)

### What happened?

`swift test` cannot pass on a developer machine that also uses Meeting Transcriber as an app. Two tests read real per-user state instead of test-owned state, so their result depends on how the developer has configured their own installation.

**1. `ProtocolGeneratorTests.testBuildSystemPromptSubstitutesLanguage` — fails**

```
ProtocolGeneratorTests.swift:216: error: XCTAssertTrue failed
```

`buildSystemPrompt` calls `loadPrompt()` with its default argument, which resolves to `AppPaths.customPromptFile` (`~/Library/Application Support/MeetingTranscriber/protocol_prompt.md`). I have customised that file and my version does not contain the `{LANGUAGE}` placeholder, so the substitution assertion fails. Anyone who has edited their protocol prompt — a supported, documented feature — gets this failure.

The interesting part is that the fix already exists one level down. `loadPrompt` takes an injectable `url`, and its doc comment says so explicitly:

> The `url` parameter exists so tests can use unique per-test paths instead of racing on the shared one.

The four `loadPrompt` tests use it correctly and are unaffected. Only `buildSystemPrompt` has no matching parameter to forward, so the three tests that call it fall through to the real user file. Two of them happen to survive because they compare a diarized against a non-diarized prompt and so hold whatever the prompt text is; only the language one asserts on content, so only it fails.

**2. `AppSettingsTests.testOpenAIAPIKeyViaKeychainHelper` — hangs**

Already noted on #583 and confirmed there. It reads a per-user rather than per-process keychain item, so a leftover from a differently signed test binary blocks on a Keychain prompt and the run never finishes. The test does manage its own slot, so the problem is the shared scope rather than the cleanup.

Both are the same class of defect: a test reaching outside its own fixtures into real user state. Neither shows up in CI, because the runner has neither the file nor the keychain item — which is exactly why it went unnoticed.

Practical effect is worse than two red tests. The keychain one hangs rather than fails, so the common workaround is to stop running the full suite locally and run only the affected suites, which is what I ended up doing on #583.

### Steps to reproduce

For the prompt one:

[...]

PR #585, opened 2026-08-07
test(app): stop buildSystemPrompt tests reading the real user prompt

alexpfau:
Closes the `buildSystemPrompt` half of #584, as requested on #583.

## The gap

`buildSystemPrompt` called `loadPrompt()` with its default argument, so it resolved to `AppPaths.customPromptFile` even under test. The three tests that call it read `~/Library/Application Support/MeetingTranscriber/protocol_prompt.md`, so their result depends on whether the developer has customised their own prompt through the app.

`loadPrompt` already took an injectable `url` for exactly this reason, and its doc comment says so. The gap was only that `buildSystemPrompt` had no matching parameter to forward. This adds one with the same default, so both production callers (`OpenAIProtocolGenerator`, `ClaudeCLIProtocolGenerator`) are unchanged.

## Both tests, not just the failing one

Only `testBuildSystemPromptSubstitutesLanguage` was red, because only it asserts on content. `testBuildSystemPromptAppendsDiarizationNoteWhenDiarized` compares a diarized against a non-diarized prompt, so it holds whatever the text happens to be — it was reading your file too, and passing by construction rather than by isolation. Since your note was that the test passes on your machine only because your customised prompt still happens to contain the placeholder, leaving the second one reading real state seemed like the same bug waiting for a different edit. Both now point at test-owned fixtures.

The two use different fixtures on purpose: the language test writes a known prompt, the diarization test names a path it never creates so both calls fall back to the built-in default.

## Why equality rather than containment

The language test now writes `Protocol language: {LANGUAGE}.` and asserts the result equals `Protocol language: Polish.`. Equality is what pins the fix. `contains("Polish")` would still pass against any prompt that happened to mention the language, so it would not notice the forwarding being dropped again — which makes it the assertion that turns this from a passing test into a regression test.

## Verification

Run with the real customised prompt file in place, since that is the condition that reproduces the failure — my copy is 4537 bytes and contains no `{LANGUAGE}`.

- Reverting the `from: promptURL` forwarding turns `testBuildSystemPromptSubstitutesLanguage` red, and the diff in the failure output is your prompt file leaking in, which is the defect itself.
- With the fix: both tests green, and the `ProtocolGenerator`, `OpenAIProtocolGenerator` and `ClaudeCLIProtocolGenerator` suites green.
- SwiftFormat 0.61.1 pinned from the workflow: 0 of 409 files need formatting. `swiftlint --strict`: 0 violations.

## Scope

The keychain hang in `AppSettingsTests.testOpenAIAPIKeyViaKeychainHelper` is untouched, so #584 stays open for it. That one is a shared-scope problem rather than a missing parameter, so it wants its own change — happy to take it if you want it, though the fix there is less obvious than this one.

### DIFF
--- a/app/MeetingTranscriber/Sources/ProtocolGenerator.swift
+++ b/app/MeetingTranscriber/Sources/ProtocolGenerator.swift
@@ -94,8 +94,13 @@ enum ProtocolGenerator {
     /// Build the localized system prompt: `loadPrompt` + `applyLanguage`
     /// + optional `diarizationNote`. Excludes the transcript itself —
     /// callers append or attach it as they see fit.
-    static func buildSystemPrompt(diarized: Bool, language: String) -> String {
-        var prompt = applyLanguage(loadPrompt(), language: language)
+    ///
+    /// `promptURL` is forwarded to `loadPrompt` and exists for the same reason:
+    /// without it tests fall through to the shared `AppPaths.customPromptFile`,
+    /// so their result depends on whether the developer has customised their own
+    /// prompt through the app.
+    static func buildSystemPrompt(diarized: Bool, language: String, promptURL: URL = AppPaths.customPromptFile) -> String {
+        var prompt = applyLanguage(loadPrompt(from: promptURL), language: language)
         if diarized { prompt += diarizationNote }
         return prompt
     }
--- a/app/MeetingTranscriber/Tests/ProtocolGeneratorTests.swift
+++ b/app/MeetingTranscriber/Tests/ProtocolGeneratorTests.swift
@@ -211,15 +211,23 @@ final class ProtocolGeneratorTests: XCTestCase {
 
     // MARK: - System Prompt Construction
 
-    func testBuildSystemPromptSubstitutesLanguage() {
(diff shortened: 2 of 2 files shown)


pasrom (approved):
Verified on 00b2a5df, and the acceptance check is stronger than the one in your description.

I reverted the `from: promptURL` forwarding and ran it on a machine where the prompt file exists and still contains the placeholder, which is the condition under which the old test passed. `testBuildSystemPromptSubstitutesLanguage` goes red there too, with my prompt file visible in the failure diff. So the regression test bites in both variants of the environment, not only in yours. That is the equality assertion earning its keep: `contains("Polish")` would have stayed green on my side, since my prompt also says Polish after substitution.

Taking the second test along was the right call for the same reason. It was green by construction rather than by isolation.

Also checked: pinned SwiftFormat 0.61.1 and `swiftlint --strict` clean, 142 tests green across the three affected suites, and no test now reaches `buildSystemPrompt` or `loadPrompt` without injection, so this class is closed here.

Leaving #584 open for the keychain half. Agreed that it wants its own change, and yes please, if you want to take it.

PR #587, opened 2026-08-07
fix(app): isolate AppSettings tests from the production Keychain account

### LINKED ISSUE #584: [Bug] Two tests fail or hang on a machine that actually runs the app (real per-user state)

alexpfau:
Closes #584 — the Keychain half you left open after #585.

## The defect

`AppSettingsTests.testOpenAIAPIKeyViaKeychainHelper` drove the **production** Keychain slot directly: service `com.meetingtranscriber`, account `openAIAPIKey` — precisely what `AppSettings.openAIAPIKey` reads and writes.

Two consequences on any machine that also runs the app:

1. **The reported hang.** The item is owned by the app's code signature, so the test binary touching it raises an authorization prompt. `swift test` blocks on it indefinitely and the suite never finishes.
2. **Credential loss, if that prompt is granted.** The test opens with an unconditional `KeychainHelper.delete(key: "openAIAPIKey")` and defers another. That is the API key the user configured in Settings — running the suite deletes it. I had not expected this second one going in; it is the same root cause, but it makes the fix worth more than a hang alone would.

Grepping the rest of the suite: `HF_TOKEN_TEST` and the `KEYCHAIN_TEST_<UUID>` in `KeychainHelperTests` are already test-owned accounts, and `AppSettings+RPC.swift` excludes the key by construction, so `RPCSettingsStateTests` never reaches the Keychain. This was the only offender.

## The fix

`AppSettings` gains an injectable Keychain account defaulting to the production one — the same shape as the `defaults:` injection already on the initializer, and as the `promptURL:` seam from #585. No production call site changes. The fixture mints a unique account per test and removes it in `tearDown`.

Asserting through the injected account is what keeps the test honest rather than merely quiet: if `AppSettings` ignored the parameter, every read would come back nil.

I also refreshed the `RPCSettingsStateTests` comment that justified not writing the Keychain by citing the now-removed `--parallel` race against this test. The stronger reason — it is a real user credential behind a signature-gated prompt — still stands, so the comment says that instead of going stale.

## file_length

`AppSettings.swift` sat at **599** of its 600-line budget, so the two added lines failed `swiftlint --strict`. Rather than suppress the rule, the existing `Computed` section (`watchApps`) moves verbatim into `AppSettings+Computed.swift`, following the `AppSettings+RPC.swift` precedent. Pure move, no behaviour change; the main file is now 589 with headroom. Happy to reshape this if you would have made room elsewhere.

## Acceptance check

Rather than reverting the fix — which would have pointed the test back at my own real API key — I mutated the initializer to use a *different* unique account, i.e. "AppSettings does not honour the injection". The test fails on exactly the intended assertion:

```
AppSettingsTests.swift:269: error: XCTAssertEqual failed:
  ("nil") is not equal to ("Optional("sk-test-key")")
```

Restored afterwards, and confirmed no mutant residue in the diff.

## Verified locally

- `swiftlint lint --strict` → 0 violations
- SwiftFormat **0.61.1** (the CI pin, via `scripts/lint.sh --format-only`) → 0/410 files
- `AppSettingsTests`, `KeychainHelperTests`, `RPCSettingsStateTests`, `PowerAssertionDetectorPatternsTests` → 86 tests, 0 failures. No Keychain prompt appeared.

**Left to CI:** `analyze`, the full test matrix, the App Store variant and `audiotap-coverage`. I skipped the `xcodebuild` and full-suite runs deliberately this time — the full suite stalls on model downloads on my machine, and I did not want to claim a green I had not actually seen. The change has no `#if APPSTORE` interaction.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppSettings+Computed.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings+Computed.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+/// Values derived from the stored `AppSettings` toggles. Split out of
+/// `AppSettings.swift` purely to keep that file under the `file_length` limit;
+/// there is no behavioural difference from declaring these inline.
+extension AppSettings {
+    /// The meeting apps the user opted to watch. Drives auto-detection: the
+    /// `WatchingController` default detector keeps only the assertion patterns
+    /// whose app is listed here (`PowerAssertionDetector.patterns(watching:)`),
+    /// read at each watch start. Freshness: FROZEN per watch session.
+    var watchApps: [String] {
+        var apps: [String] = []
+        if watchTeams { apps.append("Microsoft Teams") }
+        if watchZoom { apps.append("Zoom") }
+        if watchWebex { apps.append("Webex") }
+        if watchBrowserMeetings { apps.append(AppMeetingPattern.chromeBrowser.appName) }
+        if watchWeChat { apps.append(AppMeetingPattern.wechat.appName) }
+        if watchTencentMeeting { apps.append(AppMeetingPattern.tencentMeeting.appName) }
+        if watchFaceTime { apps.append(AppMeetingPattern.faceTime.appName) }
+        if watchWhatsApp { apps.append(AppMeetingPattern.whatsApp.appName) }
+        return apps
+    }
+}
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -98,6 +98,14 @@ final class AppSettings {
     /// processes don't race on the shared on-disk plist.
     @ObservationIgnored private let defaults: UserDefaults
 
+    /// Keychain account backing `openAIAPIKey`. Production callers pass
+    /// nothing → the real account. Tests inject a unique account so they never
+    /// touch the key the user configured in the app: the Keychain is scoped
+    /// per user, not per process, and the production item belongs to the app's
+    /// code signature, so a test binary reading it raises an authorization
+    /// prompt that blocks the run.
+    @ObservationIgnored private let apiKeyAccount: String
+
     // MARK: - Apps to Watch
 
     var watchTeams: Bool {
@@ -380,12 +388,12 @@ final class AppSettings {
     }
 
     var openAIAPIKey: String {
-        get { KeychainHelper.read(key: "openAIAPIKey") ?? "" }
+        get { KeychainHelper.read(key: apiKeyAccount) ?? "" }
         set {
             if newValue.isEmpty {
-                KeychainHelper.delete(key: "openAIAPIKey")
+                KeychainHelper.delete(key: apiKeyAccount)
             } else {
-                KeychainHelper.save(key: "openAIAPIKey", value: newValue)
+                KeychainHelper.save(key: apiKeyAccount, value: newValue)
             }
         }
     }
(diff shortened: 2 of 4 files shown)


pasrom (approved):
Verified on b001d5cc.

I confirmed the credential-loss half at the code level rather than by running it, for obvious reasons: `KeychainHelper` uses `AppPaths.logSubsystem` as the service and the old test drove the `openAIAPIKey` account, which is exactly what `AppSettings.openAIAPIKey` reads and writes. The unconditional delete on entry plus the deferred one meant the suite really did remove a configured key. You were right to weigh that above the hang.

Acceptance check reproduced independently: mutating the initializer to ignore the injected account, pointed at a throwaway slot rather than the production one, fails the test on the intended assertion, `nil` against `Optional("sk-test-key")`. So the seam cannot be dropped again without the test noticing.

Also checked: no other test touches the production slot, the only remaining literals are the test-owned `HF_TOKEN_TEST` and `KEYCHAIN_TEST_<UUID>`; the `AppSettings+Computed.swift` split is a pure move plus a file header, with the main file back to 589 lines; pinned SwiftFormat 0.61.1 and repo-wide `swiftlint --strict` clean; 72 tests green across the three affected suites with no Keychain prompt.

Splitting the file rather than suppressing the rule was the right call, and following the existing extension precedent means there is nothing to reshape.

Thanks also for saying plainly which runs you had not done instead of claiming a green you had not seen. CI has since covered exactly those: analyze, both test variants, both builds and coverage are green.

PR #600, opened 2026-08-10
feat(ui): blink the menu-bar badge when the app cannot do its job

duncanwood:
The menu bar shows a calm grey "Idle" for two different things: "nothing to do" and "I am broken". Every failure mode this app has announces itself only by notification, and Do Not Disturb suppresses those.

## What it costs

On 2026-08-04 a 62-minute interview recorded the mic perfectly and captured the other participant for 40 seconds before the tap went to digital silence for the remaining 59 minutes. `ChannelHealthController` had already detected it. The notification was suppressed, the menu bar showed the ordinary recording badge throughout, and the loss was discovered only when the transcript turned out to be one-sided — too late to redo the call.

It happened again on 2026-08-09: 19 minutes, app track at the silence floor from the first second, menu bar calm the whole time.

## The fix

Two changes.

1. `.error` and `.userAction` now animate, blinking their exclamation overlay (~1.2 s on, ~1.2 s off at the existing 0.4 s frame timer). The badge body stays still; only the overlay blinks. A static failure badge is one you scroll past.

2. A capture problem now outranks `.recording`. `ChannelHealthController` already tracked `appSilentActive` / `micSilentActive` / `recordingSilentActive`, but `BadgeKind.compute` never received them, so a recording capturing nothing showed the ordinary recording badge. That is exactly how an hour of one-sided audio reached the archive unremarked.

## Evidence

Three existing tests encoded the old "these badges are static" intent and are updated rather than deleted — they now derive the static set from `isAnimated`, so they cannot drift out of sync again. `testLargeAnimationFrameDoesNotCrash` was asserting `isTemplate` as a proxy for safe wrapping; `.error` renders a non-template red exclamation, so it now asserts the actual guarantee (frame 999 renders identically to frame 999 % frameCount).

`MenuBarBlinkTests` is new: 92 lines pinning the blink duty cycle, that the overlay alone animates, and the new precedence over `.recording`.

## Provenance

Written 2026-08-04 against the then-current main, where it passed the menu-bar suite (128 tests) and the audiotap suite (194 tests) locally. Rebased onto v0.7.0 today; it applied without conflict. I could not re-run the app package suite on this machine tonight — the test host wedges in an uninterruptible wait before reporting, which looks like a local audio-permission problem rather than anything in this change — so I am leaning on CI for the re-verification rather than claiming a local pass.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppState.swift
+++ b/app/MeetingTranscriber/Sources/AppState.swift
@@ -452,7 +452,17 @@ final class AppState {
             activeJobState: pipeline.queue.activeJobs.first?.state,
             updateAvailable: updateChecker.availableUpdate != nil,
             permissionProblem: permissions.health?.isHealthy == false,
+            captureProblem: hasCaptureProblem,
         )
+    }
+
+    /// Whether a live recording is failing to capture one of its channels.
+    /// Drives the blinking error badge — the only failure surface Do Not
+    /// Disturb cannot suppress.
+    var hasCaptureProblem: Bool {
+        channelHealth.appSilentActive
+            || channelHealth.micSilentActive
+            || channelHealth.recordingSilentActive
     }
 
     var currentStateLabel: String {
--- a/app/MeetingTranscriber/Sources/MenuBarIcon.swift
+++ b/app/MeetingTranscriber/Sources/MenuBarIcon.swift
@@ -23,10 +23,27 @@ enum BadgeKind: String, CaseIterable, Codable {
     case updateAvailable
 
     /// Whether this badge kind uses animation.
+    ///
+    /// `.error` and `.userAction` animate so their exclamation BLINKS. They are
+    /// not progress states — nothing is moving underneath — but a static badge
+    /// is exactly what a user misses. Every silent failure on 2026-08-04
+    /// (missing Accessibility grant, suppressed consent prompt, a capture tap
+    /// recording an hour of silence) surfaced only through notifications, which
+    /// Do Not Disturb suppressed. The menu bar is the one surface DND cannot
+    /// mute, so it has to move when the app cannot do its job.
     var isAnimated: Bool {
         switch self {
         case .recording, .transcribing, .diarizing, .processing: true
+        case .error, .userAction: true
 
+        default: false
+        }
+    }
+
+    /// Attention states blink their overlay rather than animating a body.
+    var blinksOverlay: Bool {
+        switch self {
+        case .error, .userAction: true
         default: false
         }
     }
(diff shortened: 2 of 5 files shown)


pasrom:
Thank you for this, and an apology first, because you are owed more than one.

This sat for a month with no reply. The review below was done on 2026-08-10, the day you opened it, and never posted.

Worse: ten hours after you opened this, #602 was opened with your framing. "Every way this app can lose a recording is reported by notification and nothing else, and a Focus mode suppresses that notification without a trace" is your argument from this PR, and the 62-minute interview it cites is the incident you reported on #524 on 2026-08-04. That work shipped. It credits neither you nor this PR, in the body or in any commit. That should not have happened, and I am sorry.

## Where it leaves this PR

The Do Not Disturb half of your premise was fixed by the change your PR prompted. An app channel that goes to digital silence while the mic carries speech, and a capture that gives up entirely, now post as time-sensitive notifications. That ships in the Homebrew build from v0.8.0 (2026-09-07); the App Store build does not carry the entitlement, so there it is still an ordinary banner.

The menu bar half was already covered when you filed, which nothing in the UI told you. A silent channel has been tinted red since May 2026: `micSilentOverlay` fills the top half of the waveform, `appSilentOverlay` the bottom half, painted over the badge rather than replacing it. On that build your interview would have shown a red bottom half from roughly two minutes in until the end, once the app channel latches after its 90-second debounce. The badge kind stayed `.recording`, exactly as you wrote. That the tint exists and explains nothing about itself is a discoverability problem in our code, not an error in your report.

## What the review found in the diff

Measured on your head commit, comparing rendered PNG bytes:

```
userAction distinct frames = 6 of 6
error distinct frames      = 2 of 6
userAction[0] == inactive[0]    -> true
userAction[3] == recording[3]   -> true
```

`renderImage` draws the exclamation mark only under `permissionOverlay || badge == .error`. `.userAction` satisfies neither, so `blinksOverlay` is dead code there, while `isAnimated = true` sends it through the live recording animation. Hence the last line: "name the speakers" would render byte-identical to an in-progress recording. "The badge body stays still; only the overlay blinks" holds for `.error`, not for `.userAction`.

`testRecordOnlyOverlayDoesNotAnimateStaticBadges` and `MenuBarIconNextFrameTests.testStaticBadgesDoNotAdvance` both moved from a fixed badge list to `where !badge.isAnimated`, which drops the two newly animated badges out of the assertion. You said so openly in the PR text, so this is not about concealment; the point is that the derived form can no longer fail for the case it was written for, and restoring the original assertion turns the branch red.

## What holds up

`.error` blinks correctly: mark present in frames 0 to 2, gone in 3 to 5, body static within the phase. If you still want that, a PR animating only `.error`, with the original assertions intact, has a good case.

These measurements are from August against a base that has since moved by 200 commits, so they would need re-running.

Closing this for the reasons above, with thanks for the report and for the detail in it.

PR #601, opened 2026-08-10
fix(app): detect Aside browser meetings

maximilianwuehr-afk:
## What changed

Add Aside to the Chromium-family browser detection path.

Aside is Chromium-based and exposes the same `NoIdleSleepAssertion` named `WebRTC has active PeerConnections` as Chrome, Brave, Edge, and Chromium, but its process name is `Aside`. The detector and browser owner mapping now include it, with matching tests and user-facing documentation.

## Why

Meeting Transcriber could not detect an active Google Meet session running in Aside because the power assertion detector did not recognize the `Aside` process name.

Browser detection remains opt-in and continues to require recording consent.

## Validation

- `cd app/MeetingTranscriber && swift test --parallel --filter PowerAssertionDetectorPatternsTests` — 14 tests passed.
- `cd app/MeetingTranscriber && swift test --parallel --skip MenuBarIconSnapshotTests` — all 2,523 tests passed.
- `./scripts/pre-push.sh --with-tests` — release build and test-target build passed.
- `./scripts/lint.sh --lint-only` — all 428 Swift files passed SwiftLint.
- Changed-file SwiftFormat check passed.
- `swift test --parallel` — one existing `MenuBarIconSnapshotTests/testStaticBadgeSnapshots` reference-image mismatch remains under this local Xcode/OS environment; no changed menu-bar files are involved.
- Repository-wide SwiftFormat currently reports 99 existing files requiring formatting with the installed formatter version; this PR does not reformat unrelated files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ flowchart TD
 
 ## Features
 
-- **Automatic meeting detection** — Recognizes Teams, Zoom, and Webex meetings via window title polling, opt-in browser meeting detection (Google Meet, Whereby, web Zoom/Teams in Chrome) gated behind a recording-consent prompt, and opt-in mic-input detection for call apps without a reliable meeting signal (WeChat, Tencent Meeting, FaceTime, WhatsApp) — per app, off by default, and unlike the browser path it starts recording without a prompt
+- **Automatic meeting detection** — Recognizes Teams, Zoom, and Webex meetings via window title polling, opt-in browser meeting detection (Google Meet, Whereby, web Zoom/Teams in Chrome, Brave, Edge, Chromium, or Aside) gated behind a recording-consent prompt, and opt-in mic-input detection for call apps without a reliable meeting signal (WeChat, Tencent Meeting, FaceTime, WhatsApp) — per app, off by default, and unlike the browser path it starts recording without a prompt
 - **Dual audio recording** — App audio ([CATapDescription](https://developer.apple.com/documentation/coreaudio/catap)) + microphone simultaneously
 - **On-device transcription** — Two engines, selectable in Settings:
   - [WhisperKit](https://github.com/argmaxinc/WhisperKit) — 99+ languages, ~1 GB model
--- a/app/MeetingTranscriber/Sources/MeetingPatterns.swift
+++ b/app/MeetingTranscriber/Sources/MeetingPatterns.swift
@@ -103,13 +103,13 @@ extension AppMeetingPattern {
     /// (issue #503). Detection is by the WebRTC power assertion (see
     /// `PowerAssertionDetector`), not window titles — a browser's window title
     /// only reflects the active tab. Every Chromium fork (Chrome, Brave, Edge,
-    /// Chromium) emits the same assertion, so all four owner names share this one
+    /// Chromium, Aside) emits the same assertion, so all five owner names share this one
     /// browser identity (one "Google Chrome" toggle). `requiresRecordingConsent`
     /// routes it through a prompt instead of auto-start; `meetingPatterns` is
     /// empty (no title-based detection in the PoC).
     static let chromeBrowser = AppMeetingPattern(
         appName: "Google Chrome",
-        ownerNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium"],
+        ownerNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium", "Aside"],
         meetingPatterns: [],
         requiresRecordingConsent: true,
     )
--- a/app/MeetingTranscriber/Sources/PowerAssertionDetector.swift
+++ b/app/MeetingTranscriber/Sources/PowerAssertionDetector.swift
@@ -59,7 +59,7 @@ class PowerAssertionDetector: MeetingDetecting {
         // "NoIdleSleepAssertion" named "WebRTC has active PeerConnections"
         // during any WebRTC call (Google Meet, Whereby, web Zoom/Teams/Webex).
         // The assertion lives in Chromium's content layer, so every fork
-        // (Chrome, Brave, Edge, Chromium) emits the same name — all four are
+        // (Chrome, Brave, Edge, Chromium, Aside) emits the same name — all five are
         // matched under one shared browser identity. Keyword-only, no
         // assertionTypes: a Chromium browser holds the same assertion type for
         // plain media playback (YouTube), so matching the type would fire on
@@ -67,7 +67,7 @@ class PowerAssertionDetector: MeetingDetecting {
         // only watched when the browser toggle adds "Google Chrome" to watchApps.
         AssertionPattern(
             appName: AppMeetingPattern.chromeBrowser.appName,
-            processNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium"],
+            processNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium", "Aside"],
             keywords: ["webrtc", "peerconnection"],
         ),
     ]
--- a/app/MeetingTranscriber/Sources/Settings/GeneralSettingsView.swift
+++ b/app/MeetingTranscriber/Sources/Settings/GeneralSettingsView.swift
@@ -41,11 +41,14 @@ struct GeneralSettingsView: View {
                 Toggle("Tencent Meeting", isOn: $settings.watchTencentMeeting)
                 Toggle("FaceTime", isOn: $settings.watchFaceTime)
                 Toggle("WhatsApp", isOn: $settings.watchWhatsApp)
-                Toggle("Browser Web Meetings (Chrome, Brave, Edge)", isOn: $settings.watchBrowserMeetings)
+                Toggle("Browser Web Meetings", isOn: $settings.watchBrowserMeetings)
                     .accessibilityIdentifier(A11yID.watchBrowserToggle)
-                Text("Detects meetings in Chromium browsers such as Chrome, Brave, and Edge (Google Meet, Whereby, web Zoom/Teams). Asks before recording.")
-                    .font(.caption)
-                    .foregroundStyle(.secondary)
+                Text(
+                    "Detects meetings in Chromium browsers such as Chrome, Brave, Edge, and Aside "
+                        + "(Google Meet, Whereby, web Zoom/Teams). Asks before recording.",
+                )
+                .font(.caption)
+                .foregroundStyle(.secondary)
                 browserConsentWarning
             }
 
--- a/app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
+++ b/app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
@@ -85,16 +85,16 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
         XCTAssertNil(detector.checkOnce())
     }
 
-    // MARK: - Chromium-family browsers (Brave / Edge / Chromium) — issue #503 follow-up
+    // MARK: - Chromium-family browsers (Brave / Edge / Chromium / Aside) — issue #503 follow-up
 
     /// The browser pattern is not Chrome-specific: every Chromium fork holds the
     /// same "WebRTC has active PeerConnections" assertion (it lives in Chromium's
-    /// content layer), so Brave/Edge/Chromium calls must detect identically. They
+    /// content layer), so Brave/Edge/Chromium/Aside calls must detect identically. They
     /// share the single "Google Chrome" browser-meetings identity (one toggle),
     /// but the detected meeting is titled by the concrete browser so a Brave call
     /// is not mislabeled "Google Chrome Call".
     func testChromiumFamilyWebRTCCallsAreDetected() {
-        for process in ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium"] {
+        for process in ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium", "Aside"] {
             let detector = PowerAssertionDetector(
                 patterns: PowerAssertionDetector.patterns(watching: ["Google Chrome"]),
                 confirmationCount: 1,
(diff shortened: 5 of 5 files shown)


pasrom (commented):
Thanks for this, and thanks especially for the validation section: full suite, lint, and a release-parity build is more than most detection patches arrive with. Two things you flagged are known and not yours to fix: the menu-bar snapshot reference is dev-environment dependent (those tests are skipped in CI on purpose), and the repository-wide SwiftFormat count comes from a newer formatter than the version CI pins, so leaving unrelated files alone was the right call.

The mechanics of the change are right: this is exactly the two lists a new fork has to join, the tests extend the two loops that exist for the family, and the user-facing text was updated along with them. My concern is not the shape of the patch, it is that its one load-bearing fact cannot be checked anywhere in this repository.

**Blocking, one item.** `"Aside"` is matched by exact, case-sensitive string equality, and nothing here can falsify it. The two extended test loops inject the same literal they assert against, so they pass whatever string is in the list. The browser e2e lane only accepts chrome, brave, edge and chromium, so no lane can drive Aside. And if the string is wrong the failure is completely silent: the "watched app is running but its assertion did not match" log line is itself gated on the allowlist, so a name that is not already in the list produces no log at all. The result of a wrong literal would be a user who enables the toggle, joins a Meet call in Aside, records nothing, and has three pieces of documentation telling them it is supported.

You have Aside installed and we do not, so the cheapest way to settle this is a paste from your machine. Two values, because the detector and the window-title lookup read two different name sources (details in the inline comments):

1. While you are in a Google Meet call in Aside:

```
pmset -g assertions | grep -i -B2 -A6 aside
```

This shows the process name as `IOPMCopyAssertionsByProcess` reports it, plus the assertion name and type, so it proves the whole match in one paste.

2. The name the window server attributes windows to:

```
osascript -e 'tell application "System Events" to get name of every process whose background only is false' | tr ',' '\n' | grep -i aside
```

If both come back as literally `Aside`, this is good to merge with the three small doc corrections below.

**Not your problem, tracked separately.** The whole Chromium family shares a single `"Google Chrome"` identity, and several things key on that identity rather than on the concrete browser (which browser a consent decline is remembered against, which browser keeps a recording considered live, which process the audio tap targets when two forks assert at once, and which window supplies the meeting title). Those predate this PR. They were mostly unreachable because the existing forks are browsers people run instead of Chrome, whereas Aside is designed to sit alongside one, so your patch makes them easier to hit. I will open a separate issue for that and will not ask you to touch it here.

I am also going to widen that miss diagnostic separately so it logs any unmatched WebRTC assertion rather than only allowlisted ones. That is the piece that would have let you prove the process name from a log instead of by inspection, and it makes the next fork discoverable without a source edit.

app/MeetingTranscriber/Sources/PowerAssertionDetector.swift
```diff
@@ -59,15 +59,15 @@ class PowerAssertionDetector: MeetingDetecting {
...
+        // (Chrome, Brave, Edge, Chromium, Aside) emits the same name — all five are
         // matched under one shared browser identity. Keyword-only, no
         // assertionTypes: a Chromium browser holds the same assertion type for
         // plain media playback (YouTube), so matching the type would fire on
         // every video — the WebRTC name is the meeting-specific signal. Opt-in:
         // only watched when the browser toggle adds "Google Chrome" to watchApps.
         AssertionPattern(
             appName: AppMeetingPattern.chromeBrowser.appName,
-            processNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium"],
+            processNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium", "Aside"],
```
pasrom, line 70: The match is exact and case-sensitive (`pattern.processNames.contains(processName)`; only the assertion *name* is lowercased, never the process name), so this literal has to be byte-exact with what `IOPMCopyAssertionsByProcess` reports in the `Process Name` key. The four names already here each have a nightly e2e leg behind them, and that lane's own comment records why: the label it takes is also the process name the fork holds its assertion under. Since the lane rejects anything outside chrome, brave, edge and chromium, `"Aside"` cannot get the same treatment, and the test loop below cannot substitute for it because it feeds this same literal back in.

Could you paste `pmset -g assertions` from an active Meet call in Aside? If the reported name is anything other than exactly `Aside` (a different capitalisation, a suffix like `Aside Browser`, or a helper process), detection will never fire and nothing will say so.

app/MeetingTranscriber/Sources/MeetingPatterns.swift
```diff
@@ -103,13 +103,13 @@ extension AppMeetingPattern {
...
     /// only reflects the active tab. Every Chromium fork (Chrome, Brave, Edge,
-    /// Chromium) emits the same assertion, so all four owner names share this one
+    /// Chromium, Aside) emits the same assertion, so all five owner names share this one
     /// browser identity (one "Google Chrome" toggle). `requiresRecordingConsent`
     /// routes it through a prompt instead of auto-start; `meetingPatterns` is
     /// empty (no title-based detection in the PoC).
     static let chromeBrowser = AppMeetingPattern(
         appName: "Google Chrome",
-        ownerNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium"],
+        ownerNames: ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium", "Aside"],
```
pasrom, line 112: Worth knowing that this list is not a copy of the one in the detector even though it currently holds the same strings: `ownerNames` is compared against `kCGWindowOwnerName` from the window list, while `processNames` is compared against the power-assertion process name. Those two are usually equal but not always, and this repository already carries a case where they differ: the Teams pattern deliberately keeps `ownerNames` and its process names apart for exactly that reason.

So the second value in the review body matters independently of the first. If Aside asserts as `Aside` but its windows are attributed to something else, detection fires and the meeting is recorded, but the title lookup never matches an Aside window and every Aside meeting is titled by the placeholder forever, with a green test suite.

Optional, and only once both names are confirmed equal: this list and the detector's could become one, since the initializer there already reaches for `AppMeetingPattern.chromeBrowser.appName` on the line above and could take `.ownerNames` the same way. That would collapse two of the four hand-synced copies of this array and make the guard test that polices their equality unnecessary. If the two names turn out to differ for Aside, ignore this: keeping the lists separate is then the correct answer.

README.md
```diff
@@ -81,7 +81,7 @@ flowchart TD
 
 ## Features
 
-- **Automatic meeting detection** — Recognizes Teams, Zoom, and Webex meetings via window title polling, opt-in browser meeting detection (Google Meet, Whereby, web Zoom/Teams in Chrome) gated behind a recording-consent prompt, and opt-in mic-input detection for call apps without a reliable meeting signal (WeChat, Tencent Meeting, FaceTime, WhatsApp) — per app, off by default, and unlike the browser path it starts recording without a prompt
+- **Automatic meeting detection** — Recognizes Teams, Zoom, and Webex meetings via window title polling, opt-in browser meeting detection (Google Meet, Whereby, web Zoom/Teams in Chrome, Brave, Edge, or Aside) gated behind a recording-consent prompt, and opt-in mic-input detection for call apps without a reliable meeting signal (WeChat, Tencent Meeting, FaceTime, WhatsApp) — per app, off by default, and unlike the browser path it starts recording without a prompt
```
pasrom: This bullet used to read "web Zoom/Teams in Chrome", which scanned as an example. As an explicit four-name list it now reads as exhaustive, and it drops Chromium, which the detector does match. A Chromium user reading this would conclude their browser is unsupported and leave the toggle off, losing detection that already works today. Please add Chromium to the list.

app/MeetingTranscriber/Sources/Settings/GeneralSettingsView.swift
```diff
@@ -41,11 +41,14 @@ struct GeneralSettingsView: View {
                 Toggle("Tencent Meeting", isOn: $settings.watchTencentMeeting)
                 Toggle("FaceTime", isOn: $settings.watchFaceTime)
                 Toggle("WhatsApp", isOn: $settings.watchWhatsApp)
-                Toggle("Browser Web Meetings (Chrome, Brave, Edge)", isOn: $settings.watchBrowserMeetings)
+                Toggle("Browser Web Meetings (Chrome, Brave, Edge, Aside)", isOn: $settings.watchBrowserMeetings)
```
pasrom: Optional, and pre-existing rather than introduced here: this label has omitted Chromium since it was written. The caption below is fine, because "Chromium browsers such as ..." names the family and is explicitly non-exhaustive. Now that the label is a four-item list it would be easy to read as complete, so adding Chromium (or dropping the parenthetical in favour of the caption) would make the two agree. Not blocking.

app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
```diff
@@ -94,7 +94,7 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
     /// but the detected meeting is titled by the concrete browser so a Brave call
     /// is not mislabeled "Google Chrome Call".
     func testChromiumFamilyWebRTCCallsAreDetected() {
-        for process in ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium"] {
+        for process in ["Google Chrome", "Brave Browser", "Microsoft Edge", "Chromium", "Aside"] {
```
pasrom, line 97: Two things here.

Extending these loops is the right move and I would keep them, but it is worth being explicit that they cannot fail for a wrong process name: the string being asserted is the string being injected, and both loops set `windowListProvider` to an empty list, so neither reaches the window-title path at all. What they do pin is that a new family member inherits the WebRTC keyword gate in both directions, which is genuinely worth having. The coverage report saying every modified line is covered reflects the static initializer running at load, not the new entry's effect being exercised.

Second, and this one is a straightforward fix: the MARK on line 88 and the doc comment on line 92 above this test still enumerate the family as "Brave / Edge / Chromium". Since you updated both source doc comments from four to five, these two are the last stale lists, and they are the ones the next person adding a fork will read first.

pasrom:
Thanks, the fixup covers all three documentation points. Dropping the parenthetical from the toggle label entirely was the better of the two options I offered: the caption below already names the family non-exhaustively, so there is now one fewer list that can go stale.

On the remaining item I did part of the work from this side. I installed Aside 1.0.728.1 from the official Homebrew cask and measured the two identity values directly:

- `CFBundleExecutable`, `CFBundleName` and `CFBundleDisplayName` are all `Aside` (bundle id `at.studio.AsideBrowser`), and the running main process shows as `Aside` in `ps`.
- `kCGWindowOwnerName`, read live from `CGWindowListCopyWindowInfo` for an open Aside window, is `Aside`.

So the `ownerNames` entry is confirmed and that half of my question is closed. Aside is also genuinely Chromium based, its process list carries `chrome_crashpad_handler` and `Aside Helper (Renderer)`.

What I could not settle is whether Aside emits the `WebRTC has active PeerConnections` assertion at all, and that is a limitation of my setup rather than an observation about Aside. I drove a local loopback `RTCPeerConnection` with a canvas capture track from a page served on localhost. In Chrome that page takes the assertion within two seconds. In Aside the same page runs (I instrumented it to report back, and both the create and the negotiate step arrive) but the connection never advances past negotiated to connected, so there is never an active peer connection for the assertion to attach to. Aside sits behind its sign-in flow here and I did not push through it.

That leaves one value open, and it is the one the pattern actually keys on. From an active call in Aside:

```
pmset -g assertions | grep -i -B2 -A6 aside
```

What would close this out is a line of the shape `pid NNNN(Aside): ... NoIdleSleepAssertion named: "WebRTC has active PeerConnections"`. The prediction is well supported now, since every other fork holds this assertion under its main process and that process is named `Aside`, but it is still a prediction, and this is the one link in the chain that fails silently if it is wrong.

If Aside turns out to hold it under a helper process instead, that is a useful answer too and the entry simply takes that name.

app/MeetingTranscriber/Sources/PowerAssertionDetector.swift
maximilianwuehr-afk: Confirmed during an active Google Meet call in Aside. Relevant `pmset -g assertions` output:

```text
pid 729(Aside): [0x0001e1c900019826] 00:04:34 NoIdleSleepAssertion named: "WebRTC has active PeerConnections"
pid 729(Aside): [0x0001e1d20005991d] 00:04:25 NoDisplaySleepAssertion named: "Video Wake Lock"
```

The assertion process name is exactly `Aside`, with the expected WebRTC assertion.

app/MeetingTranscriber/Sources/MeetingPatterns.swift
maximilianwuehr-afk: Confirmed independently during the same call:

```text
$ osascript -e 'tell application "System Events" to get name of every process whose background only is false' | tr ',' '\n' | grep -i aside
Aside
```

The window-owner name is also exactly `Aside`. I kept `ownerNames` and `processNames` separate because they represent different OS data sources, even though their Aside values currently match.

README.md
maximilianwuehr-afk: Fixed in `7e7ac25`: the exhaustive browser list now includes Chromium.

app/MeetingTranscriber/Sources/Settings/GeneralSettingsView.swift
maximilianwuehr-afk: Fixed in `7e7ac25` by dropping the exhaustive parenthetical from the toggle label. The caption remains the non-exhaustive explanation of supported Chromium browsers.

app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
maximilianwuehr-afk: Fixed in `7e7ac25`: both the MARK and doc comment now include Aside. The tests remain focused on the shared WebRTC keyword gate; the machine output above provides the process-name proof.

pasrom:
That closes it. `pid 729(Aside): NoIdleSleepAssertion named: "WebRTC has active PeerConnections"` is exactly the evidence the pattern needed, and the independent owner-name check settles the other list. Thank you for going and getting a real call rather than arguing the point.

Two things I appreciated beyond the answer itself. You included the `NoDisplaySleepAssertion named: "Video Wake Lock"` line I did not ask for, which is the same assertion a Teams WebView holds with no call in progress. It is harmless here because the browser pattern is keyword-only on `webrtc`/`peerconnection` and that name carries neither, but it is the detail that shows the mechanism was understood rather than just the request.

And you were right to decline the list merge. Two different OS data sources that happen to agree today is not the same as one source, and collapsing them would have traded a real distinction for a cosmetic saving. My comment made it conditional for that reason, and keeping them apart is the answer I would have picked too.

One thing before this can merge: please fold `7e7ac25` into `9cfd7ad`. The repository merges by rebase, so a literal `fixup!` commit would land on main as-is. `git rebase -i --autosquash` and a force push will do it. I assume that was your intent given you shaped it as a fixup commit in the first place.

Also so it is not a mystery on your side: the workflow runs show as awaiting approval because that is how this repository treats pull requests from forks. They need a maintainer to release them, and that happens per commit, so I will do it after the squash rather than twice.

pasrom (approved):

Issue #581, opened 2026-08-05
Mic and app-audio tracks aren't deduplicated when merged — a speaker with any mic/speaker echo gets every line transcribed twice


GB-D:
## Summary

When a remote participant's audio bleeds into the local microphone (e.g. no headphones, speaker output picked up by the mic), that participant's speech gets captured independently on both the `mic` track and the `app`/system-audio track. The merged transcript doesn't deduplicate these — every affected utterance appears twice, back-to-back, attributed to the same speaker.

## Evidence from a real call

52-minute Teams meeting, 3 participants. One participant's mic-track and app-track speech was diarized/named separately (`recognition_log.jsonl` shows the same person resolved under one mic-track label and three separate app-track labels). Counting exact-duplicate consecutive lines in the final transcript by speaker:

| Speaker | Total lines | Duplicate lines |
|---|---|---|
| Participant A (local, mic only) | 87 | 0 |
| Participant B (local, mic only) | 128 | 0 |
| Participant C (remote) | 1,319 | 206 (~16%) |

Every single duplicate belongs to the remote participant. Example (consecutive lines in the raw transcript):

```
[19:18] Speaker: But if it if it if it if it did come to that and say we did have like a
[19:18] Speaker: But if it if it if it if it did come to that and say we did have like a
```

The recognition log confirms the mechanism: that participant is named on the mic track (should only ever contain the local user's own voice) *and* on three separate app-track diarization clusters — consistent with their voice being picked up twice (once via system/app audio, once via echo into the mic), likely also explaining why their app-track voice fragmented into multiple clusters rather than one clean one.

## Impact

- Raw transcripts for affected calls are bloated ~15-20% with exact duplicate lines for the affected speaker.
- In this case the downstream AI-generated summary wasn't visibly degraded, but the duplication seems like pure waste at best and a plausible source of summary drift at worst (a duplicated line could shift word-count-based heuristics, or in principle get counted twice by anything downstream that isn't purely LLM-summarized).

## Suggested fix

When merging mic-track and app-track transcriptions into one chronological output, dedupe adjacent/near-simultaneous lines with matching (or near-matching, e.g. normalized/fuzzy-matched) text before final assembly — similar text within a small time window from two different tracks is far more likely to be the same utterance captured twice than two genuinely separate utterances.

Happy to share the full anonymized transcript/log excerpts if useful for reproduction.

Pred8or:
[...]

Issue #422, opened 2026-06-11
Echo when recording Slack or Zed


wereHamster:
When I manually select Slack app to record (a Slack huddle), my voice is duplicated, like as if there was a echo. I suspect it's because MeetingTranscriber captures two audio streams, one with a slight delay, perhaps due to how Slack app works internally. Is there a way to fix that? I noticed the same echo issue in Zed (https://zed.dev/) when I join a channel there.

Version 0.5.1 (10d7b00) · Homebrew; macOS Sequoia (15.7.7)

pasrom:
Thanks for the report! Your suspicion is likely correct, and I'd like to narrow down the exact mechanism with a few questions.

**Background:** MeetingTranscriber records two separate streams: a CoreAudio tap on the app's audio output (everything Slack/Zed plays), and your microphone directly. Both are transcribed and merged into one transcript. The architecture assumes the app's output contains only the *remote* side. If the app mixes your own microphone into its output (e.g. for level metering or monitoring), your voice ends up in both streams and appears twice in the transcript, slightly offset — exactly the echo you describe. For Zed there is some evidence pointing in this direction: zed-industries/zed#21649 reports audible own-voice feedback in collaboration calls.

To distinguish the possible causes, could you check:

1. **Whose voice is duplicated** — your own, the other participants', or both? And do you use headphones or speakers? (Without headphones, the *remote* voices get picked up acoustically by your mic and duplicated; your *own* voice can only duplicate if the app loops it back, or if a remote participant's echo cancellation fails.)
2. **With headphones on and MeetingTranscriber not running:** do you hear yourself in Slack/Zed at all?
3. Next to the transcript, the raw tracks are kept in `~/Library/Application Support/MeetingTranscriber/recordings/`. If you listen to the `<timestamp>_app.wav` of an affected recording: **is your own voice audible in it?** If yes, roughly how far behind your real speech (a few tens of ms vs. half a second)? That tells us whether it's a local loopback inside the app or a network echo from another participant.

**Workaround for now:** if your own voice is indeed in the app track, enable **Settings → Audio → "No Microphone (app audio only)"**. The transcript is then built from the app audio alone, which already contains both sides of the call — no duplication. Downside: your own utterances are no longer automatically labeled "Me" (voice enrollment can still name them).

Depending on what you find, the proper fix would be either deduplicating overlapping identical speech between the two tracks at merge time, or an option to disable the mic channel only for specific apps that already include it.

[...]

PR #603, opened 2026-08-10
feat(app): gate mic-track echo before transcription (opt-in)

Pred8or:
Without a headset, the microphone re-captures the remote side's loudspeaker playback. `AudioMixer.suppressEcho` already gates exactly that kind of mic energy — but it only runs inside `mix()`, and on a dual-source recording the mix is not what gets transcribed; the raw per-track files are. So the gate never reaches the path that matters, and the echo arrives as duplicated speech attributed to the mic speaker, with its embeddings polluting speaker matching.

Refs #422, #581.

## Measurement

17-minute two-party Teams call, built-in mic and speakers, no headset. Method: over the raw `_app.wav` / `_mic.wav` pair, take 1-second blocks where the app track is loud and the mic track is quieter than it — i.e. only the remote side is speaking — then cross-correlate app→mic per block with the lag constrained to 0–120 ms. 310 blocks qualified.

| | |
|---|---|
| Remote voice in `_app.wav` | −19.7 dBFS (median) |
| Same voice in `_mic.wav` | −22.8 dBFS (median) |
| Attenuation | **3.2 dB** |
| Cross-correlation | median 0.44, top decile 0.69, peak 0.92 |
| Lag | ~3 ms |

The 3 dB is the part worth stressing: the remote side arrives in the microphone at nearly the same level as in the app tap. That is not a faint bleed — it is a second, near-full copy of the far end, and both copies get transcribed.

## What this does

New config-only setting `dualTrackMicEchoSuppressionEnabled` (default off, no UI):

```
defaults write app.meetingtranscriber dualTrackMicEchoSuppressionEnabled -bool true
```

When on, the existing gate runs over `mic_16k.wav` right after resampling, so transcription, mic diarization, and the persisted `_mic_16k.wav` used for late re-diarization all read the cleaned track. `AudioMixer.suppressEchoInFile` is a file-level wrapper around the existing sample-level `suppressEcho`; the delay handling matches `mix()`.

## Trade-off

This is an energy gate, not an echo-similarity detector — unchanged from what `mix()` already does. It zeroes each 20 ms mic window whose corresponding app window is above threshold, plus a margin, so genuine double-talk on the mic side is gated out along with the echo. That is why it is opt-in rather than on by default.

## Commits

1. the gate itself, plus tests covering the clamp and empty-track cases
2. a doc fix — the setting's example named the pre-rename bundle identifier, which would have written the key into a domain nothing reads
3. the `defaultsToFalse` / `persistsToUserDefaults` pair the other settings have

I have been running this on my own recordings for two and a half weeks.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -182,6 +182,16 @@ final class AppSettings {
         didSet { defaults.set(micDeviceUID, forKey: "micDeviceUID") }
     }
 
+    /// Gate the mic track with `AudioMixer.suppressEcho` before transcription
+    /// and diarization (dual-source recordings only). Without a headset the mic
+    /// re-captures the loudspeaker playback of the remote side; `mix()` already
+    /// gates that echo, but transcription reads the raw per-track files, not
+    /// the mix. Config-only, no UI (issue #422):
+    /// `defaults write app.meetingtranscriber dualTrackMicEchoSuppressionEnabled -bool true`
+    var dualTrackMicEchoSuppressionEnabled: Bool {
+        didSet { defaults.set(dualTrackMicEchoSuppressionEnabled, forKey: "dualTrackMicEchoSuppressionEnabled") }
+    }
+
     /// Master switch for the per-channel signal indicator. When on, AppState runs a
     /// ~10 Hz level poller while recording and flips the menu-bar red when one
     /// channel goes silent while the other carries audio. Default: on.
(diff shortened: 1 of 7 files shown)


pasrom:
Thank you for this, and sorry for the long silence: four weeks without a reply to a measured, self-hosted patch is not what the work deserved.

Your diagnosis is the one the fix was built on. `AudioMixer.suppressEcho` runs inside `mix()`, and on a dual-source recording the mix is not what gets transcribed, so the gate never reaches the path that matters. Your app-to-mic block correlation is close to what the detector now does per window.

What landed instead, released in v0.8.0: an acoustic canceller between detection and transcription, opt-in and off by default, with a self-check that has to confirm the run before its output is adopted. That check exists because on a minority of recordings the model completes normally and removes essentially nothing.

The reason it is not the RMS gate: the gate mutes the microphone while the app track is loud, so during double talk it removes the local speaker along with the echo. That is why it stayed inside `mix()` when it was written, and moving it onto the transcribed track carries the cost over. A transcript-level dedup was built too, and it ships off for a related reason: at the alignment the pipeline actually uses, it classifies a local speaker who only ever talks while the far end does as echo and drops the line. That is pinned in the test suite as an expected failure rather than papered over.

So I am closing this in favour of the acoustic path, with no criticism of the work: your measurement was better than anything the issue had before it.

If you are willing, the most useful next step is testing the cancellation on the recordings you already have. In Settings, Audio, the switch is "Remove echoed remote speech before transcribing". Note that the one below it, "Remove echoed remote speech from the transcript", is the dedup with the defect described above, so the two are easy to confuse. Then report whether the far end is gone from the microphone track and whether your own words survive double talk. A recording where it removes nothing is the case I would most like to see from another setup.

PR #605, opened 2026-08-10
feat(app): add /v1/watch so watching can be controlled from outside the app

alexpfau:
Adds `GET`/`POST /v1/watch` so watching can be started and stopped from outside the app, plus the two pieces needed to actually use it: an `mt-cli watch` subcommand and a guide for wiring a stream deck key or a hotkey to it.

Motivation: the automation API could already *report* watching state through `/state`, but nothing could change it. There was no URL scheme, no global hotkey and no App Intent either, so an external trigger of any kind was impossible.

## The endpoint

`GET /v1/watch` and `POST /v1/watch` return the same body:

```json
{"watching":true,"state":"watching","badge":"recording","manualRecording":false,"permissionsHealthy":true}
```

`POST` takes `{"action":"start"|"stop"|"toggle"}` and returns the state **after** the call.

Decisions worth reviewing:

* **`start`/`stop` are idempotent and offered alongside `toggle`, not replaced by it.** A button press expresses a desired outcome, not a delta. A client that blind-toggles after a meeting already ended stays inverted for the rest of the session.
* **The response is the resulting state**, so a caller never has to follow a POST with a GET and race the next change. This is why the control path is `async`.
* **`watching` is an explicit boolean** rather than something clients infer from `state == nil`. That inference is a fine in-process shortcut and a bad thing to put across a public boundary.
* **`badge` is included** so a single request tells a physical key more than on/off.
* **Blocked control returns 409.** `toggleWatching()` already early-returns while a manual app-picker recording owns the loop; silent refusal is tolerable in-process, not across an API.
* **Under `/v1`, not `/action/*`**, which is documented as free to change.
* **No `Idempotency-Key`** — the operation is idempotent by construction.

`stopWatching()` awaits any in-flight `startTask` first. Without that, a stop issued during the mic-access window lands before the loop exists, reports "already stopped", and is then undone by the start completing. `testStopWatchingDuringInFlightStartStillStops` pins it.

## What consumes it

A stream deck plugin that polls this endpoint and renders live state on the key already exists and is hardware-tested: [alexpfau/meeting-transcriber-streamdeck](https://github.com/alexpfau/meeting-transcriber-streamdeck). I deliberately kept it in its own repository rather than proposing it here, so this PR stays a single endpoint plus its client and docs, and you take on no maintenance for a Node/TypeScript plugin and its Elgato release cycle. It is submitted to the Elgato Marketplace and awaiting review; if you would rather it live under this project instead, I am happy to move it.

`docs/stream-deck.md` links to the plugin and also documents a plain Shortcuts.app recipe, so the feature is usable with nothing but the shell.

The docs are blunt about why `toggle` is the weakest of the three actions: a key bound to it desynchronises the first time watching stops from anywhere else, and the stream deck `Multi Action Switch` that looks like the fix has exactly the same flaw. Separate `start`/`stop` keys, or a plugin that reads state back, are the honest answers.

## Verification

* Every `/v1/watch` path exercised against a running app, cross-checked against `/state.watchState` in both directions.
* 13 new tests (7 route-level over real sockets, 6 controller-level).
* Full suite, lint and `pre-push.sh` clean.
* Notably, `start` does not block on a TCC prompt — measured at 2.0 s on a build with microphone `notDetermined`, which was the main risk to the design.
* End-to-end from real hardware: a physical key starts and stops watching and reflects recording state back within one poll interval.

Happy to follow up with a `scripts/` generator that emits a ready-made Shortcuts.app shortcut if that is wanted — it is written and working, but it is Shortcuts-specific and harder to test than the rest of this, so it seemed better kept out of the change that introduces the endpoint.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppState+RPC.swift
+++ b/app/MeetingTranscriber/Sources/AppState+RPC.swift
@@ -131,7 +131,7 @@
                 liveCaptions: liveCaptionsSnapshot(),
                 watchState: watching.watchLoop?.state.rawValue,
                 pendingConsentApp: watching.watchLoop?.pendingConsentApp,
-                isManualRecording: isManualRecording,
+                isManualRecording: watching.isManualRecording,
                 notifications: notificationsSnapshot(),
                 // Built inside AppSettings (single-hop `self.` reads) to stay
                 // under the type-check budget — see `rpcSettingsSnapshot`.
@@ -140,6 +140,71 @@
                 updateStatus: updateStatusSnapshot(),
                 windows: windowsSnapshot(),
             )
+        }
+
+        /// Small, stable projection of the watching lifecycle for `/v1/watch`.
+        ///
+        /// Kept separate from `rpcStateSnapshot` on purpose: that snapshot is the
+        /// debug surface and is free to change shape, while this one is polled on
+        /// an interval by third-party controllers and carries a compatibility
+        /// promise. Locals are bound first (rather than reading the two-hop
+        /// `watching.watchLoop.…` `@Observable` chain inline) to keep the
+        /// memberwise init inside the type-check budget — see `rpcQueueStatus`.
+        func watchStatusDTO() -> WatchStatusDTO {
+            let loop = watching.watchLoop
+            let badge = currentBadge.rawValue
+            // nil means the probe has not run, which is not a permission
+            // problem. `BadgeKind.compute` treats it the same way, and the
+            // badge and this field must not disagree.
+            let healthy = permissions.health?.isHealthy != false
+            return WatchStatusDTO(
+                watching: watching.isWatching,
+                state: loop?.state.rawValue,
+                badge: badge,
+                manualRecording: watching.isManualRecording,
+                pendingConsentApp: loop?.pendingConsentApp,
+                permissionsHealthy: healthy,
+            )
+        }
+
+        /// The two `/v1/watch` seams, bundled so `buildDebugRPCServer` stays a
+        /// flat wiring list. Control is `async` because the start awaits the mic
+        /// gate before the loop exists — reporting a snapshot taken before that
+        /// settles would hand a remote key a stale answer to the press it just
+        /// made. Status is this small dedicated projection rather than `/state`,
+        /// which a polling controller must not be pinned to.
+        func watchRPCClosures() -> (
+            status: () -> WatchStatusDTO,
+            control: (WatchAction) async -> WatchControlOutcome,
+        ) {
+            let status: () -> WatchStatusDTO = { [weak self] in
+                self?.watchStatusDTO() ?? .notWatching
+            }
+            let control: (WatchAction) async -> WatchControlOutcome = { [weak self] action in
+                guard let self else { return .failed }
+                return await watching.applyWatchAction(action)
+            }
+            return (status, control)
+        }
+
+        /// The three per-job speaker-naming seams, bundled for the same reason as
+        /// `watchRPCClosures`: `buildDebugRPCServer` is a flat wiring list, and
+        /// grouping the closures that belong to one route family keeps it one.
+        func namingRPCClosures() -> (
+            status: (UUID) -> NamingStatusDTO?,
+            confirm: (UUID, [String: String]) -> Bool,
+            skip: (UUID) -> Bool,
+        ) {
+            let status: (UUID) -> NamingStatusDTO? = { [weak self] id in
+                self?.pipeline.namingStatus(forID: id)
+            }
+            let confirm: (UUID, [String: String]) -> Bool = { [weak self] id, mapping in
+                self?.pipeline.confirmNaming(jobID: id, mapping: mapping) ?? false
+            }
+            let skip: (UUID) -> Bool = { [weak self] id in
+                self?.pipeline.skipNaming(jobID: id) ?? false
+            }
+            return (status, confirm, skip)
         }
 
         /// Project the pinning-relevant properties of each named scene window.
--- a/app/MeetingTranscriber/Sources/WatchStatusDTO.swift
+++ b/app/MeetingTranscriber/Sources/WatchStatusDTO.swift
@@ -1,0 +1,88 @@
+import Foundation
+
+/// Wire shape for `GET`/`POST /v1/watch` — the meeting-watching lifecycle as a
+/// small, stable projection.
+///
+/// Deliberately *not* `/state`: that snapshot is the debug surface, a 15-section
+/// kitchen sink built for E2E drivers and documented as free to change. A
+/// third-party controller (Stream Deck key, Shortcut, shell script) polls this
+/// on an interval forever, so it needs a payload that is cheap to fetch and
+/// carries a compatibility promise — the same reasoning that split `/v1` out of
+/// `/action/*` in the first place.
+///
+/// `watching` is an explicit boolean rather than something the client infers
+/// from a nil `state`: that inference is a semantic subtlety which has no
+/// business crossing a public API boundary. `badge` is included because it is
+/// the single richest glanceable field — it folds transcribing/diarizing/
+/// processing/error/updateAvailable into one value, which is what lets a
+/// physical key render more than on/off.
+struct WatchStatusDTO: Codable, Equatable {
+    /// Whether the watch loop is running *and* not owned by a manual recording.
+    let watching: Bool
+    /// `WatchLoop.State` raw value (`idle`/`watching`/`recording`/`error`), or
+    /// nil when no loop exists.
+    let state: String?
+    /// `BadgeKind` raw value — what the menu bar icon is showing right now.
+    let badge: String
+    /// Whether an app-picker recording owns the loop. Watch control is refused
+    /// while this is true.
+    let manualRecording: Bool
+    /// App name awaiting a browser-meeting consent answer, if one is parked.
+    let pendingConsentApp: String?
+    /// False only when a permission probe has actually failed. A health check
+    /// that has not run yet reports true, matching how `BadgeKind.compute`
+    /// treats an unknown result — the badge and this field must not disagree.
+    let permissionsHealthy: Bool
+
+    /// Fallback for the window between server start and `AppState` being
+    /// reachable. Reports "not watching, nothing wrong" rather than failing the
+    /// request, so a polling client sees a steady off state instead of an error.
+    static let notWatching = Self(
+        watching: false,
+        state: nil,
+        badge: BadgeKind.inactive.rawValue,
+        manualRecording: false,
+        pendingConsentApp: nil,
+        permissionsHealthy: true,
+    )
+}
+
+/// What a `POST /v1/watch` asks for.
+///
+/// `start`/`stop` exist alongside `toggle` because a key press expresses a
+/// desired outcome, not a delta. A remote controller's view of the app is
+/// always slightly stale; if a meeting ended since its last poll, a blind
+/// toggle does exactly the wrong thing and stays inverted until someone
+/// notices. Idempotent verbs converge instead.
+enum WatchAction: String, Codable {
+    case start
+    case stop
+    case toggle
+}
+
+/// What a watch-control request actually achieved. Maps 1:1 onto the HTTP
+/// status code, so the caller can tell "already on" from "refused" without
+/// diffing state before and after.
+///
+/// Deliberately not `Codable` and not raw-value backed: it never crosses the
+/// wire. The route reads it to pick a status code and builds the body from
+/// `watchStatusDTO()`. Publishing a wire conformance on a type that has no wire
+/// presence would invite a later change to serialise it, quietly promoting
+/// these case names into the compatibility promise this file exists to keep.
+enum WatchControlOutcome: Equatable {
+    /// The request changed the watching state. → 200
+    case changed
+    /// Already in the requested state; nothing to do. → 200
+    case unchanged
+    /// Refused: a manual recording owns the loop. → 409
+    case blocked
+    /// The request was accepted but the state did not settle as asked. → 503
+    case failed
+}
+
+/// Request body for `POST /v1/watch`. An unrecognised action string fails to
+/// decode, which the route turns into a 400 — better than silently treating a
+/// typo'd verb as a toggle.
+struct WatchActionPayload: Codable, Equatable {
+    let action: WatchAction
+}
--- a/app/MeetingTranscriber/Sources/WatchingController.swift
+++ b/app/MeetingTranscriber/Sources/WatchingController.swift
@@ -35,6 +35,15 @@ final class WatchingController {
     /// finishes.
     private var startTask: Task<Void, Never>?
 
+    /// See `defaultStartJoinTimeout`.
+    private let startJoinTimeout: Duration
+
+    /// Set while `startManualRecording` is in flight, i.e. from the call until
+    /// `watchLoop` holds the manual loop. `startTask`'s counterpart for the
+    /// manual path, so `isManualRecording` covers the window before the loop
+    /// exists.
+    private var manualStartTask: Task<Void, Never>?
+
     private let settings: AppSettings
     private let notifier: any AppNotifying
     private let pipeline: PipelineController
@@ -116,6 +125,7 @@ final class WatchingController {
                 Permissions.ensureAccessibilityAccess()
             #endif
         },
+        startJoinTimeout: Duration = WatchingController.defaultStartJoinTimeout,
         makeDetector: (() -> any MeetingDetecting)? = nil,
     ) {
         self.settings = settings
@@ -127,6 +137,7 @@ final class WatchingController {
         self.ensureMicAccess = ensureMicAccess
         self.requestScreenRecording = requestScreenRecording
         self.requestAccessibility = requestAccessibility
+        self.startJoinTimeout = startJoinTimeout
         // Tests inject a deterministic detector; production defaults to one
         // filtered by the "Apps to Watch" toggles, re-read at each watch start.
         self.makeDetector = makeDetector ?? { [settings] in
@@ -177,6 +188,20 @@ final class WatchingController {
         watchLoop?.state == .recording
     }
 
+    /// Whether a manual (app-picker) recording owns the loop, or is on its way
+    /// to owning it. Exposed because `toggleWatching` silently refuses in that
+    /// state, and a silent refusal is invisible to a remote caller — the
+    /// automation API turns this into a 409.
+    ///
+    /// The in-flight half matters: `startManualRecording` assigns `watchLoop`
+    /// only after awaiting the mic gate, so reading the loop alone reports
+    /// "no manual recording" for the whole of that window, and a remote start
+    /// landing in it would build an auto loop that the manual task then
+    /// overwrites without stopping.
+    var isManualRecording: Bool {
+        manualStartTask != nil || watchLoop?.isManualRecording == true
+    }
+
     // MARK: - Start / Stop
 
     /// - Parameter userInitiated: True for a deliberate Start/Stop press, which
@@ -185,11 +210,18 @@ final class WatchingController {
     ///   system alert — including on the e2e runners, which force auto-watch on
     ///   and where a stray dialog can swallow the keystroke a lane sends.
     func toggleWatching(userInitiated: Bool = true) {
-        if let loop = watchLoop, loop.isManualRecording { return }
         if let loop = watchLoop, loop.isActive {
+            // Refuse only for a *manual* loop. Stopping a live auto loop is safe
+            // even while a manual start is registered, and the order matters:
+            // between registration and the manual task stopping that loop
+            // itself, a blanket refusal would turn a stop into an instant
+            // `.failed` — a 503 the docs describe as "did not settle within 20
+            // seconds" — and silently no-op the menu-bar toggle.
+            guard !loop.isManualRecording else { return }
             loop.stop()
             watchLoop = nil
         } else {
+            guard !isManualRecording else { return }
             // The start is async — mic access is awaited before `watchLoop` is
             // assigned — so a second toggle in that window would otherwise launch
             // a duplicate WatchLoop and rebuild the queue twice. Ignore it while
@@ -198,6 +230,16 @@ final class WatchingController {
             startTask = Task { @MainActor in
                 defer { startTask = nil }
                 await requestStartPermissions(userInitiated: userInitiated)
+
+                // A manual start can register while this task is parked on the
+                // permission gate, and nothing below re-checks. Whichever side
+                // assigned `watchLoop` last would win and orphan the other loop
+                // — still running, with no reference left to stop it. Bailing
+                // here rather than at the assignment also keeps `pipeline.rebuild()`
+                // from replacing the queue the manual loop is already wired to.
+                // Nothing leaks: `loop.start()` has not run, so there is no
+                // self-retaining watch task yet.
+                guard !isManualRecording else { return }
 
                 syncEngines?()
                 pipeline.rebuild()
@@ -248,62 +290,174 @@ final class WatchingController {
         }
     }
 
+    // MARK: - Idempotent control (automation API)
+
+    /// How long a control call waits for an in-flight start before giving up.
+    /// Generous against the ~2 s warm start, short enough that a wedged request
+    /// still answers well inside a polling controller's patience. Injectable so
+    /// a test can pin the give-up path without waiting it out.
+    static let defaultStartJoinTimeout: Duration = .seconds(20)
+
+    /// Join an in-flight start, giving up after `startJoinTimeout`. True when
+    /// the start settled, false on expiry.
+    ///
+    /// The bound is the point. On a fresh install `toggleWatching` awaits
+    /// `AVCaptureDevice.requestAccess`, which does not return until the dialog
+    /// is answered — and on an `LSUIElement` app that dialog can sit unnoticed
+    /// behind other windows. An unbounded join would hold the HTTP handler and
+    /// its connection open for as long as the prompt is up, and because every
+    /// verb joins the same task, each later POST would park and leak another
+    /// connection.
+    ///
+    /// Polls `startTask` rather than racing `await startTask?.value` in a task
+    /// group. Two things rule the group out: `Task.value` ignores cancellation,
+    /// so the losing child keeps waiting, and a group awaits every child before
+    /// it returns — the join would outlive its own timeout. Giving up must also
+    /// not cancel the start itself, which is a user action already in progress.
+    private func joinStart() async -> Bool {
+        let deadline = ContinuousClock.now + startJoinTimeout
+        while startTask != nil {
+            if Task.isCancelled { return false }
+            if ContinuousClock.now >= deadline { return false }
+            try? await Task.sleep(for: .milliseconds(50))
+        }
+        return true
+    }
+
+    /// Idempotent start. Awaits the in-flight async start — mic gate, engine
+    /// sync, queue rebuild, loop construction — so a caller that reports state
+    /// back observes the settled result instead of a snapshot taken mid-launch.
+    /// `toggleWatching` alone cannot offer that: it returns the moment the task
+    /// is spawned, and `startTask` is private.
+    ///
+    /// Starts as not user-initiated. Nobody is at the machine when this arrives
+    /// — that is the whole point of the endpoint — so the optional Accessibility
+    /// prompt must not be raised, exactly as for auto-watch.
+    @discardableResult
+    func startWatching() async -> WatchControlOutcome {
+        if isManualRecording { return .blocked }
+        if isWatching { return .unchanged }
+        toggleWatching(userInitiated: false)
+        // `toggleWatching` assigns `startTask` synchronously, so this always
+        // joins the task it just created — or, if another caller already had a
+        // start in flight, the one whose existence made our call a no-op.
+        guard await joinStart() else { return .failed }
+        // Re-check: a manual start can register while the auto start is parked,
+        // which makes that start bail and leaves `watchLoop` nil. The join
+        // settled and the dialog was answered, so this is the conflict 409
+        // describes, not the "did not settle" 503.
+        if isManualRecording { return .blocked }
+        return isWatching ? .changed : .failed
+    }
+
+    /// Idempotent stop. Awaits a pending start first: otherwise a stop issued
+    /// during the mic-access window lands before the loop exists, reports
+    /// "already stopped", and is then overwritten by the start completing.
+    ///
+    /// A manual recording is reported as `.unchanged` rather than `.blocked`:
+    /// `isWatching` is false by definition while one owns the loop, so the
+    /// requested end state already holds and there is nothing to refuse. That
+    /// is the asymmetry with `startWatching`, where `.blocked` is right because
+    /// starting would clobber the recording.
+    @discardableResult
+    func stopWatching() async -> WatchControlOutcome {
+        guard await joinStart() else { return .failed }
+        guard isWatching else { return .unchanged }
+        toggleWatching()
+        return isWatching ? .failed : .changed
+    }
+
+    /// Apply a watch action, resolving `.toggle` against settled state.
+    ///
+    /// The leading await matters for `.toggle`: deciding against a mid-launch
+    /// snapshot would read "not watching" for a loop that is seconds from
+    /// running, and start a second one. Settling first means a toggle racing a
+    /// start converges to on — the caller asked for a flip and gets a definite
+    /// answer, rather than a coin toss on where the mic prompt happened to be.
+    @discardableResult
+    func applyWatchAction(_ action: WatchAction) async -> WatchControlOutcome {
+        guard await joinStart() else { return .failed }
+        switch action {
+        case .start: return await startWatching()
+        case .stop: return await stopWatching()
+        case .toggle: return isWatching ? await stopWatching() : await startWatching()
+        }
+    }
+
     func startManualRecording(pid: pid_t, appName: String, title: String) {
+        // `toggleWatching`'s counterpart. The correctness added here rests on
+        // `manualStartTask` being accurate, and without this a second start
+        // replaces the field while the first is still in flight, whose `defer`
+        // then clears the second's registration and reopens the window.
+        guard manualStartTask == nil else { return }
+        manualStartTask = Task { @MainActor in
+            defer { manualStartTask = nil }
+            await performManualRecording(pid: pid, appName: appName, title: title)
+        }
+    }
+
+    /// Body of `startManualRecording`, split out so the task closure stays a
+    /// two-liner and the work is readable on its own.
+    private func performManualRecording(pid: pid_t, appName: String, title: String) async {
+        // Settle an auto start before reading `watchLoop`. Mid-flight it is
+        // still nil, so the stop below would find nothing and the auto task
+        // would later assign over the manual loop, orphaning a live recording
+        // with no way to stop it.
+        _ = await joinStart()
+
         // Stop auto-watch if active
         if let loop = watchLoop, loop.isActive, !loop.isManualRecording {
             loop.stop()
             watchLoop = nil
         }
 
-        Task { @MainActor in
-            _ = await ensureMicAccess()
+        _ = await ensureMicAccess()
 
-            pipeline.ensureQueue()
+        pipeline.ensureQueue()
 
-            // Manual recording never polls the detector, so WatchLoop's default
-            // (unfiltered) detector here is inert, and so is the consent gate
-            // that hangs off it: no detection means no prompt, hence no deny
-            // list to consult, hence the default in-memory store. If this path
-            // ever gains auto-detection, route it through `makeDetector()` like
-            // toggleWatching so the "Apps to Watch" filter still applies, and
-            // pass `denyListStore:` so a "Never for this app" is honoured here
-            // too.
-            let loop = WatchLoop(
-                recorderFactory: makeRecorderFactory(),
-                pipelineQueue: pipeline.queue,
-                pollInterval: settings.pollInterval,
-                noMic: settings.noMic,
-                micDeviceUID: settings.micDeviceUID.isEmpty ? nil : settings.micDeviceUID,
-                verboseDiagnostics: { [settings] in settings.verboseDiagnostics },
-                recordOnly: { [settings] in settings.recordOnly },
-                recordOnlyDestination: { [settings] in
-                    .production(parent: settings.effectiveOutputDir)
-                },
-                notifier: notifier,
+        // Manual recording never polls the detector, so WatchLoop's default
+        // (unfiltered) detector here is inert, and so is the consent gate
+        // that hangs off it: no detection means no prompt, hence no deny
+        // list to consult, hence the default in-memory store. If this path
+        // ever gains auto-detection, route it through `makeDetector()` like
+        // toggleWatching so the "Apps to Watch" filter still applies, and
+        // pass `denyListStore:` so a "Never for this app" is honoured here
+        // too.
+        let loop = WatchLoop(
+            recorderFactory: makeRecorderFactory(),
+            pipelineQueue: pipeline.queue,
+            pollInterval: settings.pollInterval,
+            noMic: settings.noMic,
+            micDeviceUID: settings.micDeviceUID.isEmpty ? nil : settings.micDeviceUID,
+            verboseDiagnostics: { [settings] in settings.verboseDiagnostics },
+            recordOnly: { [settings] in settings.recordOnly },
+            recordOnlyDestination: { [settings] in
+                .production(parent: settings.effectiveOutputDir)
+            },
+            notifier: notifier,
+        )
+        watchLoop = loop
+
+        // Wire channel-health monitoring + error notification on state
+        // transitions — same hook the auto-detect path installs, so the
+        // red-tint indicator and asymmetric-silence notification work
+        // for manual recordings too.
+        attachStateChangeHandler(to: loop, notifyOnRecording: false)
+
+        // Use cached health check result instead of live probe
+        if let health = permissions.health {
+            loop.permissionChecker = { health }
+        }
+
+        do {
+            try await loop.startManualRecording(pid: pid, appName: appName, title: title)
+            notifier.notify(
+                title: "Manual Recording",
+                body: "Recording: \(title)",
             )
-            watchLoop = loop
-
-            // Wire channel-health monitoring + error notification on state
-            // transitions — same hook the auto-detect path installs, so the
-            // red-tint indicator and asymmetric-silence notification work
-            // for manual recordings too.
-            attachStateChangeHandler(to: loop, notifyOnRecording: false)
-
-            // Use cached health check result instead of live probe
-            if let health = permissions.health {
-                loop.permissionChecker = { health }
-            }
-
-            do {
-                try await loop.startManualRecording(pid: pid, appName: appName, title: title)
-                notifier.notify(
-                    title: "Manual Recording",
-                    body: "Recording: \(title)",
-                )
-            } catch {
-                notifier.notify(title: "Error", body: error.localizedDescription)
-                watchLoop = nil
-            }
+        } catch {
+            notifier.notify(title: "Error", body: error.localizedDescription)
+            watchLoop = nil
         }
     }
 
--- a/app/MeetingTranscriber/Tests/DebugRPCServerIntegrationTests.swift
+++ b/app/MeetingTranscriber/Tests/DebugRPCServerIntegrationTests.swift
@@ -43,6 +43,8 @@
             confirmNaming: @escaping (UUID, [String: String]) -> Bool = { _, _ in false },
             skipJobNaming: @escaping (UUID) -> Bool = { _ in false },
             transcribe: @escaping (URL, Double) async -> BlockingTranscribeResult = { _, _ in .noFile },
+            watchStatus: @escaping () -> WatchStatusDTO = { .notWatching },
+            watchControl: @escaping (WatchAction) async -> WatchControlOutcome = { _ in .failed },
         ) async throws -> URL {
             let server = DebugRPCServer(
                 port: 0,
@@ -57,6 +59,8 @@
                 confirmNaming: confirmNaming,
                 skipJobNaming: skipJobNaming,
                 transcribe: transcribe,
+                watchStatus: watchStatus,
+                watchControl: watchControl,
             )
             self.server = server
             server.start()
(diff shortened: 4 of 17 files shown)


app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,6 +193,54 @@ final class WatchingController {
...
+    /// is spawned, and `startTask` is private.
+    @discardableResult
+    func startWatching() async -> WatchControlOutcome {
+        if isManualRecording { return .blocked }
+        if isWatching { return .unchanged }
+        toggleWatching()
+        // `toggleWatching` assigns `startTask` synchronously, so this always
+        // joins the task it just created — or, if another caller already had a
+        // start in flight, the one whose existence made our call a no-op.
+        await startTask?.value
```
pasrom: **`start` can park here indefinitely, and no follow-up request can break it out.**

On a fresh install with mic status `.notDetermined`, `toggleWatching()` reaches `_ = await ensureMicAccess()` (line 160), which in production is `Permissions.ensureMicrophoneAccess()` and therefore `await AVCaptureDevice.requestAccess(for: .audio)`. That does not return until the dialog is answered. This `await startTask?.value` joins it, so the HTTP handler never produces a response, and `receive(connection:accumulated:)` puts no deadline on `await self.route(request)`: the `NWConnection` and its handler task stay alive for as long as the prompt is open.

The obvious remedy does not apply, because `stopWatching()` (line 220) and `applyWatchAction` (line 236) both open with the same await. Every subsequent `POST /v1/watch`, whatever the verb, parks on the same task and leaks another connection. `GET /v1/watch` stays serviceable, since each connection gets its own task, which makes the wedge harder to notice rather than easier.

The launch health check does not defuse it either: `PermissionHealthCheck.checkMicrophoneLive` skips the probe for any status other than `.authorized`, so a fresh install genuinely first prompts on the first watch start. On an `LSUIElement` app that dialog can sit behind other windows unnoticed.

Worth being precise about what is new: pre-PR the same task already awaited the prompt, but nothing joined it in a response path. The blocking HTTP contract is this PR's addition. Bounding the join with a timeout that returns `.failed` would fix it and would incidentally make the documented `503` reachable.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,6 +193,54 @@ final class WatchingController {
...
+    // MARK: - Idempotent control (automation API)
+
+    /// Idempotent start. Awaits the in-flight async start — mic gate, engine
+    /// sync, queue rebuild, loop construction — so a caller that reports state
+    /// back observes the settled result instead of a snapshot taken mid-launch.
+    /// `toggleWatching` alone cannot offer that: it returns the moment the task
+    /// is spawned, and `startTask` is private.
+    @discardableResult
+    func startWatching() async -> WatchControlOutcome {
+        if isManualRecording { return .blocked }
```
pasrom, line 338: **This guard reads clear during an in-flight manual start, so the `409` contract does not hold in that window.**

`startManualRecording` assigns `watchLoop` only after `await ensureMicAccess()` inside a bare `Task`, and has no `startTask` equivalent (the stored `startTask` is set only in `toggleWatching`). While that task is parked on the mic gate, `watchLoop` is nil: `isManualRecording` is false here, `isWatching` is false, and `guard startTask == nil` passes. An auto `WatchLoop` gets built, assigned and started. The manual task then resumes and overwrites `watchLoop`, dropping the auto loop without `stop()`.

The orphan does not wind down. `WatchLoop.start()` spawns `Task { [weak self] in guard let self else { return }; await self.watchLoop() }`; the `guard let self` holds self strongly for the task's duration, self holds `watchTask`, and the loop exits only on a cancellation that nothing can now trigger. Its recorder factory is still functional, so a detected meeting can start a second concurrent recording.

Both interleavings are bad. If the manual task resumes first, the auto task's later `watchLoop = loop` orphans the *manual* loop mid-recording instead, and `stopManualRecording` then targets the wrong one.

On scope, to be fair: the race is not this PR's doing. Every line involved is context, and the same orphaning was already reachable by clicking menu-bar Start Watching during an in-flight picker start. What this PR adds is a remote caller that can land in that window at any moment, plus the documented promise that watch control is refused while a manual recording owns the loop. Giving `startManualRecording` the same `startTask` treatment would close it.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,6 +193,54 @@ final class WatchingController {
...
+        return isWatching ? .changed : .failed
+    }
+
+    /// Idempotent stop. Awaits a pending start first: otherwise a stop issued
+    /// during the mic-access window lands before the loop exists, reports
+    /// "already stopped", and is then overwritten by the start completing.
+    @discardableResult
+    func stopWatching() async -> WatchControlOutcome {
+        await startTask?.value
+        if isManualRecording { return .blocked }
```
pasrom: **`stop` returns `409` even though the requested end state already holds.**

`isWatching` is `watchLoop?.isActive == true && watchLoop?.isManualRecording == false`, so during a manual recording it is false by definition. The caller asked for "not watching" and that is already true, but this `.blocked` check sits ahead of the `guard isWatching` on the next line, so the route emits `409`.

That sits awkwardly with the rest of the PR, which argues that a key press expresses a desired outcome and maps already-being-there to `.unchanged` and `200`. The route comment justifying the `409` reasons only about the case that applies to `start` and `toggle`, a silent refusal that looks like success. For `stop` there is nothing to refuse.

Not clear-cut, and there is a real counter-argument: a controller may well want to be told that a recording owns the loop rather than getting a bland `200`. If that is the intent, it is worth saying so here, because as written the check reads as inherited by symmetry with `startWatching`, where `.blocked` is unambiguously right because starting would clobber the recording.

app/MeetingTranscriber/Sources/AppState+RPC.swift
```diff
@@ -142,6 +142,71 @@
...
+        /// Kept separate from `rpcStateSnapshot` on purpose: that snapshot is the
+        /// debug surface and is free to change shape, while this one is polled on
+        /// an interval by third-party controllers and carries a compatibility
+        /// promise. Locals are bound first (rather than reading the two-hop
+        /// `watching.watchLoop.…` `@Observable` chain inline) to keep the
+        /// memberwise init inside the type-check budget — see `rpcQueueStatus`.
+        func watchStatusDTO() -> WatchStatusDTO {
+            let loop = watching.watchLoop
+            let badge = currentBadge.rawValue
+            // `!= false` rather than `?? true`: nil means the probe has not run,
```
pasrom: **This comment contrasts two spellings that are exactly equivalent.**

`health` is `HealthCheckResult?` and `isHealthy` is a non-optional `Bool`, so `permissions.health?.isHealthy` is `Bool?`, and the truth tables match in all three cases: `nil != false` and `nil ?? true` are both true, `.some(true)` is true either way, `.some(false)` is false either way.

The behaviour the comment describes (nil means the probe has not run, which is not a permission problem) is right, and `?? true` implements exactly that. The "rather than" framing tells a future reader there is a subtle reason not to simplify, and there is not. The contrast would only carry weight against `== true` or `?? false`, neither of which the comment names.

app/MeetingTranscriber/Sources/WatchStatusDTO.swift
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/// What a `POST /v1/watch` asks for.
+///
+/// `start`/`stop` exist alongside `toggle` because a key press expresses a
+/// desired outcome, not a delta. A remote controller's view of the app is
+/// always slightly stale; if a meeting ended since its last poll, a blind
+/// toggle does exactly the wrong thing and stays inverted until someone
+/// notices. Idempotent verbs converge instead.
+enum WatchAction: String, Codable, CaseIterable {
```
pasrom: **`CaseIterable` has no reader.**

Grepping the app and `tools/mt-cli` at this head finds no `allCases` use of `WatchAction`; every reference is a decode site, a switch or a test. `mt-cli` declares its own separate `Watch.Action`, where `CaseIterable` genuinely is load-bearing for ArgumentParser, which may be where this came from.

`Codable` on this enum is used (the `WatchActionPayload` decode), so only the `CaseIterable` is dead.

app/MeetingTranscriber/Sources/WatchStatusDTO.swift
```diff
@@ -0,0 +1,82 @@
...
+    case start
+    case stop
+    case toggle
+}
+
+/// What a watch-control request actually achieved. Maps 1:1 onto the HTTP
+/// status code, so the caller can tell "already on" from "refused" without
+/// diffing state before and after.
+enum WatchControlOutcome: String, Codable, Equatable {
```
pasrom: **The `String` raw type and `Codable` are both unused on this one.**

Every reference is a closure signature or a bare-case switch. The sole consumption is the status-code mapping in the route, and the response body is built entirely from `watchStatusDTO()`. This outcome is never encoded, never decoded, and `.rawValue` is never read. (`Equatable` is used by the tests and would be synthesized regardless.)

Worth dropping given this file's own framing: the type carries a compatibility promise, and publishing `Codable` on something that never crosses the wire invites a later change to start serialising it, quietly turning these case names into a wire contract.

tools/mt-cli/Sources/MTCLI.swift
```diff
@@ -11,11 +11,48 @@ struct MTCLI: AsyncParsableCommand {
...
+    )
+
+    @Argument(help: "status (default), start, stop, or toggle.")
+    var action: Action = .status
+
+    func run() async throws {
+        let client = try RPCClient.loadDefault()
+        let data = action == .status
+            ? try await client.get("/v1/watch")
+            : try await client.post("/v1/watch", json: ["action": action.rawValue])
```
pasrom: **This POST takes the 5 s fast-path default against an endpoint whose whole design is to block until the state settles.**

`post(_:json:timeout:)` defaults to `RPCClient.requestTimeoutSeconds = 5`, whose own comment justifies the value with "the rest of the API is fast". This endpoint breaks that premise deliberately: the route awaits `watchControl`, which awaits `applyWatchAction`, which awaits `startTask` across the mic gate, engine sync, queue rebuild and loop construction. The PR body measures 2.0 s warm, so the headroom is thin even without a first-run TCC prompt.

The failure mode is not a clean one. `RPCClient.request` maps only `.cannotConnectToHost`, so `.timedOut` surfaces as a bare "The request timed out" with a non-zero exit. Meanwhile the start completes server-side anyway, because `toggleWatching` spawns an unstructured task that a client disconnect does not cancel. The script reports failure while watching is on, which is the one outcome an idempotent verb was supposed to rule out.

`/screenshot` was given `screenshotTimeoutSeconds = 15` for exactly this reason. This call wants its own generous timeout too.

tools/mt-cli/Sources/MTCLI.swift
```diff
@@ -11,11 +11,48 @@ struct MTCLI: AsyncParsableCommand {
...
+        FileHandle.standardOutput.write(data)
+        FileHandle.standardOutput.write(Data("\n".utf8))
```
pasrom, line 55: **These four lines are `postAction`'s body.**

`postAction(_:_:)` further down does load client, POST JSON, write body, write newline, and its doc comment says it is shared by every action subcommand that returns the server's outcome JSON unchanged. Its `[String: String]` signature fits `["action": action.rawValue]`.

Minor, and to be fair about it: the ternary above means these lines serve the GET branch too, there is no equivalent GET helper, and `State.run()`, `UITree` and `ConfirmBrowserConsent` already inline the same sequence. So this follows the file's existing style rather than creating the split. Routing only the POST branch through the helper saves one call site at the cost of an asymmetry, which may not be worth it.

app/MeetingTranscriber/Tests/DebugRPCServerIntegrationTests.swift
```diff
@@ -1089,6 +1093,119 @@
             )
         }
 
+        // MARK: - /v1/watch
```
pasrom, line 1096: **Nothing exercises the production wiring behind this route.**

All seven route tests below go through `startServer(watchStatus:watchControl:)` with literal stubs, and the controller tests call `startWatching`, `stopWatching` and `applyWatchAction` directly. `watchRPCClosures` appears only at its definition and its single call site in `AppState`; no test references it, and `watchStatusDTO()` is never invoked from a test.

That gap is load-bearing, because `DebugRPCServer.init` declares defaults (`{ .notWatching }` and `{ _ in .failed }`). Deleting the two wiring arguments at the `AppState` call site still compiles, and every test in this file stays green.

`AppStateTests.testBuildDebugRPCServerServesV1OverSocket` already drives the real wiring over a socket for jobs, naming and transcribe. It constructs these closures but never requests `/v1/watch`, so it does not cover this. Extending that test, or adding a watch assertion to `scripts/test_rpc.sh`, would close it cheaply.

app/MeetingTranscriber/Tests/WatchingControllerTests.swift
```diff
@@ -197,4 +197,122 @@ final class WatchingControllerTests: XCTestCase {
...
+        addTeardownBlock { await controller.watchLoop?.stop() }
+        await controller.startWatching()
+
+        let outcome = await controller.startWatching()
+
+        XCTAssertEqual(outcome, .unchanged)
+        XCTAssertTrue(controller.isWatching, "an idempotent start must not toggle watching off")
+    }
+
+    func testStopWatchingStopsAndIsThenUnchanged() async {
```
pasrom, line 440: **The only new test with no cleanup at all.**

It starts a real loop on the next line and relies on `stopWatching()` actually stopping, which is precisely the behaviour under test. If that regresses, the loop survives the test and keeps polling for the rest of the subprocess, on top of the assertion failure.

Sibling tests use `addTeardownBlock { await controller.watchLoop?.stop() }`, and `testControlIsBlockedDuringManualRecording` uses a `defer`. Either would do here. (The leaked loop polls the silent stub detector from `makeController`, not real IOKit, so the blast radius is a stray task rather than real device polling.)

app/MeetingTranscriber/Tests/WatchingControllerTests.swift
```diff
@@ -197,4 +197,122 @@ final class WatchingControllerTests: XCTestCase {
...
+        // Not trailing-closure: a trailing closure binds to `makeDetector`.
+        // swiftlint:disable:next trailing_closure
+        let controller = makeController(ensureMicAccess: {
+            await micGate.wait()
+            return true
+        })
+        addTeardownBlock { await controller.watchLoop?.stop() }
+
+        controller.toggleWatching() // start, parked on the mic gate
+        async let stopping = controller.stopWatching()
```
pasrom, line 473: **Nothing orders the stop's arrival, so the guarantee this test claims is probabilistic.**

First, what does work: with `await startTask?.value` deleted from `stopWatching`, this test normally fails. The stop's main-actor job is enqueued at the `async let`, ahead of the start task's post-gate continuation, so it reads `watchLoop == nil`, returns `.unchanged`, and both assertions fall over.

But that outcome rests on the child task's start latency on the global pool. If it is scheduled late enough to land past both the gate round-trip and the start's entire post-gate continuation (lines 161 to 191 run as one main-actor job with no await in them), the revert passes instead. Under `--parallel` that window is real.

So the `AsyncGate` doc comment below overclaims with "deterministic instead of timing-dependent": the gate controls when the start unparks, not when the stop arrives, and only the latter decides the outcome. A `waitFor(...)` on an observable "start has parked" flag before the `async let`, the idiom already used elsewhere in this file, would make the ordering real.

app/MeetingTranscriber/Tests/WatchingControllerTests.swift
```diff
@@ -197,4 +197,122 @@ final class WatchingControllerTests: XCTestCase {
...
+        XCTAssertEqual(started, .blocked)
+        XCTAssertEqual(stopped, .blocked)
+        XCTAssertEqual(controller.watchLoop?.isManualRecording, true, "control must not disturb a manual recording")
+    }
+}
+
+/// One-shot async gate: lets a test park an injected `async` seam and release it
+/// on cue, so a race (stop arriving mid-start) is deterministic instead of
+/// timing-dependent.
+private actor AsyncGate {
```
pasrom: **Fifth copy of this primitive in the test suite.**

The same continuation-parking gate already exists as `FinishGate` in `LiveTranscriptionEnglishStreamingTests`, and as inline continuation fields in `LiveCaptionFeedTests`, `ParkedEngine`, `LiveTranscriptionCoordinatorTests`, and (looking a little further) `SingleFlightTests`, `SpeakerNamingSessionTests` and `WatchLoopBrowserConsentTests`. `Tests/TestHelpers.swift` is the shared home for cross-file test utilities and already hosts `waitFor`.

Not a blocker, and an honest caveat: most existing copies are continuation fields embedded in mock actors with slightly different resume semantics, so consolidating them is a small refactor per site rather than a mechanical swap. This standalone actor is the cleanest version of the pattern in the tree, which makes it the natural one to promote.

docs/automation-api.md
```diff
@@ -202,6 +204,57 @@ Responses:
...
+
+Responses:
+
+- `200 OK` — the request was satisfied. This covers *both* "the state changed"
+  and "it was already like that": asking an already-watching app to `start` is a
+  satisfied request, not an error.
+- `409 Conflict` — refused because a manual (app-picker) recording owns the
+  watch loop. Stop that recording first. The body is still a `WatchStatusDTO`,
+  with `manualRecording: true`, so a client can explain the refusal.
+- `503 Service Unavailable` — the action was attempted but the state did not
```
pasrom, line 257: **This outcome is not reachable.**

`.failed` maps to `503`, but the only path producing `.failed` for a start is `guard let self else { return .failed }` in `watchRPCClosures`, meaning a deallocated `AppState`. A denied microphone never gets there: `toggleWatching` deliberately discards the gate result (`_ = await ensureMicAccess()`) and builds and starts the loop regardless, `WatchLoop.start()` sets the phase synchronously, so `isWatching` is true and `startWatching` returns `.changed`, which is `200`.

`permissionsHealthy` does not catch it either. It reads the cached snapshot, and on a fresh install that snapshot records mic `.notDetermined`, which `HealthCheckResult.problems` does not count as a problem, so the field is true.

Net effect: the user denies the prompt, the response says `watching: true, permissionsHealthy: true`, a Stream Deck key lights green, and the mic channel is silent for every meeting that follows. (App audio still records, since it hangs off a separate grant, so it is a silent local channel rather than nothing at all.)

Either make the `503` real by acting on the gate result, or drop the sentence. As written it documents a safety net that is not there.

docs/stream-deck.md
```diff
@@ -0,0 +1,171 @@
...
+toggle off and back on.
+
+## A Stream Deck plugin, if you want live state
+
+[**Meeting Transcriber for Stream Deck**](https://github.com/alexpfau/meeting-transcriber-streamdeck)
+is a plugin that polls `/v1/watch` and repaints the key, so it shows whether
+watching is on even when it was turned on from the app. Three actions — toggle,
+start, stop.
+
+[Download the latest `.streamDeckPlugin`](https://github.com/alexpfau/meeting-transcriber-streamdeck/releases/latest)
```
pasrom, line 32: **The download instruction needs one sentence on what the token grants.**

Upfront: I know the plugin is yours, and this is not an objection to it, nor to linking it here. The point is narrower than it may read.

The token is not scoped to watching. It opens the whole server: all of `/v1`, plus the `/action/*` surface and `/screenshot`. The token discussion earlier in this file covers mode `0600`, the loopback binding and rotation, but never says what the token actually authorises.

"It reads the token file itself, so there is no pairing step and nothing to configure" is written purely as a convenience, and it is also the thing a reader should weigh: a doc committed to this repo is telling them to download and double-click a binary that then helps itself to that token. The provenance note a few lines down is good and covers the part a reader could not otherwise work out.

One sentence on what the token grants, so a reader can decide for themselves whether to hand it over, would settle this.

docs/stream-deck.md
```diff
@@ -0,0 +1,171 @@
...
+step and nothing to configure.
+
+It is a separate project rather than part of this repo: it ships on Elgato's
+release cadence, not this app's, and nobody without a Stream Deck should have to
+download it.
+
+If you would rather not install a plugin, the rest of this guide does the same
+job with a Shortcut — minus the live state.
+
+## The simplest thing that works
```
pasrom, line 51: **The default walkthrough builds the binding the doc later argues against.**

This section walks the reader through a Shortcut named "Toggle Meeting Watching" posting `{"action":"toggle"}`, and step 2 hard-codes that name into the Stream Deck URL. "Start and stop beat toggle" further down then calls `toggle` "the obvious binding and the worst one" and explains how it desynchronises.

Softer than it first looks, since step 1 already links forward to that section and the later section explicitly sanctions the one-key case. Still, the tutorial's default recipe, and the Shortcut name a reader will have baked into their key by the end of it, is the binding that the rest of this doc, the CLI help and the `WatchAction` doc comment all argue against. Building `start` and `stop` keys here, with toggle noted as the one-key fallback, would line the walkthrough up with the advice.

pasrom (changes requested):
Review of the `/v1/watch` surface. Every point is anchored inline.

The three that would hold up a merge are at the top of `WatchingController.swift` and in `docs/automation-api.md`: a `start` that can park on the microphone TCC prompt with no way to break out of it, a documented `503` that cannot be reached, and the `409` promise not holding in the window where a manual recording is still starting up.

The rest is smaller: one timeout mismatch in the CLI, a coverage gap over the production wiring, three test-hygiene points, and four notes on the two docs.

Each point was checked against the merge base, so nothing here is untouched surrounding code that the diff merely sits next to. Where a mechanism predates this PR but the new endpoint is what makes it reachable or contract-breaking, the comment says so explicitly.

On method: this is a read of the diff at `9110d4e0`, not a run against a build. So every point cites the code it rests on and is meant to be checked rather than taken on trust. Tooling assisted; the judgement calls and the priorities are mine.

alexpfau:
All fifteen addressed in `933c332`, separate commit as before.

The three blockers were right, and the first one was worse than the review could see from a diff read — details below, because chasing it turned up a deadlock in my own fix.

## The three blockers

**`start` parking indefinitely.** `joinStart` now gives up after 20 s and returns `.failed`, which is also what makes the documented 503 reachable.

My first attempt raced `await startTask?.value` against a `Task.sleep` in a task group. That deadlocks, and I only found it because the mutation run hung instead of failing: `Task.value` ignores cancellation, so `cancelAll()` does not release the losing child, and a task group awaits every child before it returns. The join would therefore outlive its own timeout — exactly the bug it was added to fix, just moved. It now polls `startTask` against a deadline instead. Polling is the less elegant shape, and it is the one that works here: giving up must not cancel the start, which is a user action already in progress, and structured concurrency gives no way to stop awaiting a child without cancelling it. `testStartWatchingGivesUpWhenTheStartNeverSettles` pins both the outcome and the fact that it returns at all.

**The 409 window.** `startManualRecording` now gets `startTask`'s counterpart, `manualStartTask`, and `isManualRecording` covers it. That closes the interleaving you described and, since `toggleWatching` reads the same property, the menu-bar route into it too. The reverse interleaving — manual start landing during an in-flight auto start, orphaning the *manual* loop — is closed by settling the pending start inside the manual task before it reads `watchLoop`.

**The unreachable 503.** Documented as what it now is. I did not act on the mic gate result: `toggleWatching` discarding it is deliberate and app audio records without the grant, so turning a denial into a failed start would be a behaviour change well outside this PR. Instead the doc says plainly that a denial gives you `200` with `watching: true` and points at `permissionsHealthy`, rather than promising a safety net that is not there.

## `stop` during a manual recording

Changed to `.unchanged`/200. Your counter-argument is real — a controller may want to know a recording owns the loop — but the response body already carries `manualRecording: true` on every reply, so that information is not lost by returning 200. Given it is available either way, the semantics decide it: the caller asked for "not watching", and that already holds. 409 stays for `start`, where it is unambiguous.

If you would rather keep the symmetry, say so and I will put it back with the reasoning written down instead.

## The rest

| Note | What changed |
|---|---|
| CLI timeout mismatch | Own 30 s constant, set above the server's 20 s bound so the server always answers first |
| `postAction` helper | **Not** taken — see below |
| No coverage of production wiring | Socket test drives the real closures; deleting the wiring arguments now reddens it |
| Missing teardown | Added |
| Probabilistic ordering | Ordered on `AsyncGate.hasWaiter` before the `async let` |
| Fifth copy of the gate | Promoted to its own file, with the accessor that makes ordering possible |
| `CaseIterable` unused | Dropped |
| Raw type + `Codable` unused | Dropped, with your reasoning about wire conformances recorded in the doc comment |
| Equivalent-spellings comment | Rewritten to describe the behaviour only |
| 503 not reachable | Above |
| Token scope undocumented | One paragraph on what the token opens, before the download instruction |
| Walkthrough builds `toggle` | Now builds start and stop keys; toggle is the one-key fallback |

On `postAction`: I left the POST inlined. Your own caveat is the reason — the ternary serves the GET branch too, there is no GET helper, and three existing commands inline the same sequence. Routing half of it through the helper buys one call site and costs an asymmetry, which reads worse than the repetition.

## Verification

Each fix was checked against the mutation it is meant to catch:

| Mutation | Result |
|---|---|
| Drop the join from `stopWatching` | Race test red |
| Revert `isManualRecording` to loop-only | Manual-window test red (`"failed"` vs `"blocked"`) |
| Delete both wiring arguments at the `AppState` call site | Socket test red at the request count (503, the default stub) |

Worth noting the wiring mutation no longer compiles in its literal form — dropping the arguments leaves `watch` unused, which is an error here. I mutated the binding to `_` as well to get a compiling version.

SwiftFormat **0.61.1** and `swiftlint --strict` clean; the refactor into `performManualRecording` and the move of `AsyncGate` out of `TestHelpers.swift` were both driven by that (`closure_body_length` and `file_length` respectively — `TestHelpers.swift` is at its 600-line limit). 139 tests green across `WatchingControllerTests`, `AppStateTests` and `DebugRPCServerIntegrationTests`.

One unrelated observation: `AppStateTests.testEnqueueFilesUsesSidecarMetadataWhenPresent` failed once for me under the combined filter with `FFmpegHelper.swift:134: *** -[NSConcreteTask terminationStatus]: task still running`, then passed on three consecutive re-runs and in isolation. Looks like a pre-existing flake in that helper rather than anything here, but flagging it since I saw it.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,58 +211,155 @@ final class WatchingController {
...
+    }
+
+    /// Body of `startManualRecording`, split out so the task closure stays a
+    /// two-liner and the work is readable on its own.
+    private func performManualRecording(pid: pid_t, appName: String, title: String) async {
+        // Settle an auto start before reading `watchLoop`. Mid-flight it is
+        // still nil, so the stop below would find nothing and the auto task
+        // would later assign over the manual loop, orphaning a live recording
+        // with no way to stop it.
+        _ = await joinStart()
```
pasrom, line 406: **This result is discarded, and past the 20 s bound that brings the orphan back.**

When the auto start is the thing that is wedged, which is the case the bound exists for, `joinStart` gives up here and the manual path carries on anyway. `watchLoop` is still nil, so the "Stop auto-watch if active" check below finds nothing; the manual path then parks on the same dialog, builds its loop and assigns. The auto task's tail has no re-check of `isManualRecording` anywhere between `await ensureMicAccess()` and `watchLoop = loop; loop.start()`, so whichever side assigns last wins and the other loop is orphaned: still running, with nothing left holding a reference that could stop it.

Both orders are bad, and they are not symmetric. The orphaned *auto* loop is the unbounded one: it keeps polling and starts a second concurrent recording on the next detected meeting, invisible in the UI. The orphaned *manual* loop at least ends on pid exit or `maxDuration`, so that one is a hot mic with a dead Stop rather than lost audio. In that order there is also a follow-on: if `loop.startManualRecording` throws, the catch's `watchLoop = nil` drops the auto loop as well and both are orphaned.

The fix I would suggest is not here but in the auto start task, immediately after its `_ = await ensureMicAccess()`: `guard !isManualRecording else { return }`. That is timing-independent, and both halves of the widened `isManualRecording` cover it. Putting it just before `watchLoop = loop` instead would also stop the orphan, but the tail would still run `pipeline.rebuild()` first, which replaces `pipeline.queue` after the manual path already wired its loop to the old object, leaving the manual recording to enqueue into a queue nothing observes. Returning early there leaks nothing, since `loop.start()` has not run and no self-retaining `watchTask` exists yet.

Acting on the `false` here instead is the tempting alternative, but it turns an unanswered dialog into a dead Record App button, which seems worse than the race.

app/MeetingTranscriber/Sources/AppState+RPC.swift
```diff
@@ -142,6 +142,71 @@
...
+            let badge = currentBadge.rawValue
+            // nil means the probe has not run, which is not a permission
+            // problem. `BadgeKind.compute` treats it the same way, and the
+            // badge and this field must not disagree.
+            let healthy = permissions.health?.isHealthy != false
+            return WatchStatusDTO(
+                watching: watching.isWatching,
+                state: loop?.state.rawValue,
+                badge: badge,
+                manualRecording: isManualRecording,
```
pasrom: **This is `AppState`'s own loop-only copy, so the widened predicate never reaches the body.**

`WatchingController.isManualRecording` is now `manualStartTask != nil || watchLoop?.isManualRecording == true`, but `AppState.isManualRecording` is still `watching.watchLoop?.isManualRecording == true`, and that is what lands here and in `/state`.

So in exactly the window this commit turned into a 409, the 409 body reports `manualRecording: false`. The bullet added to `docs/automation-api.md` in this same commit promises the opposite: "with `manualRecording: true`, so a client can explain the refusal". The window is not a sliver either, it spans the manual task's `joinStart` (up to 20 s behind a parked auto start), the mic gate, and the recorder start.

One-line fix: forward to `watching.isManualRecording`. Worth grepping for other copies of the old expression while you are in there. A grep for the *name* finds this one but it reads like an ordinary caller; only a grep for the body shows it is a second definition.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,58 +211,155 @@ final class WatchingController {
...
+        guard await joinStart() else { return .failed }
+        switch action {
+        case .start: return await startWatching()
+        case .stop: return await stopWatching()
+        case .toggle: return isWatching ? await stopWatching() : await startWatching()
+        }
+    }
+
     func startManualRecording(pid: pid_t, appName: String, title: String) {
+        manualStartTask = Task { @MainActor in
```
pasrom, line 393: **No in-flight guard, and the `defer` nils unconditionally.**

`toggleWatching` has `guard startTask == nil else { return }`; this has no counterpart. Two manual starts in quick succession both run, the second replaces the field, and when the first finishes, its `defer { manualStartTask = nil }` clears the second's registration while that one is still in flight. `isManualRecording` then reads false again and the window this commit closed reopens.

I cannot show a path from the UI, since the picker window closes right after the call, so treat this as hardening rather than a demonstrated bug. The reason I would still do it: the new correctness now rests on this field being accurate. `guard manualStartTask == nil else { return }` closes it.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,58 +211,155 @@ final class WatchingController {
...
+    /// Polls `startTask` rather than racing `await startTask?.value` in a task
+    /// group. Two things rule the group out: `Task.value` ignores cancellation,
+    /// so the losing child keeps waiting, and a group awaits every child before
+    /// it returns — the join would outlive its own timeout. Giving up must also
+    /// not cancel the start itself, which is a user action already in progress.
+    private func joinStart() async -> Bool {
+        let deadline = ContinuousClock.now + startJoinTimeout
+        while startTask != nil {
+            if ContinuousClock.now >= deadline { return false }
+            try? await Task.sleep(for: .milliseconds(10))
```
pasrom: **Degenerates into a hot spin if this task is ever cancelled.**

Once the enclosing task is cancelled, `try? await Task.sleep` throws immediately on every iteration, so the loop spins nil-check, deadline-check, instant-throw on the main actor for the rest of the deadline.

Unreachable today: `DebugRPCServer.receive` runs each request in an unstructured `Task { @MainActor in ... }` that is never stored or cancelled, so a client disconnect does not cancel the handler. Still worth an `if Task.isCancelled { return false }` in the loop, given this method exists specifically to bound pathological waits. It would not cancel the start, which you never do anyway.

Separately, 10 ms is tighter than the problem needs: 100 main-actor wakeups a second per parked caller, against a roughly 2 s normal start and a 20 s bound. 50 to 100 ms would cost nothing observable.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -186,58 +211,155 @@ final class WatchingController {
...
+    /// during the mic-access window lands before the loop exists, reports
+    /// "already stopped", and is then overwritten by the start completing.
+    ///
+    /// A manual recording is reported as `.unchanged` rather than `.blocked`:
+    /// `isWatching` is false by definition while one owns the loop, so the
+    /// requested end state already holds and there is nothing to refuse. That
+    /// is the asymmetry with `startWatching`, where `.blocked` is right because
+    /// starting would clobber the recording.
+    @discardableResult
+    func stopWatching() async -> WatchControlOutcome {
```
pasrom, line 363: **A stop can return an instant 503 in the sliver where a manual start is registered but the auto loop is still running.**

Moving "Stop auto-watch if active" out of the synchronous body of `startManualRecording` and into the spawned task created a state that could not exist before: `manualStartTask != nil` while an auto loop is still active. A `stopWatching` landing there passes `guard isWatching`, calls `toggleWatching()`, which now refuses at `if isManualRecording { return }`, and returns `.failed`, so 503, immediately and with no timeout involved. The docs describe 503 as "did not settle within 20 seconds". The menu-bar toggle in the same sliver silently no-ops where it used to stop the auto loop.

Self-healing (the manual task stops the auto loop moments later, and a retry converges) and the window is about one main-actor hop, so a nit. Worth flagging because it is the one behaviour change in the `performManualRecording` split that the commit message does not mention; the error path and everything else diff clean against the parent.

app/MeetingTranscriber/Tests/WatchingControllerTests.swift
```diff
@@ -197,4 +199,173 @@ final class WatchingControllerTests: XCTestCase {
...
+
+    /// A start parked on an unanswered microphone dialog must not hold the HTTP
+    /// handler and its connection open for as long as the prompt is up. The join
+    /// gives up and reports `.failed`, which the route turns into a 503.
+    ///
+    /// Also pins that giving up actually returns: an earlier revision raced the
+    /// join against a sleep in a task group, which deadlocked, because
+    /// `Task.value` ignores cancellation and a group awaits every child before
+    /// returning.
+    func testStartWatchingGivesUpWhenTheStartNeverSettles() async {
```
pasrom, line 536: **This one hangs rather than fails if the give-up path regresses.**

If the join goes unbounded again, `await controller.startWatching()` never returns, and XCTest has no per-test timeout, so under `--parallel` this method's subprocess wedges until the CI job timeout.

Fine as a tripwire, since a hang is loud and this is exactly the regression you want to notice. Worth knowing it is a hang and not a red assert, though, because that is a slower and less obvious signal than the other mutation checks in this round.

pasrom (changes requested):
Three of these are new; the rest of the round is good.

The 20 s bound and the manual-start window are both closed the way I hoped, and catching the task-group deadlock in your own first attempt is the part that matters. On `stop`: keep `.unchanged`, your point that `manualRecording` rides in every body settles it. I have resolved the fifteen threads from the last round.

Two to fix, plus one that landed under you:

1. `performManualRecording` discards `joinStart()`'s result, so past the 20 s bound the orphaned loop is back. The guard belongs in the auto start task right after its mic gate rather than at the assignment, or `pipeline.rebuild()` still strands the manual loop on a replaced queue. Details inline.
2. The widened `isManualRecording` never reached the DTO. `AppState` keeps its own loop-only copy, so the new 409 answers `manualRecording: false` and contradicts the doc line this same commit added.
3. `main` moved yesterday and the PR is conflicting now. #607 gave `toggleWatching` a `userInitiated` parameter defaulting to `true`, and `startWatching()` calls it bare, so after the rebase a remote `/v1/watch` start would raise the Accessibility dialog at a machine nobody is sitting at. That parameter's own doc comment sets the rule ("the auto-watch path passes false so an unattended launch raises no system alert"), and a Stream Deck press from another room is about as unattended as it gets. That is a decision rather than a mechanical merge, which is why I am not making it for you.

Four nits inline, all optional.

Happy to push the two fixes myself if you would rather not. Your call.

alexpfau:
Rebased onto `53d090b` and took the `userInitiated` change you flagged. Conflicts are resolved, the branch is 7 ahead / 0 behind main, and GitHub now reports it mergeable.

## The cross-PR fix

`0e3e5b4`. `startWatching` now calls `toggleWatching(userInitiated: false)`.

Your reading was right on both halves. `stopWatching`'s call takes the stop branch and never reaches the permission block, so it needed nothing. `applyWatchAction(.toggle)` inherits the fix rather than needing its own, since a toggle resolving to a start delegates to `startWatching` — but I pinned that path separately anyway, because "inherits it today" is not the same as "cannot regress".

Reverting the call to a bare `toggleWatching()` reddens exactly `testRemoteStartDoesNotRequestAccessibility` and `testRemoteToggleToStartDoesNotRequestAccessibility`, and nothing else.

## Conflict resolutions worth checking

Three places where your merged work and this branch touched the same lines. All additive, but they are the spots to look at first:

- **`requestStartPermissions`** — kept as you wrote it; the idempotent-control block now sits after it rather than in its place.
- **`toggleWatching`** — kept your `userInitiated:` signature and doc comment, with this branch's `if isManualRecording { return }` guard, which is the `manualStartTask`-aware version rather than the loop-only one.
- **`performManualRecording`** — kept your improved comment about the consent gate and the deny-list store, at this branch's indentation, since the body had been extracted out of the task closure to satisfy `closure_body_length`.

## Verification

144 tests green across `WatchingControllerTests`, `AppStateTests` and `DebugRPCServerIntegrationTests`.

Lint is clean via `scripts/lint.sh` — 0 violations in 441 files. Note for anyone else rebasing: the pin moved to SwiftFormat **0.62.1** in `scripts/tool-versions.sh`, and 0.61.1 now fails outright on `.swiftformat` with `Unknown rule 'wrapIfStatementBodies'`, so an old local copy will not just disagree, it will refuse to run.

The earlier round's mutation checks still hold on the rebased tree: dropping the join from `stopWatching`, reverting `isManualRecording` to loop-only, and deleting the wiring arguments each redden their own test.

alexpfau:
All three fixed, plus the four nits. `df5fa56`, and `0e3e5b4` before it.

On timing: item 3 was already done when you wrote it. I rebased and pushed `0e3e5b4` at 19:40 UTC, five minutes after this review went up, so we crossed. Same conclusion, same reasoning — `startWatching` passes `userInitiated: false`, `stopWatching` needed nothing because its call takes the stop branch, and `applyWatchAction(.toggle)` inherits it but is pinned separately anyway.

## 1. The discarded `joinStart` result

Took your placement rather than the obvious one. The guard is in the auto start task immediately after its permission gate, and the `pipeline.rebuild()` argument is the part I would have missed: bailing at the assignment still leaves the manual loop wired to a queue that is about to be replaced, which is a quieter failure than the orphan itself.

I also agree on the alternative you rejected. Acting on the `false` in the manual path trades a race for a dead Record App button behind an unanswered dialog, which is worse in the case that actually happens to people.

## 2. `AppState`'s own copy

Forwarded. Your note about grepping for the body rather than the name was the useful part — the name alone reads like an ordinary caller. A grep for the expression now shows one definition and no other copies.

**Not pinned by a test, and I want to be straight about why.** Catching the in-flight half means constructing a state where `manualStartTask` is set but no manual loop exists yet, and that needs a parked mic gate on `AppState`'s own controller. `AppState.init` does not expose the seam — which was your call last round on #607, and a good one for the reason you gave then. The forwarding is verified by mutation at the source level only. If this bug changes your mind about that seam, say so and I will add it; I did not want to reverse a decision of yours on my own.

## The four nits

All taken. The re-entrancy guard, the cancellation check, 10 ms → 50 ms, and the `toggleWatching` refusal narrowed to a *manual* loop so a stop in that sliver no longer returns an instant 503.

The 503 sliver was the one worth catching. It is a real behaviour change the commit message did not mention, and it inverts the endpoint's own contract: a documented "did not settle within 20 seconds" arriving in about one main-actor hop.

On the hang-instead-of-red test: leaving it, and your framing is right that it is a tripwire. Worth adding that it cannot easily be converted — racing the call against a deadline in a task group would hit exactly the deadlock I fixed in `joinStart`, because a group awaits every child and `Task.value` ignores cancellation. Abandoning the wait needs an unstructured task and a polled flag, which is more machinery than a hang-as-signal is worth.

## Verification, and a correction to my own method

Both new tests are checked against the mutation they target:

| Mutation | Result |
|---|---|
| Drop the auto-task `guard !isManualRecording` | `testParkedAutoStartBailsWhileAManualStartIsInFlight` red, alone |
| Drop `guard manualStartTask == nil` | `testSecondManualStartWhileOneIsInFlightIsIgnored` red, alone |

My first versions of both **passed under mutation**, and the reason is worth recording. The manual path's own `joinStart` waits out the parked auto start and then stops its loop, which masks the orphan entirely. Reproducing your scenario needs the join to *give up*, so the test injects a 50 ms `startJoinTimeout` and parks the two paths on separate gates — sharing one leaves the resume order undefined. An earlier attempt also leaned on a real manual loop staying manual, which it does not: a recording of pid 99 ends on its own, so `isManualRecording` flipped false before the guard ran and the test failed for a reason that had nothing to do with the code.

Two process notes from that: my mutation harness was greping only for failures, so a build error read as "no failures" — it now distinguishes them. And a test that passes under mutation is worth more attention than one that fails, since it looks like success.

146 tests green across the three affected suites. `scripts/lint.sh` clean, 0 violations in 441 files, against the 0.62.1 pin.

pasrom:
Heads-up before you pick anything else up: a closer look at `df5fa56` turned up three small things, and since the fold is still ahead of us anyway, I am pushing them to your branch as `fixup!` commits rather than sending you round again. Nothing structural, about ten lines total. Details in the review right after this.

Hold off on folding for a few minutes so we do not collide.

pasrom (approved):
Approving. I pushed the three remaining points myself as `fixup!` commits rather than sending you round a fourth time: `5455731`, `8d651df`, `94313b7`. They target `cf056a0`, `51d4358` and `093b58a`, so they cost you nothing in history and disappear in the fold that was already ahead of us.

## What they change

**1. `AppState.isManualRecording` goes back to loop-only, and both RPC sites read `watching.isManualRecording` instead.**

The forward at the property reached a third consumer neither of us named: `MeetingTranscriberApp.swift:111` gates the menu-bar "Stop Recording" item on it. So during the in-flight window that item appeared, displaced "Record App..." (the menu's `else if` sees `.idle`), and did nothing when pressed.

There is also a narrower window where it does harm rather than nothing. Pressed after `watchLoop = loop` but before `WatchLoop.startManualRecording` sets `manualRecordingInfo`, the guard there returns early but the controller still nils its own reference. Once `performManualRecording` returns and its local goes out of scope, nothing holds the loop: the state-change handler captures it weakly and the monitor task's `guard let self` has not run yet. It deallocates with a live recorder and the recording is never enqueued. Unreachable before this branch, because the item stayed hidden until that info was set.

The split is deliberate rather than a revert: the menu asks "is there a recording I can stop", the automation API asks whether a manual recording owns *or is about to own* the loop. Different questions, so two predicates. `/state` takes the wider one, for consistency with the 409 it explains.

My "grep the body, not the name" is what steered you past this, and it was half the advice. The body grep finds second definitions, the name grep finds consumers. It needed both.

**2. `startWatching` re-checks after the join.**

When the guard you added bails, `watchLoop` stays nil, so the old line returned `.failed` and the route answered 503 with `manualRecording: true` in the body. But the join settled and the dialog was answered, so 503 describes something that did not happen. It is the conflict 409 exists for, and it is the same body/status contradiction you fixed for the 409 body.

Pinned by `testStartRefusedByAMidFlightManualStartReportsBlocked`. It orders the race through the existing `requestScreenRecording` seam, which fires inside the auto task after the mic gate and before your guard, so the manual start registers mid-flight without the test betting on scheduling. Dropping the re-check turns it red with `"failed" is not equal to "blocked"`.

**3. The status table still said 503 means a denied microphone**, which the prose you added further up now contradicts.

## On the seam

No, this does not change my mind, and thank you for asking rather than reversing it. With the read at the DTO site it is one line naming a property your own tests already pin, so what stays untested is which property that line names, which is visible on inspection. A construction seam in `AppState.init` to cover one field of one DTO is out of proportion. If more of `/v1` starts depending on AppState-level composition, it earns itself and I will say so.

## Not in scope, and I will file it

A second manual start while one is *live* passes your new guard, since `manualStartTask` is nil again by then, and `performManualRecording` deliberately skips stopping a manual loop. So it overwrites a live one without enqueuing it, and that orphan keeps capturing because its monitor task retains it. Identical at the parent commit, so not yours. Only worth saying because the new guard's comment should not be read as covering it.

## Optional

The injected 50 ms `startJoinTimeout` in the parked-start test now leaves one poll cycle of margin, since the interval is also 50 ms. It only bites under whole-process starvation, but roughly 250 ms would be quieter on the CI machine.

## Verification

147 tests green across `WatchingControllerTests`, `AppStateTests` and `DebugRPCServerIntegrationTests`. `scripts/lint.sh` clean, 0 violations in 441 files, against the 0.62.1 pin.

## Before merge

The fold, which is now yours plus my three. And one round with `run-e2e`, since those lanes are label-gated and skipped on every run so far, while this branch changes watch start, watch stop and the manual-recording path.

Good round. The task-group deadlock and the two tests that passed under mutation were both yours to catch and you caught them; that is the part that does not show up in a diff.

pasrom:
Folded and force-pushed, so **re-point your local branch before touching it again**: `git fetch origin && git reset --hard origin/watch-control-api` (or re-clone the branch). Sorry for the rewrite, it is the one step that cannot be additive.

Eleven commits became four, split strictly by path:

| Commit | Holds |
| --- | --- |
| `33ad240` `feat(app)` | all `Sources/` |
| `0648da3` `test(app)` | all `Tests/` |
| `1d9e255` `feat(build)` | `tools/mt-cli/` |
| `01cec31` `docs` | `docs/` + `README.md` |

Dissolved into those: the SwiftFormat commit (its hunks belonged to the files the same PR had just added, so it was never a separate logical change), `911c3b5` which spanned four targets, `0e3e5b4`, `df5fa56`, and my three fixups.

Verified behaviour-preserving rather than assumed: `git diff 94313b7d HEAD` is empty, so the end state is byte-identical to what CI last went green on. Both intermediate commits also build on their own, `swift build` at the first and `--build-tests` at the second, so the history stays bisectable.

One consequence worth naming: a fold keeps the *target* commit's author, so the three fixes I pushed now sit under your name rather than mine. That is inherent to folding, not an accident, but you should know it is there.

Nothing needed from you on this. Remaining before merge is the `run-e2e` round.

pasrom:
Correcting myself on the `run-e2e` ask: it was not actionable and you should not wait on it. Those lanes execute PR code on the self-hosted Mac mini, so they are gated on `head.repo.full_name == github.repository` (`e2e.yml:61`, `e2e-app.yml:60`) and never run for a fork PR. I applied the label, watched it skip anyway, and have removed it again so it does not read as a pending requirement.

The coverage still arrives: `e2e.yml` runs on every push to `main`, which is the designed path for fork contributions. So nothing is outstanding on your side.

Also filed the manual-recording overwrite as #624, with the note that it predates this branch and that your re-entrancy guard was never meant to cover it.

Nothing left blocking from where I sit.

PR #607, opened 2026-08-10
fix(app): request Accessibility permission at watch start

### LINKED ISSUE #606: Accessibility permission is reported as a problem but never requested
## Summary

The app reports Accessibility as a permission problem — red "!" badge on the menu bar icon plus a notification — but it never asks the user for that permission. `Permissions.ensureAccessibilityAccess()` is the only code that would show the macOS Accessibility prompt, and it has no callers. The user is left with a warning about a problem the app never offers to fix, and has to work out on their own that the answer is System Settings → Privacy & Security → Accessibility.

## What happens

1. `PermissionHealthCheck.runLive()` calls `checkAccessibilityLive()`, which returns `.denied` when `AXIsProcessTrusted()` is false.
2. That flows into `HealthCheckResult.problems`, so `isHealthy` is false.
3. `MenuBarIcon` composites the red "!" badge and `NotificationManager` posts the permission notification.
4. Nothing anywhere calls `Permissions.ensureAccessibilityAccess()`, so the prompt that would let the user grant it in one click never appears.

Step 4 is easy to confirm: `git grep -n ensureAccessibilityAccess` returns exactly one hit, the declaration itself at `Permissions.swift:82`. The function also carries `// swiftlint:disable:next unused_declaration` on line 81, which suggests the dead code was noticed at some point and silenced rather than wired up or removed.

## Why the permission is genuinely needed

This is not a case where the health check is simply checking for something the app does not use. `ParticipantReader.readParticipants(pid:)` calls `AXUIElementCreateApplication(pid)` (`ParticipantReader.swift:29`) against the meeting app's process to read the participant roster, and it is called from `WatchLoop.swift:398`. Cross-process Accessibility API access is TCC-gated, so without the grant that read fails and the roster is unavailable for speaker naming and diarization.

So the health check is right to flag it. The gap is that the app diagnoses the problem without ever offering the fix.

## Suggested fix

Wire `ensureAccessibilityAccess()` into the watch-start path the same way screen recording already is. `WatchingController.swift:85` takes the screen recording prompt as an injected default closure:

```swift
requestScreenRecording: @escaping () -> Void = { Permissions.ensureScreenRecordingAccess() },
```

An equivalent `requestAccessibility` parameter defaulting to `{ _ = Permissions.ensureAccessibilityAccess() }` would keep the same testability and the same one-shot prompt semantics, since `ensureAccessibilityAccess()` already guards itself with `accessibilityPromptLock` and returns early when `AXIsProcessTrusted()` is true.

The alternative — dropping Accessibility from the health check to silence the badge — would be the wrong direction, because `ParticipantReader` really does need the grant and would silently degrade instead.

## Environment

macOS 26, current `main`. Happy to open a PR for this if the approach sounds right.

alexpfau:
Closes #606.

## The problem

`PermissionHealthCheck` treats a missing Accessibility grant as a problem: it composites the red `!` badge onto the menu bar icon and posts a notification. But nothing ever called `Permissions.ensureAccessibilityAccess()`, so the app diagnosed the problem and never offered the fix. The only way to grant it was to find the Accessibility pane in System Settings unaided. The function was dead code — it carried a `// swiftlint:disable:next unused_declaration` suppression to keep the analyze lane quiet, which is a fair tell that it had lost its caller.

## The fix

Ask for the grant in `toggleWatching`, next to the existing microphone and Screen Recording requests. That is where the code path needing it begins: `WatchLoop.readParticipants` reaches for the Teams roster through `AXUIElementCreateApplication`, and its only call site is the auto-detect recording path that watching starts. Prompting from the health check instead would fire on every activation and interrupt someone trying to work — the same reason the Screen Recording request already lives here.

Unlike its two siblings the request is **narrowed on three axes**, because full computer control is the most invasive grant on macOS and this one only enriches a recording:

| Gate | Why |
|---|---|
| `userInitiated` | Auto-watch reaches `toggleWatching` three seconds after launch, so an unconditional request raises a system alert with no click of any kind — on every launch, and mid-lane on the e2e runners, where a dialog can take the keystroke `--naming-escape` sends |
| `settings.watchTeams` | The roster read is the grant's only consumer and is itself Teams-gated. Someone watching only Zoom, Webex or a browser must not be asked for it |
| `#if !APPSTORE` | The sandbox denies the Accessibility API against other processes, so the App Store variant would ask for a grant it can never use — and following the prompt would leave a sandboxed user permanently `.broken`, a state whose documented remedy cannot clear it |

`toggleWatching` gains `userInitiated`, defaulting to true so a deliberate press keeps asking; the auto-watch notification path passes false.

### No one-shot flag

The previous revision relied on `ensureAccessibilityAccess` being one-shot per session. That was the wrong granularity in both directions, and it is now removed rather than re-armed. The alert can be dismissed unanswered, and can stack behind the Screen Recording alert raised two lines earlier, so a process-wide flag burns on that dismissal and never asks again inside the session — precisely the dead end this PR set out to remove. With the call bounded to a deliberate Start Watching with Teams watching enabled, every call is already a moment the user asked for the feature that needs the grant.

Both `permission_denied` warnings go with it. `AXIsProcessTrustedWithOptions` reports trust as of the call and never waits for the user, so `user_denied_prompt` was logged while the dialog was still on screen and nothing had been denied; `already_prompted` logged the guard doing its job. Both were newly reachable here, both ship in the diagnostics bundle users attach to bug reports, and both would misdirect triage. The `Bool` return had no consumer once the call site discarded it, so the signature is now `Void`, matching `ensureScreenRecordingAccess`.

## On `startManualRecording`

An earlier revision of this description claimed that path "never reads participants, so it has no use for the grant". Both halves are true and neither is the reason it fails — as measured in review, it gates on overall permission health, so an outstanding Accessibility grant blocks it via `HealthCheckResult.problems` regardless of what it uses. That claim is withdrawn. The gate itself is tracked separately as #616 and is not carried here; adding a request to that path would paper over a gate that is wrong for a permission the app's own Settings calls "Optional".

## Testing

Three cases now pin the behaviour, and each was verified by mutation rather than assumed:

| Mutation | Result |
|---|---|
| Drop the `watchTeams` half of the gate | `testToggleWatchingSkipsAccessibilityWhenTeamsNotWatched` red, alone |
| Drop the `userInitiated` half | `testAutoWatchStartDoesNotRequestAccessibility` red, alone |
| Drop the gate entirely | Both red |
| Comment out `requestAccessibility()` | `testToggleWatchingRequestsAccessibility` red |

`AppStateTests.testToggleWatchingCreatesWatchLoop` and `testToggleWatchingMakesLoopActive` reach the production defaults, because `AppState.init` does not expose the permission seams. They now drive the auto-watch path, which exercises the same start without a real TCC call from a unit test — a lighter fix than widening that init, whose body type-check time is already budgeted.

Verified locally:

- SwiftFormat **0.61.1**, the version pinned at this branch point, clean (the pin has since moved to 0.62.1; re-checked below)
- `swiftlint lint --strict` — clean
- 124 tests green across `WatchingControllerTests`, `AppStateTests`, `PermissionsTests` and `PermissionHealthCheckTests`

## Not addressed here

The participant-PID question (whether `PowerAssertionDetector` hands `ParticipantReader` a helper PID with no roster panel, making the grant buy nothing today) is left as the follow-up you filed it as. It deserves a measurement against a live Teams call rather than a guess, and it does not block closing the "reported but never requested" gap.

## Landed as

Rebased onto `main` and folded into a single commit, since the two revisions are one logical change that was split only by review.

The rebase surfaced one thing neither side could see before it. `main` had meanwhile grown the watch-start closure (the loop it builds takes another argument now), so the three lines added here pushed it one line past `closure_body_length`. The `swiftlint --strict` run above was genuinely clean at the branch point and only fails against current `main`. Fixed by moving the three permission requests into `requestStartPermissions`, which gives them one name instead of suppressing the rule. The pinned SwiftFormat version drifted the same way, 0.61.1 at the branch point and 0.62.1 now.

Re-verified on the rebased commit: the full mutation table reproduces with the same per-case granularity, `./scripts/lint.sh` and `swiftlint analyze --strict` clean against the currently pinned versions, 2591 tests green, and a release build of both the Homebrew and the sandboxed variant, which is what proves the `#if !APPSTORE` default argument still compiles to a valid closure.

### DIFF
--- a/app/MeetingTranscriber/Sources/Permissions.swift
+++ b/app/MeetingTranscriber/Sources/Permissions.swift
@@ -77,28 +77,29 @@ enum Permissions {
         return false
     }
 
-    private static let accessibilityPromptLock = OSAllocatedUnfairLock(initialState: false)
-    // swiftlint:disable:next unused_declaration
-    static func ensureAccessibilityAccess() -> Bool {
-        if AXIsProcessTrusted() { return true }
-        let alreadyPrompted = accessibilityPromptLock.withLock { prompted -> Bool in
-            if prompted { return true }
-            prompted = true
-            return false
-        }
-        guard !alreadyPrompted else {
-            logger.warning("permission_denied resource=accessibility status=already_prompted")
-            return false
-        }
+    /// Ask macOS for Accessibility, every time the caller decides the moment is
+    /// right.
+    ///
+    /// Unlike its Screen Recording sibling this keeps no one-shot flag. The
+    /// alert can be dismissed without an answer, and it can stack behind the
+    /// Screen Recording one raised moments earlier; a process-wide flag burns on
+    /// that dismissal and never asks again, which is the dead end this call was
+    /// added to remove. `WatchingController` gates it on a deliberate Start
+    /// Watching with Teams watching enabled, so "every call" already means
+    /// "every time the user asks for the feature that needs it".
+    ///
+    /// Deliberately silent. `AXIsProcessTrustedWithOptions` reports the trust
+    /// state as of the call and never waits for the user, so it returns false
+    /// while the dialog is still on screen. Logging that as a denial would put a
+    /// warning nobody caused into the diagnostics bundle users attach to bug
+    /// reports. `PermissionHealthCheck` reports the real state on the next
+    /// activation.
+    static func ensureAccessibilityAccess() {
+        guard !AXIsProcessTrusted() else { return }
         // `kAXTrustedCheckOptionPrompt` is a C global imported as
         // `Unmanaged<CFString>!`, which Swift 6 treats as shared mutable
         // state. The value is set by AppKit at process load and never
         // mutates; bridge once via a nonisolated(unsafe) wrapper.
-        let options = [Self.axPromptKey: true] as CFDictionary
-        let granted = AXIsProcessTrustedWithOptions(options)
-        if !granted {
-            logger.warning("permission_denied resource=accessibility status=user_denied_prompt")
-        }
-        return granted
+        _ = AXIsProcessTrustedWithOptions([axPromptKey: true] as CFDictionary)
     }
 }
--- a/app/MeetingTranscriber/Sources/WatchingController.swift
+++ b/app/MeetingTranscriber/Sources/WatchingController.swift
@@ -63,6 +63,34 @@ final class WatchingController {
     /// check reports the state.
     private let requestScreenRecording: () -> Void
 
+    /// Accessibility request, fired at watch start. Injectable so tests skip
+    /// the real TCC prompt.
+    ///
+    /// `PermissionHealthCheck` already reports a missing Accessibility grant as
+    /// a problem — red menu-bar badge plus a notification — because
+    /// `ParticipantReader` needs it to read the Teams roster via
+    /// `AXUIElementCreateApplication`. Without this call nothing ever showed the
+    /// prompt, so the app diagnosed the problem but never offered the fix and
+    /// the user had to find the Accessibility pane unaided.
+    ///
+    /// It belongs at watch start for the same reason the Screen-Recording
+    /// request does: the health check runs on every activation, so prompting
+    /// there would interrupt a user who is trying to work.
+    ///
+    /// Narrower than its two siblings on three axes, because full computer
+    /// control is the most invasive grant on macOS and this one only enriches a
+    /// recording. It fires only from a user-initiated toggle, since auto-watch
+    /// otherwise raises a system alert three seconds after every launch with no
+    /// click of any kind; only when `watchTeams` is on, since the roster read is
+    /// the grant's sole consumer and is itself Teams-gated; and the production
+    /// default is compiled out of the sandboxed build, which cannot use the API
+    /// against another process at all.
+    ///
+    /// The result is ignored, like the other two gates. Watching, capture and
+    /// transcription all work without the grant, so a refusal must not block the
+    /// start.
+    private let requestAccessibility: () -> Void
+
     /// Meeting detector factory for the auto-detect path. Injectable so tests can
     /// supply a deterministic detector instead of the IOKit-backed
     /// `PowerAssertionDetector`.
@@ -83,6 +111,11 @@ final class WatchingController {
         liveTranscription: LiveTranscriptionCoordinator,
         ensureMicAccess: @escaping () async -> Bool = { await Permissions.ensureMicrophoneAccess() },
         requestScreenRecording: @escaping () -> Void = { Permissions.ensureScreenRecordingAccess() },
+        requestAccessibility: @escaping () -> Void = {
+            #if !APPSTORE
+                Permissions.ensureAccessibilityAccess()
+            #endif
+        },
         makeDetector: (() -> any MeetingDetecting)? = nil,
     ) {
         self.settings = settings
@@ -93,6 +126,7 @@ final class WatchingController {
         self.liveTranscription = liveTranscription
         self.ensureMicAccess = ensureMicAccess
         self.requestScreenRecording = requestScreenRecording
+        self.requestAccessibility = requestAccessibility
         // Tests inject a deterministic detector; production defaults to one
         // filtered by the "Apps to Watch" toggles, re-read at each watch start.
         self.makeDetector = makeDetector ?? { [settings] in
@@ -145,7 +179,12 @@ final class WatchingController {
 
     // MARK: - Start / Stop
 
-    func toggleWatching() {
+    /// - Parameter userInitiated: True for a deliberate Start/Stop press, which
+    ///   is the one moment where asking for an optional permission is warranted.
+    ///   The auto-watch path passes false so an unattended launch raises no
+    ///   system alert — including on the e2e runners, which force auto-watch on
+    ///   and where a stray dialog can swallow the keystroke a lane sends.
+    func toggleWatching(userInitiated: Bool = true) {
         if let loop = watchLoop, loop.isManualRecording { return }
         if let loop = watchLoop, loop.isActive {
             loop.stop()
@@ -158,8 +197,7 @@ final class WatchingController {
             guard startTask == nil else { return }
             startTask = Task { @MainActor in
                 defer { startTask = nil }
-                _ = await ensureMicAccess()
-                requestScreenRecording()
+                await requestStartPermissions(userInitiated: userInitiated)
 
                 syncEngines?()
                 pipeline.rebuild()
@@ -192,6 +230,21 @@ final class WatchingController {
                 watchLoop = loop
                 loop.start()
             }
+        }
+    }
+
+    /// The permissions a watch start asks for, in the order the user meets
+    /// them. Only the mic gate is awaited, because capture cannot begin without
+    /// it; the other two register the app in their System Settings pane and are
+    /// reported by `PermissionHealthCheck` afterwards, so a refusal delays
+    /// nothing here.
+    ///
+    /// - Parameter userInitiated: see `toggleWatching`.
+    private func requestStartPermissions(userInitiated: Bool) async {
+        _ = await ensureMicAccess()
+        requestScreenRecording()
+        if userInitiated, settings.watchTeams {
+            requestAccessibility()
         }
     }
 
--- a/app/MeetingTranscriber/Sources/MeetingTranscriberApp.swift
+++ b/app/MeetingTranscriber/Sources/MeetingTranscriberApp.swift
@@ -141,7 +141,7 @@ struct MeetingTranscriberApp: App {
             }
             .onReceive(NotificationCenter.default.publisher(for: .autoWatchStart)) { _ in
                 if !appState.isWatching {
-                    appState.watching.toggleWatching()
+                    appState.watching.toggleWatching(userInitiated: false)
                 }
             }
             .onReceive(NotificationCenter.default.publisher(for: .showSpeakerNaming)) { _ in
--- a/app/MeetingTranscriber/Tests/AppStateTests.swift
+++ b/app/MeetingTranscriber/Tests/AppStateTests.swift
@@ -257,7 +257,11 @@ final class AppStateTests: XCTestCase { // swiftlint:disable:this type_body_leng
         let (state, _) = makeState()
         addTeardownBlock { state.watching.watchLoop?.stop() }
 
-        state.watching.toggleWatching()
+        // `AppState.init` does not expose the permission seams, so this is the
+        // one path that reaches the production defaults. Auto-watch exercises
+        // the same start without the optional Accessibility prompt, which would
+        // otherwise be a real TCC call from a unit test.
+        state.watching.toggleWatching(userInitiated: false)
         await waitFor(state.watching.watchLoop != nil)
 
         XCTAssertNotNil(state.watching.watchLoop)
@@ -267,7 +271,7 @@ final class AppStateTests: XCTestCase { // swiftlint:disable:this type_body_leng
         let (state, _) = makeState()
         addTeardownBlock { state.watching.watchLoop?.stop() }
 
-        state.watching.toggleWatching()
+        state.watching.toggleWatching(userInitiated: false)
         await waitFor(state.watching.watchLoop?.isActive == true)
 
         XCTAssertEqual(state.watching.watchLoop?.isActive, true)
--- a/app/MeetingTranscriber/Tests/WatchingControllerTests.swift
+++ b/app/MeetingTranscriber/Tests/WatchingControllerTests.swift
@@ -29,9 +29,12 @@ final class WatchingControllerTests: XCTestCase {
     private func makeController(
         ensureMicAccess: @escaping () async -> Bool = { true },
         requestScreenRecording: @escaping () -> Void = {},
+        requestAccessibility: @escaping () -> Void = {},
+        watchTeams: Bool = true,
         makeDetector: @escaping () -> any MeetingDetecting = { makeSilentDetector() },
     ) -> WatchingController {
         let settings = AppSettings()
+        settings.watchTeams = watchTeams
         let notifier = RecordingNotifier()
         let pipeline = PipelineController(settings: settings, notifier: notifier)
         pipeline.queue = PipelineQueue(logDir: tmpDir)
@@ -56,6 +59,7 @@ final class WatchingControllerTests: XCTestCase {
             liveTranscription: liveTranscription,
             ensureMicAccess: ensureMicAccess,
             requestScreenRecording: requestScreenRecording,
+            requestAccessibility: requestAccessibility,
             makeDetector: makeDetector,
         )
     }
@@ -79,6 +83,64 @@ final class WatchingControllerTests: XCTestCase {
         await waitFor(requested)
 
         XCTAssertTrue(requested, "toggleWatching must ask for Screen Recording")
+    }
+
+    // MARK: - requestAccessibility seam
+
+    /// The health check already flags a missing Accessibility grant with a red
+    /// menu-bar badge and a notification, but nothing used to ask for it, so the
+    /// app reported a problem it never offered to fix. Asking on a deliberate
+    /// Start Watching — where the participant read that needs it happens — is
+    /// what closes that gap, and pinning it here keeps the request from being
+    /// dropped again.
+    func testToggleWatchingRequestsAccessibility() async {
+        var requested = false
+        // Not trailing-closure: a trailing closure binds to the last param
+        // (`makeDetector`), not `requestAccessibility`.
+        // swiftlint:disable:next trailing_closure
+        let controller = makeController(requestAccessibility: { requested = true })
+        addTeardownBlock { await controller.watchLoop?.stop() }
+
+        controller.toggleWatching()
+        await waitFor(requested)
+
+        XCTAssertTrue(requested, "toggleWatching must ask for Accessibility")
+    }
+
+    /// The roster read is the grant's only consumer and is itself Teams-gated,
+    /// so someone watching only Zoom, Webex or a browser must never be asked for
+    /// full computer control. The prompt also used to burn a process-wide
+    /// one-shot, so enabling Teams later in the same session then never asked
+    /// for the grant that had by then become necessary.
+    func testToggleWatchingSkipsAccessibilityWhenTeamsNotWatched() async {
+        var requested = false
+        let controller = makeController(
+            requestAccessibility: { requested = true },
+            watchTeams: false,
+        )
+        addTeardownBlock { await controller.watchLoop?.stop() }
+
+        controller.toggleWatching()
+        await waitFor(controller.watchLoop != nil)
+
+        XCTAssertFalse(requested, "watchTeams off must not ask for Accessibility")
+    }
+
+    /// Auto-watch reaches `toggleWatching` three seconds after launch with no
+    /// click of any kind, so an unconditional request raised a system alert on
+    /// every launch. It also fires mid-lane on the e2e runners, which force
+    /// auto-watch on, where a dialog can take the keystroke the naming lane
+    /// sends and fail it for a reason unrelated to the code under test.
+    func testAutoWatchStartDoesNotRequestAccessibility() async {
+        var requested = false
+        // swiftlint:disable:next trailing_closure
+        let controller = makeController(requestAccessibility: { requested = true })
+        addTeardownBlock { await controller.watchLoop?.stop() }
+
+        controller.toggleWatching(userInitiated: false)
+        await waitFor(controller.watchLoop != nil)
+
+        XCTAssertFalse(requested, "auto-watch must not ask for Accessibility")
     }
 
     // MARK: - ensureMicAccess seam


app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -152,6 +175,7 @@ final class WatchingController {
                 defer { startTask = nil }
                 _ = await ensureMicAccess()
                 requestScreenRecording()
+                requestAccessibility()
```
pasrom: **The stated reason for skipping `startManualRecording` does not hold.** The PR text says that path "never reads participants, so it has no use for the grant". Both halves are true, and neither is why it fails. It fails because it gates on overall permission health rather than on what it uses:

```swift
// WatchLoop.swift:224-227
let health = await permissionChecker()
if !health.isHealthy {
    throw RecorderError.permissionDenied(health.notificationBody)
}
```

and `HealthCheckResult.problems` (PermissionHealthCheck.swift:87-90) turns `accessibility == .denied` into a problem.

Measured on this branch, with Screen Recording healthy and Accessibility the only outstanding permission, so `isHealthy` was false for that one reason:

| Accessibility | "Record App..." |
|---|---|
| `denied` | notification "Error / Permission problem: Accessibility permission denied", no prompt, nothing recorded |
| `healthy` | "Manual Recording", records |

Same app, same action, same order. The only difference is the grant.

**To be clear about scope: this is not a regression you introduced.** That gate behaves identically on `main`, and your change strictly improves the situation, since a user who presses Start Watching first can now fix it at all. I have filed the gate itself as #616 rather than asking you to carry it here.

What I would change in this PR is the justification. As written it reads as "this path is fine without the grant", and the measurement says it is not. Either soften the claim to what you actually verified, or land #616 first so the claim becomes true. Adding `requestAccessibility()` here would be the third option and I would argue against it: it would paper over a gate that is wrong for a permission the app's own Settings calls `"Optional — enables mute detection and meeting naming"` (AdvancedSettingsView.swift:56), and that the auto-detected path records without.

pasrom: **Not gated on `watchTeams`, although the grant has exactly one consumer and that consumer is Teams-gated.**

```swift
// WatchLoop.swift:397-398
if meeting.pattern.appName == "Microsoft Teams",
   let names = ParticipantReader.readParticipants(pid: meeting.windowPID),
```

A user who turned Teams off in "Apps to Watch" and watches only Zoom, Webex or browser meetings is asked for full computer control for something the app will never do.

The second-order effect is worse than the nuisance: that unwanted prompt burns the one-shot flag. If the same user enables Teams later in the same session, the next watch start takes the `already_prompted` branch and never asks for the grant they now actually need.

Gating on `settings.watchTeams` fixes both, and it is the idiom already in this file: `defaultDetectors` filters by `settings.watchApps` instead of instantiating everything unconditionally. Of everything in this review this is the cheapest change with the largest effect, so it is the one I would start with.

pasrom: **This can turn an e2e lane red with nothing wrong in the code under test.**

Both live lanes force auto-watch on:

```
scripts/e2e-app.sh:397      defaults write "$DEV_BUNDLE_ID" autoWatch -bool true
scripts/e2e-browser.sh:235  defaults write "$BUNDLE_ID" autoWatch -bool true
```

So three seconds after launch the runner's app takes this call. If the build on that machine is not Accessibility-trusted, a system dialog appears unprompted, mid-lane.

The `--naming-escape` lane is the one that hurts. It sends a real Escape through the WindowServer to whatever is frontmost, which is exactly why that lane needs an Accessibility grant on the runner in the first place. A permission dialog that appears around the same moment takes the keystroke, the naming panel is never dismissed, and the lane fails for a reason that has nothing to do with the change being tested. There is a second-order cost too: whatever the dialog resolves to becomes persistent runner TCC state, changed as a side effect of CI rather than deliberately.

I cannot check the runner from here, so this hinges on one question: is the dev build on that machine already in the Accessibility list? If the grant there covers the keystroke-sending process rather than the app, this fires. Gating on `settings.watchTeams` would incidentally close it for the browser lane, but not for `e2e-app.sh`.

pasrom: **The one-shot flag has the wrong granularity in both directions, and this PR is what makes that reachable.**

It is process-scoped, so:

- *Too sticky inside a session.* Measured on this branch: within one process, three consecutive calls produced one `status=user_denied_prompt` and then two `status=already_prompted`. Only the very first attempt can ever show the dialog. A user who dismissed it, or never saw it because it stacked behind the Screen Recording alert raised two lines earlier in this same block, has no way back short of quitting and relaunching. That is the dead end this PR set out to remove.
- *Not sticky enough across sessions.* See the note on the doc comment above: a decline is forgotten at the next launch, forever.

A deliberate press of Start Watching is exactly the moment when re-asking is justified, and it is the moment where nothing happens. Re-arming the claim when the process is still untrusted at a later watch start would fix the first half; persisting a decline would fix the second. `ensureScreenRecordingAccess` already reasons about the neighbouring case in its doc comment ("if the grant is revoked mid-session, the next watch start still asks"), it just does not cover a prompt that was never answered.

pasrom: **Follow-up rather than a blocker, but worth settling before spending the most invasive TCC grant on macOS:** it is not clear the grant buys anything today.

`ParticipantReader` is handed `meeting.windowPID`, which `PowerAssertionDetector` sets to the PID of the assertion holder (`windowPID: match.pid`, PowerAssertionDetector.swift:182), and the Teams pattern deliberately matches helper processes:

```swift
// PowerAssertionDetector.swift:39
processNames: ["MSTeams", "Microsoft Teams", "Microsoft Teams WebView", "Microsoft Teams (work or school)"],
```

If the in-call assertion is held by the WebView helper, `AXUIElementCreateApplication` gets a process with no roster panel and no matching window title, all three read strategies return nil, and `participants` stays empty. `MicInputDetector` documents the same failure mode for its own PID source ("the reported PID is the daemon's ... and ParticipantReader has no window to read").

Code-derived only, I have no Teams call to check it against. One measurement showing a non-empty roster coming back through the production detector path would settle it, and it would also answer whether this prompt is worth asking for at all.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -63,6 +63,27 @@ final class WatchingController {
...
+    ///
+    /// It belongs at watch start for the same reason the Screen-Recording
+    /// request does: the health check runs on every activation, so prompting
+    /// there would interrupt a user who is trying to work.
+    ///
+    /// The result is ignored, like the other two gates. The grant is only needed
+    /// to enrich a Teams recording with its participant list; watching, capture
+    /// and transcription all work without it, so a refusal must not block the
+    /// start. `ensureAccessibilityAccess` is itself one-shot per session, so a
+    /// user who declines is not asked again on the next toggle.
```
pasrom: **"a user who declines is not asked again on the next toggle" holds only inside one process, and the prompt is not always the result of a toggle.**

`accessibilityPromptLock` is an `OSAllocatedUnfairLock` on a static, not a persisted decision, so the prompt returns at the next launch and the one after that, indefinitely. As written the sentence reads as a general suppression it does not provide.

And `toggleWatching` is not only reached by a click:

```swift
// MeetingTranscriberApp.swift:94-99
if (CommandLine.arguments.contains("--auto-watch")
    || UserDefaults.standard.bool(forKey: "autoWatch"))
    && !suppressAutoWatch {
    DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
        NotificationCenter.default.post(name: .autoWatchStart, object: nil)
    }
}
```

with the label's `onReceive` at MeetingTranscriberApp.swift:142 calling `watching.toggleWatching()`.

Measured on this branch with "start watching automatically" on and the grant revoked: the "would like to control this computer using accessibility features" alert appeared about three seconds after launch, with no click of any kind. Dismissed without granting, quit, relaunched: it appeared again. So a user who enabled auto-watch gets a system alert on every single launch, for a grant that only enriches a Teams roster.

Either persist the decline, or fire the request only from the user-initiated path.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -83,6 +104,7 @@ final class WatchingController {
         liveTranscription: LiveTranscriptionCoordinator,
         ensureMicAccess: @escaping () async -> Bool = { await Permissions.ensureMicrophoneAccess() },
         requestScreenRecording: @escaping () -> Void = { Permissions.ensureScreenRecordingAccess() },
+        requestAccessibility: @escaping () -> Void = { _ = Permissions.ensureAccessibilityAccess() },
```
pasrom: **Minor, and host-dependent, but the default puts a real TCC side effect inside a unit test.**

`AppState.init` (AppState.swift:231-238) constructs `WatchingController` without overriding this seam, and `AppStateTests.testToggleWatchingCreatesWatchLoop` (AppStateTests.swift:260) and `testToggleWatchingMakesLoopActive` (:270) call `state.watching.toggleWatching()` on exactly that object. `WatchingControllerTests` stubs the seam; those two cannot.

Whether that surfaces depends on the host. I ran both cases and saw nothing: on my machine the responsible process holds the Accessibility grant, so `AXIsProcessTrusted()` is true inside `xctest` and `ensureAccessibilityAccess` returns at its first line without prompting or logging. On a host without that grant the same call takes the prompt branch, and since `--parallel` forks a subprocess per test method the per-process one-shot does not dedupe it.

So this is not a blocker and it may well never fire on your runners. It is still worth a stub: a unit test reaching a TCC API at all is a side effect nobody reading `AppStateTests` would expect, and the failure mode on a fresh CI host is a dialog nobody is there to answer.

pasrom: **No `#if !APPSTORE` here or at the call site**, unlike the other host-only capabilities in this codebase.

The sandbox denies the Accessibility API against other processes, so the App Store variant now asks for full computer control it can never use. Worse than a wasted prompt: a sandboxed user sits at `accessibility == .denied` today, and if they follow the new prompt and add the app in System Settings, `AXIsProcessTrusted()` flips to true while `probeAccessibility()` still fails, so

```swift
// PermissionHealthCheck.swift:363-369
if !trusted { return .denied }
return probeSucceeds ? .healthy : .broken
```

returns `.broken`. They are left with a permanent red menu-bar badge and an `.accessibilityBroken` problem whose stated remedy, toggling the permission off and on, cannot clear it. Before this PR nothing pushed a sandboxed user into that state.

I did not run the sandboxed build, so treat the `.broken` half as unverified. The guard looks warranted regardless of how that comes out.

app/MeetingTranscriber/Sources/Permissions.swift
```diff
@@ -78,7 +78,6 @@ enum Permissions {
     }
 
     private static let accessibilityPromptLock = OSAllocatedUnfairLock(initialState: false)
-    // swiftlint:disable:next unused_declaration
     static func ensureAccessibilityAccess() -> Bool {
```
pasrom: **Making this function live turns an unanswered prompt into a logged denial.**

Line 82 has already established the process is untrusted, and `AXIsProcessTrustedWithOptions` reports the trust state as of the call rather than waiting for the user. So on the first watch start of an untrusted session line 97 returns false while the dialog is still on screen, and line 99 logs `permission_denied resource=accessibility status=user_denied_prompt` at warning level.

Measured on this branch. The alert went up, I answered it with "Open System Settings" and never denied anything, and the log had already recorded:

```
09:48:23.756  [com.meetingtranscriber:Permissions] permission_denied resource=accessibility status=user_denied_prompt
```

The same line appeared in a second run that was likewise never denied. There is no awaited variant of the AX call, so the honest fix is to drop the `if !granted` warning and let `PermissionHealthCheck` report the real state on the next activation.

The line is not new, but it was unreachable until this PR removed the `unused_declaration` suppression, so it never appeared before. `DiagnosticExporter` collects on `subsystem CONTAINS 'com.meetingtranscriber'`, which is where this logger writes, so it ships in the bundle users attach to bug reports and will misdirect triage.

(Anchored here because line 99 is outside the diff.)

pasrom: **Second logging issue, also newly reachable:** line 89 logs `permission_denied resource=accessibility status=already_prompted` at warning level.

`toggleWatching` is not once per session, so every start after the first produces one. Measured, three calls inside a single process:

```
10:05:15  status=user_denied_prompt
10:05:20  status=already_prompted
10:05:24  status=already_prompted
```

In the exported log those two are indistinguishable at a glance from a real denial. `already_prompted` is not a denial, it is the guard doing its job. Drop it, or move it to debug level.

app/MeetingTranscriber/Sources/Permissions.swift
```diff
@@ -78,7 +78,6 @@ enum Permissions {
...
         if AXIsProcessTrusted() { return true }
         let alreadyPrompted = accessibilityPromptLock.withLock { prompted -> Bool in
```
pasrom: **Two cleanups that only matter now that this function is live.**

**The one-shot is hand-rolled.** `claimFirst(_:)` at line 30 exists for exactly this, and says so: "Shared by the prompt guards below, which would otherwise each hand-roll the same test-and-set in a slightly different shape." Only `ensureScreenRecordingAccess` uses it; the plural was harmless while this was dead code. Now there are two divergent one-shot implementations, and a future change to the claim semantics applied to `claimFirst` would silently skip this path. That is not hypothetical given the note above on re-arming the claim. Lines 83 to 91 collapse to:

```swift
guard claimFirst(accessibilityPromptLock) else { return false }
```

(dropping the warning per the note above), which also makes this read like its Screen Recording sibling. Not offering it as a click-to-apply suggestion because the replacement spans lines outside the diff.

**The `Bool` return has no consumer.** The only caller the PR adds discards it (`{ _ = Permissions.ensureAccessibilityAccess() }`, WatchingController.swift:107), and the seam it feeds is typed `() -> Void`. While the function was unreachable that was invisible. Either drop the return type, matching `ensureScreenRecordingAccess`, or use it. As it stands the signature promises a decision the call site has already decided to ignore, which is also what makes the `false` returned from the `already_prompted` branch look meaningful when it is not.

pasrom (changes requested):
Closing the "diagnose but never offer the fix" gap is the right goal, and the injectable seam plus its wiring test are exactly the shape this file already uses. What I cannot follow is the placement: the request is unconditional, and it fires in situations where nothing asked for it.

I built this branch and ran it, so most of the notes below are observations rather than readings. Where a note is only code-derived, or where I could not reproduce something, I say so in the note itself.

Scope, so the weight lands where it belongs. Six notes are about behaviour this PR introduces. Three are about lines it does not touch but makes reachable for the first time by removing the `unused_declaration` suppression, so their symptoms ship with this change. Two are neither: the participant-PID note is a pre-existing question I have flagged as a follow-up, and the manual-recording note is a critique of the PR's stated reasoning rather than of its code, with the underlying bug filed separately as #616.

Read the e2e-runner note first. It is the one that can turn a lane red without anything being wrong in the code under test. After that, the cheapest change with the largest effect is gating the request on `settings.watchTeams`.

alexpfau:
Thanks — the measurements did the work here. Three of these I could not have found by reading, particularly the alert appearing three seconds after launch with no click, and `user_denied_prompt` being logged against a prompt you never answered.

You were right that gating on `watchTeams` was the place to start. It collapsed more than it looked like it would.

Pushed as a separate commit (`299f783`) rather than amended, as before.

## The eight points this changes

| Note | What changed |
|---|---|
| Not gated on `watchTeams` | Gate added; follows the `defaultDetectors` idiom you pointed at |
| e2e lane can go red | Request no longer fires from auto-watch at all, which covers `e2e-app.sh` too, not just the browser lane |
| One-shot wrong granularity | Flag removed rather than re-armed — reasoning below |
| Doc comment overclaims | Rewritten; the "not asked again on the next toggle" sentence is gone, and the auto-watch path is named |
| `unanswered prompt → logged denial` | `user_denied_prompt` warning dropped |
| `already_prompted` at warning level | Branch no longer exists |
| Hand-rolled one-shot | Gone, so no second divergent implementation |
| `Bool` return has no consumer | Signature now `Void` |
| Missing `#if !APPSTORE` | Added around the production default |
| Real TCC in `AppStateTests` | Those two tests now drive the auto-watch path |
| `startManualRecording` justification | Claim withdrawn from the PR description |

## Why the flag is removed rather than re-armed

You offered re-arming for the too-sticky half and persistence for the not-sticky-enough half. Removing it does the first and makes the second moot, and I think it is the smaller change once the gate is in.

The flag existed to stop nagging. With the request bounded to a deliberate Start Watching with Teams watching enabled, there is no unattended path left to nag from — which is your own framing: a deliberate press is exactly the moment when re-asking is justified. Persisting a decline would then work against that, since the recorded decline we would be honouring is frequently just a dismissal, and we cannot tell the two apart.

One consequence worth stating plainly: an untrusted user who presses Start Watching repeatedly now sees the alert each time. That is deliberate, and it is the behaviour your comment argued for.

This does leave `claimFirst` with a single caller again, so its "Shared by the prompt guards below" is now literally singular. I left the comment alone rather than widen the diff — happy to fix the wording if you would rather it read true.

## On `AppStateTests`

I looked at widening `AppState.init` with an injectable seam and decided against it: that init carries a documented body-type-check budget near the 300 ms limit, and adding a closure parameter to solve a test-only problem seemed a poor trade. Driving `toggleWatching(userInitiated: false)` instead keeps those two tests off the TCC API entirely, and it exercises a real production path rather than a stub. If you would rather have the explicit seam, say so and I will add it.

## Verification

Each new test was checked against the mutation it is meant to catch, not just for being green:

| Mutation | Result |
|---|---|
| Drop `settings.watchTeams` from the gate | Teams case red, alone |
| Drop `userInitiated` from the gate | Auto-watch case red, alone |
| Drop the gate entirely | Both red |

SwiftFormat **0.61.1** clean (caught one `redundantStaticSelf` that Homebrew's 0.65.0 did not flag — thanks for the earlier note about the pin, I checked against the pinned build this time), `swiftlint --strict` clean, 124 tests green across `WatchingControllerTests`, `AppStateTests`, `PermissionsTests` and `PermissionHealthCheckTests`.

## Left alone

The participant-PID question stays yours — it wants a measurement against a live Teams call, and I would rather settle it with data than fold a guess into this PR. Same for #616; I have withdrawn the claim in the description rather than reach into that gate from here.

app/MeetingTranscriber/Sources/WatchingController.swift
pasrom: Withdrawing the claim is the right call. The gate itself is fixed on `main` now, and not in the shape this note asked for: it narrows the refusal to the permissions that path actually needs, rather than adding a request to it. Nothing left here.

pasrom: Confirmed at my end. I reran the mutation: dropping `settings.watchTeams` reddens `testToggleWatchingSkipsAccessibilityWhenTeamsNotWatched` alone, dropping `userInitiated` reddens `testAutoWatchStartDoesNotRequestAccessibility` alone. The consumer chain still holds too, `ParticipantReader.readParticipants` has exactly one caller and it sits behind `meeting.pattern.appName == "Microsoft Teams"`, which only `watchTeams` puts into `watchApps`.

pasrom: Covered. Both lanes reach watching through `defaults write ... autoWatch`, so they arrive via `.autoWatchStart`, which now passes `userInitiated: false`. No dialog can be raised on the runner from this path.

pasrom: Removal works, and it is a better answer than either of the two I offered. With the call bounded to a deliberate start there is no unattended path left to nag from, so the flag has nothing to protect, and dropping it removes the dismissal-burns-the-flag failure in the same move.

pasrom: Agreed, this stays outside the PR. It wants a measurement against a live Teams call, and folding in a guess would be the wrong way to spend that grant.

app/MeetingTranscriber/Sources/WatchingController.swift
pasrom: No seam needed, your reading is right. Driving the auto-watch path keeps both tests off the real TCC API and exercises a production path instead of a stub, and it leaves `AppState.init` alone, whose body type-check budget is documented as sitting near the limit.

app/MeetingTranscriber/Sources/Permissions.swift
pasrom: The `Void` return is right. On the wording, leave it: `claimFirst` already had exactly one caller on `main`, because this function hand-rolled its own lock instead of using it. The plural was stale before this PR and is not yours to pay off.

pasrom (dismissed):
This is good. `299f7835` covers every note, and the parts that carry weight hold up when checked rather than read. All eleven threads are resolved.

## What I verified rather than took on trust

I reran your mutation table in a worktree at the PR head, and it reproduces with the same per-half granularity you reported:

| Mutation | Result |
|---|---|
| Drop `settings.watchTeams` from the gate | `testToggleWatchingSkipsAccessibilityWhenTeamsNotWatched` red, alone |
| Drop `userInitiated` from the gate | `testAutoWatchStartDoesNotRequestAccessibility` red, alone |
| Reverted | 13/13 green, tree clean |

The four suites run green here too (124 tests).

Two facts the change leans on, checked independently:

- The consumer chain is genuinely Teams-only. `ParticipantReader.readParticipants` has exactly one caller, behind `meeting.pattern.appName == "Microsoft Teams"`, and `watchTeams` is what puts Teams into `watchApps` in the first place. So a Zoom-only or browser-only user can never reach the grant's consumer, and now never gets asked for it.
- The e2e argument lands. Both lanes turn watching on with `defaults write ... autoWatch`, so they arrive through `.autoWatchStart`, which now passes `userInitiated: false`. Nothing on that path can raise a dialog on a runner any more.

Your two questions are answered in their threads: no seam wanted in `AppState.init`, and leave the `claimFirst` comment alone, since it already had a single caller on `main` and the stale plural is not yours to pay off.

Dropping the flag rather than re-arming it was the better call, and it was a third option I had not put on the table.

## One thing to know before you rebase #605

#605 adds `startWatching()`, which calls a bare `toggleWatching()`. After this lands, that picks up the new default `userInitiated: true`, so `POST /v1/watch` with `action: start` would raise the Accessibility alert on a machine nobody is sitting at, which is the exact failure this commit removes. `stopWatching()` is fine, its call takes the stop branch and never reaches the gate.

Nothing to do here. It is a one-line change plus a test in #605, which needs a rebase anyway.

I will take it from here, including the rebase and folding the commits. Thanks for the careful second round, particularly for checking against the pinned SwiftFormat rather than the one your machine had.

pasrom (approved):

PR #627, opened 2026-08-15
fix(app): observe output-folder changes in settings

### LINKED ISSUE #626: [Bug] Settings shows the previous output folder until the app is relaunched
### App version

0.7.0 (Homebrew Cask, `pasrom/meeting-transcriber`)

### Build variant

Homebrew Cask

### macOS version

27.0 (build 26A5406e)

### What happened?

Picking a new output folder in Settings persists correctly, but the UI keeps displaying the **previous** path. The label only catches up after the app is relaunched.

The stored preference is already correct at that point, so this looks display-only rather than a lost setting. Decoding the persisted bookmark right after the change shows the new folder:

```
$ defaults export app.meetingtranscriber - | ...
customOutputDirBookmark → file:///Users/<me>/<newly-picked-folder>   # path redacted
```

…while the settings UI still showed the old location.

Guess at the cause: the displayed path is derived once from `customOutputDirBookmark` (resolved into a cached string/`@State` at launch), and the folder picker writes the new bookmark without invalidating that derived value. Relaunching re-resolves it, which matches the observed behaviour.

### Steps to reproduce

1. Open Settings → Output and note the current output folder.
2. Click the folder picker and choose a different directory.
3. Confirm the picker.
4. The path shown in Settings is still the old one. ⇒ expected: the newly chosen path.
5. Quit and relaunch the app, open Settings → Output again.
6. The new path is now displayed correctly.

### Additional notes

I only verified the display and the persisted bookmark — I have not confirmed whether the **pipeline** also uses the stale path until relaunch, or whether only the label lags. If the running pipeline reads the same cached value, protocols written before the first relaunch would land in the old folder, which would make this more than cosmetic. Worth checking on your side.

### Diagnostic log

_Not attached — the issue is reproducible purely from the Settings UI and the persisted preference above._

danielsogl:
Fixes #626.

## What

Choosing a new output folder in Settings persisted correctly, but the label kept showing the previous path until the app was relaunched.

## Why

`AppSettings` is `@Observable`, and every other setting in it is a stored property that writes through in `didSet`:

```swift
var openAIModel: String {
    didSet { defaults.set(openAIModel, forKey: "openAIModel") }
}
```

`customOutputDirBookmark` was the one exception — a computed passthrough straight to `UserDefaults`:

```swift
var customOutputDirBookmark: Data? {
    get { defaults.data(forKey: "customOutputDirBookmark") }
    set { defaults.set(newValue, forKey: "customOutputDirBookmark") }
}
```

With no stored property behind it, `@Observable` has nothing to register access or mutation against, so `OutputSettingsView` — whose label derives from `settings.effectiveOutputDir` → `customOutputDir` → `customOutputDirBookmark` — was never invalidated. A relaunch re-evaluates `body` from scratch and picks up the new value, which matches the reported behaviour exactly.

Only the display was affected: I confirmed on my own install that the pipeline wrote recordings and protocols to the newly picked directory while Settings still showed the old one.

## How

- Back the bookmark with a stored property persisted via `didSet`, seeded from `defaults` in `init` — the pattern the rest of the file already uses.
- Guard the stale-bookmark repair inside `customOutputDir` with a conditional reassignment. That write happens in a *getter*; harmless while it only touched `UserDefaults`, but now that the property is observed, an unconditional write would invalidate a view during its own `body` evaluation. Same reasoning as the existing note on `asymmetricSilenceWarningSeconds`.

## Tests

Three added to `AppSettingsTests`:

- `testCustomOutputDirBookmarkNotifiesObservers` — `withObservationTracking` over `effectiveOutputDir` must fire when the bookmark changes.
- `testClearCustomOutputDirNotifiesObservers` — same for the Reset button path.
- `testCustomOutputDirBookmarkPersistsAcrossInstances` — guards the other half of the move: the bookmark must still round-trip through `UserDefaults` across launches, and clearing must remove the key.

The observation tests use a plain (non-security-scoped) bookmark: creating a scoped one needs an entitlement the test binary doesn't carry, and the behaviour under test is independent of the bookmark flavour. This mirrors what `RPCSettingsStateTests` already does.

Verified the tests actually gate the fix — with the `AppSettings` change stashed, both observation tests fail while the persistence test stays green.

## Verification

- `swift test --filter AppSettingsTests` → 56/56 pass.
- Full `swift test`: 17 failures, but the same 17 occur on an unmodified `main` in this environment (`SettingsViewTests`, `GeneralSettings*`, `MouseInjectionTests`, `…WithoutAnApplication`, etc. — UI tests that need an app host). Diff of failing-test names between `main` and this branch is empty.
- I could not run `./scripts/lint.sh` — no SwiftLint/SwiftFormat locally. Longest added line is 91 chars (limit 160) and the trailing-comma style is preserved, but a CI lint pass is the real check.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppSettings+OutputDirectory.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings+OutputDirectory.swift
@@ -1,0 +1,67 @@
+import Foundation
+
+/// Everything derived from `customOutputDirBookmark`. The bookmark itself has to
+/// stay a stored property on the class so `@Observable` can track it; the rest
+/// lives here to keep `AppSettings.swift` under its `file_length` budget.
+extension AppSettings {
+    /// Resolved URL from the security-scoped bookmark, or nil when none is set
+    /// or the bookmark no longer resolves. Read-only: security-scoped *access*
+    /// is the caller's job — every call site does its own paired
+    /// `startAccessingSecurityScopedResource()` / `stopAccessing…`.
+    ///
+    /// A stale bookmark still resolves, so this deliberately does not repair it.
+    /// `body` reads this through `effectiveOutputDir`, and repairing here would
+    /// write to observed state from inside a view update. `repairStaleCustomOutputDirBookmark()`
+    /// does that once at launch instead.
+    var customOutputDir: URL? {
+        var isStale = false
+        return resolveCustomOutputDir(isStale: &isStale)
+    }
+
+    /// The effective output directory: custom choice or ~/Downloads/MeetingTranscriber/.
+    var effectiveOutputDir: URL {
+        customOutputDir ?? AppPaths.downloadsProtocolsDir
+    }
+
+    /// Store a user-selected directory as a security-scoped bookmark.
+    func setCustomOutputDir(_ url: URL) {
+        guard let data = makeBookmark(for: url) else { return }
+        customOutputDirBookmark = data
+    }
+
+    /// Clear the custom output directory, reverting to the default.
+    func clearCustomOutputDir() {
+        customOutputDirBookmark = nil
+    }
+
+    /// Re-create the bookmark when macOS reports it stale (the folder moved or
+    /// was renamed). Call once at launch, off the view-update path — see the note
+    /// on `customOutputDir`. No-op when no bookmark is set or it still resolves.
+    func repairStaleCustomOutputDirBookmark() {
+        var isStale = false
+        guard let url = resolveCustomOutputDir(isStale: &isStale), isStale,
+              let refreshed = makeBookmark(for: url)
+        else { return }
+        customOutputDirBookmark = refreshed
+    }
+
+    // MARK: - Helpers
+
+    private func resolveCustomOutputDir(isStale: inout Bool) -> URL? {
+        guard let data = customOutputDirBookmark else { return nil }
+        return try? URL(
+            resolvingBookmarkData: data,
+            options: .withSecurityScope,
+            relativeTo: nil,
+            bookmarkDataIsStale: &isStale,
+        )
+    }
+
+    private func makeBookmark(for url: URL) -> Data? {
+        try? url.bookmarkData(
+            options: .withSecurityScope,
+            includingResourceValuesForKeys: nil,
+            relativeTo: nil,
+        )
+    }
+}
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -409,53 +409,10 @@ final class AppSettings {
 
     // MARK: - Output Directory
 
-    /// Security-scoped bookmark for a user-chosen output directory.
+    /// Security-scoped output-dir bookmark. Stored so `@Observable` can track it.
+    /// Everything derived from it lives in `AppSettings+OutputDirectory.swift`.
     var customOutputDirBookmark: Data? {
-        get { defaults.data(forKey: "customOutputDirBookmark") }
-        set { defaults.set(newValue, forKey: "customOutputDirBookmark") }
-    }
-
-    /// Resolved URL from the security-scoped bookmark. Calls `startAccessingSecurityScopedResource()`.
-    var customOutputDir: URL? {
-        guard let data = customOutputDirBookmark else { return nil }
-        var isStale = false
-        guard let url = try? URL(
-            resolvingBookmarkData: data,
-            options: .withSecurityScope,
-            relativeTo: nil,
-            bookmarkDataIsStale: &isStale,
-        ) else { return nil }
-        if isStale {
-            // Re-create bookmark from resolved URL
-            if let newData = try? url.bookmarkData(
-                options: .withSecurityScope,
-                includingResourceValuesForKeys: nil,
-                relativeTo: nil,
-            ) {
-                customOutputDirBookmark = newData
-            }
-        }
-        return url
-    }
-
-    /// Store a user-selected directory as a security-scoped bookmark.
-    func setCustomOutputDir(_ url: URL) {
-        guard let data = try? url.bookmarkData(
-            options: .withSecurityScope,
-            includingResourceValuesForKeys: nil,
-            relativeTo: nil,
-        ) else { return }
-        customOutputDirBookmark = data
-    }
-
-    /// Clear the custom output directory, reverting to the default.
-    func clearCustomOutputDir() {
-        customOutputDirBookmark = nil
-    }
-
-    /// The effective output directory: custom choice or ~/Downloads/MeetingTranscriber/.
-    var effectiveOutputDir: URL {
-        customOutputDir ?? AppPaths.downloadsProtocolsDir
+        didSet { defaults.set(customOutputDirBookmark, forKey: "customOutputDirBookmark") }
     }
 
     // MARK: - Diagnostics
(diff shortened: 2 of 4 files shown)


danielsogl:
Force-pushed: the first push tripped `file_length` on `AppSettings.swift`. I trimmed my added comments rather than restructuring your file — it now sits at exactly 600 lines, i.e. right on the cap.

Heads-up for whoever touches it next: any further addition to `AppSettings.swift` will trip the rule again. The Output Directory block (`customOutputDir`, `setCustomOutputDir`, `clearCustomOutputDir`, `effectiveOutputDir` — the stored property has to stay in the class) is the obvious extraction candidate into an `AppSettings+OutputDirectory.swift`, mirroring the existing `AppSettings+RPC.swift`. I deliberately left that out of this PR to keep the bug fix reviewable — happy to add it here or in a separate PR if you'd prefer.

`./scripts/lint.sh` is clean locally now (0 violations, 0 files need formatting), `AppSettingsTests` 56/56.

pasrom:
Reviewed at `cd2c0a19`. The fix is right and the mechanism matches what the rest of the file already does. One thing I'd like changed before merge; everything else is optional follow-up.

**The new guard cannot fire, and its comment says otherwise**

```swift
if isStale {
    // Re-create bookmark from resolved URL. Conditional: the property is
    // observed, so an unconditional write here would invalidate a view.
    if let newData = try? url.bookmarkData(
        options: .withSecurityScope,
        includingResourceValuesForKeys: nil,
        relativeTo: nil,
    ), newData != customOutputDirBookmark {
        customOutputDirBookmark = newData
    }
}
```

The block is only entered when `isStale`, and in that case the freshly created bookmark necessarily differs from the stored one: if the bytes were equal, the bookmark wouldn't be stale. So the condition is always true exactly where it was meant to protect, the write happens anyway, and it happens inside a getter that `body` reads through `effectiveOutputDir` (`OutputSettingsView`, `GeneralSettingsView.swift:162`).

In practice this costs one extra render pass in the rare stale case and converges as soon as the new bookmark resolves cleanly, so it isn't a correctness bug. What I'd rather not leave in the file is the comment, which states a protection that isn't there. A getter that mutates observed state is also the kind of construct that bites once someone adds a second reader.

Cleanest fix: move the repair out of the getter, into `setCustomOutputDir` or a one-time pass at launch. Then both the condition and its comment disappear. If you'd rather keep the diff small, correcting the comment to describe what the code actually does is enough for me.

**Optional follow-ups** (most predate this PR, mentioned because you're in this file anyway):

- The doc comment on `customOutputDir` still claims it calls `startAccessingSecurityScopedResource()`. It doesn't; every caller does its own paired start/stop.
- `openAIAPIKey` is still a computed passthrough with no stored backing, bound via `SecureField` in the same view, so it has exactly the invalidation gap this PR just closed for the bookmark.
- `clearCustomOutputDir()` assigns nil unconditionally, so a no-op clear still writes and still invalidates. `asymmetricSilenceWarningSeconds` guards against that with a conditional reassignment and a comment explaining why.
- The `try?` on bookmark creation swallows the error silently; `verboseDiagnostics` is the existing gate for that kind of logging.
- Tests: nothing constructs a stale bookmark, so the one new branch is untested. And both observation tests only assert that `onChange` fired, never that `effectiveOutputDir` changed. With a plain bookmark that `.withSecurityScope` may not resolve, they'd stay green under either behaviour; an `XCTAssertNotEqual` on the value before/after would pin the user-visible half.

+1 on the `AppSettings+OutputDirectory.swift` extraction you suggested. At exactly 600 lines the next addition trips `file_length` again, and moving the Output Directory block would carry the getter fix along with it.

pasrom (changes requested):
Requesting changes on one point only, detailed in https://github.com/pasrom/meeting-transcriber/pull/627#issuecomment-5301825879.

The `newData != customOutputDirBookmark` guard cannot fire in the stale case it was added for: inside `if isStale` the freshly created bookmark necessarily differs from the stored one, so the write happens anyway, from a getter that `body` reads through `effectiveOutputDir`. The comment above it states a protection that isn't there.

Either move the repair out of the getter (into `setCustomOutputDir`, or a one-time pass at launch), or correct the comment to describe what the code actually does. Either resolves this.

The fix itself is right and the mechanism matches the rest of the file. Everything else in that comment is optional follow-up and not gating.

danielsogl:
Rebased onto `main` (which now carries #628) and addressed the review.

**The gating point — you were right, and the guard is gone entirely.**

Inside `if isStale` a freshly created bookmark necessarily differs from the stored one, so `newData != customOutputDirBookmark` was always true exactly where it claimed to protect. Rather than correct the comment, I took the option you preferred: the repair is out of the getter.

- `customOutputDir` is now a pure resolve — no writes on the path `body` reads.
- `repairStaleCustomOutputDirBookmark()` does the re-creation, called once from `AppState.makeDefaultSettings()`, the single production construction site.

I also took the `AppSettings+OutputDirectory.swift` extraction, since you +1'd it and it carries the getter fix along as you said. The stored property stays on the class (`@Observable` needs it there); everything derived from it moved. `AppSettings.swift` is down from 600 to 557, so the cap has headroom again.

**On the tests — your last bullet was exactly right, and worse than you put it.**

Both observation tests used a plain `dir.bookmarkData()`. I checked: a plain bookmark does not resolve under `.withSecurityScope`, so `customOutputDir` was nil and `effectiveOutputDir` sat on the Downloads fallback the whole time. Asserting the value would have failed even with the fix in place — the tests were pinning nothing but the notification.

They now go through `setCustomOutputDir`, the real API (security-scoped creation and resolution both work fine in a non-sandboxed test binary — I verified that separately before relying on it), and assert the derived value the UI renders. Added:

- `testReadingEffectiveOutputDirDoesNotMutateTheBookmark` — pins the property your review is really about: reading the derived value must not write observed state. It fails loudly if the repair ever migrates back into the getter.
- `testRepairLeavesAResolvableBookmarkAlone` — the no-op half of the repair. The stale branch itself needs a folder moved out from under a live bookmark, which I could not reproduce in-process; it stays covered by the launch call site only, and the doc comment says so rather than implying coverage that isn't there.
- Value assertions on the notify/clear/persist tests.

Re-verified the tests still gate the fix: reverting the stored property to the old computed passthrough fails both notification tests.

**Not addressed, deliberately** — the `openAIAPIKey` invalidation gap, the unconditional write in `clearCustomOutputDir()`, and the silent `try?` on bookmark creation. All three predate this PR and none are on the path #626 reports; happy to take any of them as follow-ups, but they'd widen a display-bug fix. The `customOutputDir` doc comment that wrongly claimed it calls `startAccessingSecurityScopedResource()` is fixed, since I was rewriting it anyway.

Verification: `./scripts/lint.sh` clean (447 files, 0 violations, 0 need formatting). `AppSettingsTests` 58/58. `swift test --parallel --skip ModelPreloadTests` leaves the same 15 failures as unmodified `main` on this machine (macOS 27 SwiftUI view-introspection, unrelated) — diff of failing-test names against the baseline is empty.

pasrom (approved):
Verified at `a83d174a`. The repair is out of the getter and `customOutputDir` is a pure resolve; `repairStaleCustomOutputDirBookmark()` has exactly one production call site in `AppState.makeDefaultSettings()`. The dead condition and its wrong comment are gone rather than patched, which is the outcome I wanted.

Checked on my machine: SwiftLint `--strict` clean on all three files, `AppSettingsTests` 58/58, `AppSettings.swift` down to 557 lines.

Good catch on the tests being weaker than I described: a plain bookmark not resolving under `.withSecurityScope` meant they were pinning the notification and nothing else. Going through the real API and asserting the derived value is the right fix, and documenting the stale branch as covered by the launch call site only, instead of implying coverage that isn't there, is the honest way to leave it.

Fine to take the three untouched items as follow-ups.

Issue #531, opened 2026-07-26
Getting this warning related to macos version name


Thrilok28021996:
Warning: Calling string comparison format for `depends_on macos:` is deprecated! Use `depends_on macos: :sonoma` instead

pasrom:
Thanks for the report. Good news first: this is a cosmetic warning only, the install still succeeds (it's a `Warning:`, not an `Error:`, and brew exits 0).

We're deliberately keeping `depends_on macos: ">= :sonoma"` for now. Homebrew 5.1.11 (2026-05-08) both added this deprecation and, in the same commit, changed a bare `:sonoma` symbol to mean "Sonoma or newer". On older Homebrew a bare `:sonoma` still means *exactly* Sonoma, which would block installs on Sequoia and Tahoe. The current `">= :sonoma"` form is evaluated as a proper minimum on both old and new Homebrew, so it keeps working for everyone who hasn't updated brew yet.

We'll move to the bare `:sonoma` form once older Homebrew installs have aged out. Until then the warning is harmless and can be safely ignored.

Issue #441, opened 2026-06-17
Fix deprecated depends_on macos: string format in casks


BigCactusLabs:
Installing via the tap on Homebrew 6.0.0 throws a deprecation warning for both casks:

```
Warning: Calling string comparison format for `depends_on macos:` is deprecated! Use `depends_on macos: :sonoma` instead.
  /opt/homebrew/Library/Taps/pasrom/homebrew-meeting-transcriber/Casks/meeting-transcriber.rb:10
  /opt/homebrew/Library/Taps/pasrom/homebrew-meeting-transcriber/Casks/meeting-transcriber@beta.rb:10
```

Line 10 of both `Casks/meeting-transcriber.rb` and `Casks/meeting-transcriber@beta.rb` uses the deprecated **numeric** string-comparison form (e.g. `>= 14.2`). The fix is to use the macOS **symbol** instead — but keep the `>=` operator:

```ruby
depends_on macos: ">= :sonoma"
```

**Don't use the bare `:sonoma` that the warning text literally suggests.** The warning abbreviates the recommendation, but a bare symbol is risky:

- Per the [Cask Cookbook](https://docs.brew.sh/Cask-Cookbook), a bare symbol denotes a specific release, and `">= :symbol"` is the documented (non-deprecated) form for a *minimum* version.
- Real-world reports show a bare symbol behaving as exact/near-exact match and **blocking installs on newer macOS** (see [farion1231/cc-switch#3957](https://github.com/farion1231/cc-switch/issues/3957), where `:monterey` blocked macOS 13+).
- Concretely: this report came from **macOS 26 (Tahoe)** — a bare `:sonoma` could block exactly that machine. `">= :sonoma"` keeps "Sonoma or newer," which is what you want.

**Caveat on the 14.2 requirement:** `depends_on macos:` only supports *major* releases (Sonoma = 14.x); minor versions like 14.2 cannot be expressed in any form. So the cask can gate at "14.x or newer" at best — it cannot enforce the 14.2 floor that CATapDescription audio capture needs. If precise 14.2 enforcement matters, that check belongs in the app at runtime (refuse to launch / degrade gracefully on < 14.2); the cask can't do it.

Worth a quick local verification that `">= :sonoma"` clears the warning on Homebrew 6.0 before tagging.

**Environment:** Homebrew 6.0.0, macOS 26 (Tahoe), arm64.

[...]

PR #631, opened 2026-08-15
fix(build): drop the deprecated string form of the cask macOS floor

### LINKED ISSUE #630: Cask uses deprecated `depends_on macos: ">= :sonoma"` string form
Every `brew` command that touches the cask prints a deprecation warning telling the user to report it to the tap:

```
Warning: Calling string comparison format for `depends_on macos:` is deprecated! Use `depends_on macos: :sonoma` instead.
Please report this issue to the pasrom/homebrew-meeting-transcriber tap (not Homebrew/* repositories), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/pasrom/homebrew-meeting-transcriber/Casks/meeting-transcriber.rb:10
```

Both casks in this repo carry the deprecated form:

- `Casks/meeting-transcriber.rb:10`
- `Casks/meeting-transcriber@beta.rb:10`

The floor does not change. A cask's `depends_on macos:` is always parsed with the `>=` comparator (`Cask::DSL::DependsOn#macos=` calls `MacOSRequirement.parse(args, comparator: ">=")`), so the bare symbol means the same thing — verified locally, both forms resolve to `>= 14`, only the string form warns.

Note this repo is not the whole fix: `release.yml` seeds a cask into the tap repo only when the tap does not already carry it, and afterwards only rewrites `version` and `sha256`. The live warning therefore persists until `pasrom/homebrew-meeting-transcriber` gets the same one-line change; fixing it here keeps the source of truth right and covers any future re-seed.

danielsogl:
Closes #630.

`depends_on macos: ">= :sonoma"` is deprecated in Homebrew, so every `brew install` / `upgrade` / `info` touching the cask prints a warning asking the user to report it to the tap.

Homebrew's own replacement for the `>=` string form is the bare symbol. The floor is unchanged: a cask's `depends_on macos:` always parses with the `>=` comparator, so both forms resolve to `>= 14` — checked locally against the installed Homebrew, where only the string form warns.

Not the whole fix on its own: `release.yml` seeds a cask into `pasrom/homebrew-meeting-transcriber` only when the tap does not already carry it, and thereafter rewrites just `version` and `sha256`. The warning users see comes from the tap's copy and needs the same one-line change there; this PR fixes the source of truth and any future re-seed.

### DIFF
--- a/Casks/meeting-transcriber.rb
+++ b/Casks/meeting-transcriber.rb
@@ -7,7 +7,7 @@ cask "meeting-transcriber" do
   desc "Auto-transcribe and summarize meetings"
   homepage "https://github.com/pasrom/meeting-transcriber"
 
-  depends_on macos: ">= :sonoma"
+  depends_on macos: :sonoma
 
   conflicts_with cask: "meeting-transcriber@beta"
 
--- a/Casks/meeting-transcriber@beta.rb
+++ b/Casks/meeting-transcriber@beta.rb
@@ -7,7 +7,7 @@ cask "meeting-transcriber@beta" do
   desc "Auto-transcribe and summarize meetings (pre-release)"
   homepage "https://github.com/pasrom/meeting-transcriber"
 
-  depends_on macos: ">= :sonoma"
+  depends_on macos: :sonoma
 
   conflicts_with cask: "meeting-transcriber"
 


pasrom:
Thanks for the PR, and the mechanics in your description are accurate: the deprecation is real, on a current Homebrew both forms resolve to `>= 14`, and you correctly spotted that the tap copy is what users actually hit.

We are still keeping the string form, for a reason that does not show up on a single machine. Homebrew 5.1.11 (2026-05-08, commit `0e87cc8941`) introduced this deprecation **and** changed the bare symbol's comparator in the same commit. Before it, `cask/dsl/depends_on.rb` built a bare symbol as `MacOSRequirement.new([args.first], comparator: "==")`; after it, everything goes through `MacOSRequirement.parse(args, comparator: ">=")`. Homebrew added its own "when a symbol is used / creates a minimum MacOSRequirement" test in that same commit.

Verified by running the old code rather than only reading it, via a checkout of tag 5.1.10:

| brew | `:sonoma` | `">= :sonoma"` |
|---|---|---|
| 5.1.10 | `==` 14, Sequoia and Tahoe not satisfied | `>=` 14, satisfied |
| 6.0.17 | `>=` 14, satisfied | `>=` 14, satisfied, plus the warning |

So on a pre-5.1.11 Homebrew this change would block installs on Sequoia and Tahoe, while the current string form parses as a minimum on every release. 30-day analytics today: 8.70% of clients are below 5.1.11 (1,848,406 of 21,243,063 events), down from 9.30% two weeks ago.

One correction to our own earlier answers, since it turned up while checking yours: in #531 I called the warning harmless. That is not true for everyone. With `HOMEBREW_DEVELOPER=1`, Homebrew turns the deprecation into a hard error, and `brew info --cask meeting-transcriber` exits 1 against the live tap. So this is a trade between two breakages, not between noise and a breakage. We are keeping the current form because the failure it causes is legible and self-inflicted: it names the file and line, and the user can unset the variable. The other direction produces "requires macOS Sonoma" on a machine running something newer, which the user cannot act on.

That balance flips as old Homebrew ages out, so this is deferred rather than declined, with a revisit around October. Leaving #630 open as the tracking issue for it. See also #441, #531, and #570, an identical change I opened myself and closed again.

PR #642, opened 2026-08-21
feat(app): add transcript output privacy options

niftan5:
## Privacy rationale

The transcript is required internally for diarization and protocol generation,
but users may not want to retain the raw conversation after a meeting. The new
output options allow users to omit the automatic full-transcript appendix and
the separately stored raw transcript while retaining the generated meeting
minutes.

The options control local output retention and automatic attachment. The
transcript is still provided to the selected protocol generator; when that
provider is external, its privacy terms apply.

## Retention behaviour

- Both options default to enabled, preserving existing behaviour.
- Raw text artifacts (`.txt` and transcript-bearing segment sidecars) are
  removed only after a protocol was saved successfully.
- Recordings follow their existing retention policy and are never deleted by
  these options.
- When protocol generation is disabled, fails, or a job is interrupted, the
  draft transcript is retained to avoid data loss and can be removed manually.

## Validation

- SwiftFormat and SwiftLint: passed
- Swift tests: settings, workflow, recovery, snapshot, and full suite passed
- Homebrew and App Store release builds: passed

### DIFF
--- a/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
@@ -130,10 +130,14 @@ extension PipelineQueue {
     /// Best effort on purpose: a job that cannot write here will report the real
     /// problem when stage 3 tries again and does not swallow it.
     private func saveTranscriptDraft(_ transcript: String, ctx: JobContext, outputDir: URL) {
-        _ = try? ProtocolGenerator.saveTranscript(
+        guard let path = try? ProtocolGenerator.saveTranscript(
             transcript, basename: ctx.slug,
             dir: outputDir.appendingPathComponent("protocols"),
-        )
+        ) else { return }
+        if let idx = jobs.firstIndex(where: { $0.id == ctx.jobID }) {
+            jobs[idx].transcriptPath = path
+            jobs[idx].namingSlug = ctx.slug
+        }
     }
 
     /// Thin orchestrator: take the next waiting job and run it through the
@@ -659,7 +663,10 @@ extension PipelineQueue {
         finalTranscript: String, transcription: TranscriptionOutput,
         ctx: JobContext, workDir: URL, outputDir: URL,
     ) async throws {
-        // --- Save Transcript & Audio (always) ---
+        // --- Save Transcript & Audio ---
+        // Keep the transcript until the terminal state, even when the user
+        // opted out of a separate raw file: late speaker naming still needs to
+        // rewrite it before generating the final protocol.
         let protocolsDir = outputDir.appendingPathComponent("protocols")
         let txtPath = try ProtocolGenerator.saveTranscript(finalTranscript, basename: ctx.slug, dir: protocolsDir)
         logger.info("[\(ctx.shortID, privacy: .public)] transcript_saved file=\(txtPath.lastPathComponent, privacy: .private)")
@@ -762,9 +769,11 @@ extension PipelineQueue {
                 diarized: diarized,
                 meetingStartTime: meetingStartTime,
             )
-            let fullMD = protocolMD + "\n\n---\n\n## Full Transcript\n\n" + transcript
+            let markdown = includeFullTranscriptInProtocol
+                ? protocolMD + "\n\n---\n\n## Full Transcript\n\n" + transcript
+                : protocolMD
             let mdPath = try ProtocolGenerator.saveProtocol(
-                fullMD, basename: basename, dir: protocolsDir,
+                markdown, basename: basename, dir: protocolsDir,
             )
             logger.info("[\(shortID, privacy: .public)] protocol_saved file=\(mdPath.lastPathComponent, privacy: .private)")
             if let idx = jobs.firstIndex(where: { $0.id == jobID }) {
--- a/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
@@ -769,7 +769,7 @@ extension PipelineQueue {
                 diarized: diarized,
                 meetingStartTime: meetingStartTime,
             )
-            let markdown = includeFullTranscriptInProtocol
+            let markdown = transcriptOutputOptions(forJobID: jobID).includeFullTranscriptInProtocol
                 ? protocolMD + "\n\n---\n\n## Full Transcript\n\n" + transcript
                 : protocolMD
             let mdPath = try ProtocolGenerator.saveProtocol(
--- a/app/MeetingTranscriber/Sources/PipelineQueue.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue.swift
@@ -26,7 +26,7 @@ struct DiarizationRun {
 
 @MainActor
 @Observable
-// swiftlint:disable:next attributes
+// swiftlint:disable:next attributes type_body_length
 class PipelineQueue {
     /// Internal setter (not `private(set)`) because the stage and recovery
     /// extension methods in sibling files (PipelineQueue+Stages.swift,
--- a/app/MeetingTranscriber/Sources/PipelineQueue.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue.swift
@@ -535,7 +535,8 @@ class PipelineQueue {
         guard oldState != newState || error != nil else { return }
         jobs[index].state = newState
         if let error { jobs[index].error = error }
-        if (newState == .done || newState == .error), !transcriptOutputOptions(forJobID: id).saveRawTranscriptSeparately {
+        if newState == .done || newState == .error,
+           !transcriptOutputOptions(forJobID: id).saveRawTranscriptSeparately {
             removeRawTranscriptArtifacts(for: index)
         }
         recordStageTransition(from: oldState, to: newState, jobID: id)
--- a/app/MeetingTranscriber/Sources/PipelineQueue.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue.swift
@@ -59,6 +59,12 @@ class PipelineQueue {
     let echoDedupEnabled: Bool
     let numSpeakers: Int
     let micLabel: String
+    /// Captured when the queue is built so a running job has stable output
+    /// semantics even if the user changes Settings mid-processing.
+    let includeFullTranscriptInProtocol: Bool
+    /// See `includeFullTranscriptInProtocol`. When false, the temporary raw
+    /// transcript is removed as soon as the job reaches a terminal state.
+    let saveRawTranscriptSeparately: Bool
     let speakerMatcherFactory: () -> SpeakerMatcher
     let vadConfig: VADConfig?
     /// nil disables JSONL logging. AppState injects a real instance for production;
@@ -234,6 +240,8 @@ class PipelineQueue {
         echoDedupEnabled = true
         self.numSpeakers = 0
         self.micLabel = "Me"
+        self.includeFullTranscriptInProtocol = true
+        self.saveRawTranscriptSeparately = true
         self.speakerMatcherFactory = speakerMatcherFactory
         self.snapshotWriter = snapshotWriter
         self.vadConfig = nil
@@ -310,6 +318,8 @@ class PipelineQueue {
         echoDedupEnabled: Bool = true,
         numSpeakers: Int = 0,
         micLabel: String = "Me",
+        includeFullTranscriptInProtocol: Bool = true,
+        saveRawTranscriptSeparately: Bool = true,
         speakerMatcherFactory: @escaping () -> SpeakerMatcher = PipelineQueue.throwawayMatcherFactory(),
         snapshotWriter: @escaping @Sendable ([PipelineJob], URL) throws -> Void = PipelineSnapshot.save,
         vadConfig: VADConfig? = nil,
@@ -341,6 +351,8 @@ class PipelineQueue {
         // of micLabel for both the tagging and the re-split, so sanitizing here
         // keeps them consistent.)
         self.micLabel = micLabel == DiarizationProcess.remoteSpeakerLabel ? "Me" : micLabel
+        self.includeFullTranscriptInProtocol = includeFullTranscriptInProtocol
+        self.saveRawTranscriptSeparately = saveRawTranscriptSeparately
         self.speakerMatcherFactory = speakerMatcherFactory
         self.snapshotWriter = snapshotWriter
         self.vadConfig = vadConfig
@@ -483,6 +495,9 @@ class PipelineQueue {
         guard oldState != newState || error != nil else { return }
         jobs[index].state = newState
         if let error { jobs[index].error = error }
+        if (newState == .done || newState == .error), !saveRawTranscriptSeparately {
+            removeRawTranscript(for: index)
+        }
         recordStageTransition(from: oldState, to: newState, jobID: id)
         eventLog.append(jobID: id, event: "state_change", from: oldState, to: newState)
         saveSnapshot()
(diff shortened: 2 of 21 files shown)


app/MeetingTranscriber/Sources/SpeakerNamingStore.swift
```diff
@@ -81,6 +81,14 @@ struct SpeakerNamingStore {
         try? FileManager.default.removeItem(at: recordingsDir.appendingPathComponent("\(slug)_naming.json"))
     }
 
+    /// Delete only the cached transcript segments. These contain verbatim
+    /// speech, unlike the audio/naming sidecars, and must not outlive a job
+    /// when separate raw-transcript output is disabled.
+    func deleteTranscriptSegments(slug: String?) {
+        guard let slug, let recordingsDir else { return }
+        try? FileManager.default.removeItem(at: recordingsDir.appendingPathComponent("\(slug)_segments.json"))
```
pasrom: `try?` swallows every error here, while the transcript deletion a few lines away logs a warning and records one on the job. For a cleanup whose whole purpose is that a file stops existing, a silent failure is the one outcome the user must not be left with: they would believe the verbatim segments are gone when they are still on disk. Worth reporting the failure the same way the `.txt` path does.

app/MeetingTranscriber/Sources/PipelineQueue.swift
```diff
@@ -368,14 +397,37 @@ class PipelineQueue {
...
         eventLog.append(jobID: job.id, event: "enqueued", from: nil, to: job.state)
         saveSnapshot()
         logger.info("Enqueued job: \(job.meetingTitle, privacy: .private) (\(job.id))")
         triggerProcessing()
     }
 
+    /// Resolves the output policy for an existing job. The fallback keeps
+    /// snapshots from versions before per-job settings compatible.
+    /// Internal because pipeline stages in sibling files use it.
+    func transcriptOutputOptions(forJobID jobID: UUID) -> TranscriptOutputOptions {
```
pasrom: This fallback is correct for jobs that predate the new fields, but there is a live path that also lands here and should not: orphan recovery builds its job and appends it to `jobs` directly instead of going through `enqueue`, so neither option is ever stamped, both persist as nil, and a later restart can resolve the same job through a different fallback value than the one that was in force when it was picked up. Stamping the options on that path would make the per-job guarantee hold everywhere.

app/MeetingTranscriber/Sources/PipelineQueue.swift
```diff
@@ -476,6 +528,10 @@ class PipelineQueue {
         guard oldState != newState || error != nil else { return }
         jobs[index].state = newState
         if let error { jobs[index].error = error }
+        if newState == .done || newState == .error,
+           !transcriptOutputOptions(forJobID: id).saveRawTranscriptSeparately {
```
pasrom: This is the blocking one. The condition asks whether the job reached a terminal state and whether the user wants the raw transcript kept, but never whether a protocol was actually produced. Three paths therefore delete the only text that exists:

1. **Provider `.none`.** The generator factory returns nil for `.none`, so `generateProtocol` returns immediately without writing any `.md`, the job still reaches `.done`, and the transcript is deleted. This is not an edge case: with `.none` selected, *every* job produces zero text output. For an imported file the output folder can then hold no text artifact at all, though the imported audio itself stays untouched in the user's own folder.
2. **Protocol generation failure.** The catch adds the warning "Protocol generation failed — transcript saved", the job then reaches `.done`, and this line deletes exactly that transcript. The warning becomes untrue. (That `addWarning` call is in `PipelineQueue+Stages.swift`, outside this diff's hunks, so I could not anchor a comment on it.)
3. **An `.error` reached after the draft was written.** `saveTranscriptDraft` runs only once a non-empty transcription exists, so early failures have nothing to lose. But a failure after that point, for instance in the final rewrite before protocol generation, removes the draft that is today the artifact which survives a failed run.

There is a further symptom that makes the first two invisible rather than merely silent: the completion notification takes its title from `protocolPath != nil`, and the `onJobStateChange` callback fires a few lines below this deletion, so it receives the already-mutated job. For the two `.done` cases above, the user is told **"Transcript Saved"** at the moment the transcript is removed. (An `.error` job sends its error notification instead, so this applies to `.none` and to generation failure, not to the third path.)

Smallest fix: add `jobs[index].protocolPath != nil` to this condition. That closes all three at one site, and it makes the notification title true again.

One consequence worth handling here rather than discovering later: with `.none` selected the guard means the transcript is now kept on every job, forever, while the toggle still reads as on. That is the right trade, since retention is recoverable and deletion is not, but it should not be silent. A job warning in the branch that declines to delete would cover `.none`, generation failure and the post-draft `.error` in one place. Gate that warning on `transcriptPath != nil` as well, otherwise it fires on early errors where no transcript was ever written and tells the user something false in the other direction.

app/MeetingTranscriber/Sources/PipelineQueue.swift
```diff
@@ -500,6 +556,31 @@ class PipelineQueue {
         terminalJobStore?.record(JobStatusDTO(job: job))
     }
 
+    /// Delete transcript-bearing output once no pipeline stage needs it. The
+    /// speaker-naming flow deliberately reaches `.done` only after it has read
+    /// and rewritten these files, so this also covers late confirmation.
+    private func removeRawTranscriptArtifacts(for index: Int) {
```
pasrom: This deletion performs no security-scoped access on the user's output directory. The repository's contract is explicit, in the doc comment on `AppSettings.customOutputDir`: security-scoped *access* is the caller's job, and every call site does its own paired `startAccessingSecurityScopedResource()` / `stopAccessing…`.

Worth pairing with the note in `WatchLoop+RecordOnly.swift`, which spells out the part that is easy to get wrong: access has to be started on the URL that resolved from the bookmark, not on a child path, because calling it on a child silently fails. So the fix here is to wrap the cleanup in access to the bookmark-resolved output root, not to the protocols subfolder.

In the sandboxed App Store build a user-selected output folder is reachable only through that bookmark, so this is where the cleanup is most likely not to work. I have not measured the sandboxed runtime, and unit tests do not run inside it, so treat that consequence as reasoned rather than observed. The missing scope itself is plain in the code.

app/MeetingTranscriber/Sources/SpeakerNamingStore.swift
pasrom: `try?` swallows every error here. For a cleanup whose whole purpose is that a file stops existing, a silent failure is the one outcome the user must not be left with: they would believe the verbatim segments are gone while they are still on disk.

I know the neighbouring functions in this store swallow errors the same way, so this is not a local inconsistency but a deliberate deviation I am asking for: the transcript removal in `PipelineQueue` already logs and records a job warning when it fails, and this deletion carries the same promise. It does mean threading a result or a thrown error back out through `SpeakerNamingSession.removeTranscriptSegments` to the queue, so it is small but not a one-liner.

app/MeetingTranscriber/Sources/PipelineQueue.swift
pasrom, line 420: This fallback is correct for jobs that predate the new fields, but there is a live path that also lands here and should not: orphan recovery builds its job and appends it to `jobs` directly instead of going through `enqueue`, so neither option is ever stamped, both persist as nil, and a later restart can resolve the same job through a different fallback value than the one that was in force when it was picked up. Stamping the options on that path would make the per-job guarantee hold everywhere.

app/MeetingTranscriber/Tests/WorkflowIntegrationTests.swift
```diff
@@ -398,6 +519,18 @@ final class WorkflowIntegrationTests: XCTestCase {
         XCTAssertFalse(h.queue.jobs.first?.warnings.isEmpty ?? true)
     }
 
+    func testWorkflowProtocolFailureCleansRawOutputWhenDisabled() async throws {
```
pasrom: This test pins the destructive outcome as intended behaviour: after a failed protocol run it asserts the transcript artifacts are gone. Under the guard above they must survive, so this assertion needs to be inverted rather than kept.

I am flagging it explicitly because the natural reading is the opposite: with this test in place, the obvious way to make things pass is to adjust the guard until the existing assertion holds, which would put the defect back. The behaviour this test describes is the thing to change.

app/MeetingTranscriber/Tests/WorkflowIntegrationTests.swift
```diff
@@ -152,6 +168,85 @@ final class WorkflowIntegrationTests: XCTestCase {
...
+                .contains { $0.hasSuffix(".txt") },
+        )
+        let recordingsDir = tmpDir.appendingPathComponent("recordings")
+        XCTAssertFalse(
+            try FileManager.default.contentsOfDirectory(atPath: recordingsDir.path)
+                .contains { $0.hasSuffix("_segments.json") },
+        )
+    }
+
+    func testWorkflowCapturesCurrentOutputOptionsWhenJobIsEnqueued() async throws {
```
pasrom, line 246: Good test, and the mutation of the provider's return value after construction is exactly the right idea. One gap: the settings are changed only *before* the job is enqueued, never after. So it would still pass if production read live settings during protocol generation instead of the values captured on the job, which is the mistake the per-job capture exists to prevent. Changing the options after `enqueue` and before processing would close that.

app/MeetingTranscriber/Tests/WorkflowIntegrationTests.swift
```diff
@@ -90,6 +96,16 @@ final class WorkflowIntegrationTests: XCTestCase {
         )
     }
 
+    private func assertNoRawTranscriptArtifacts(_ h: Harness) {
```
pasrom, line 100: This helper asserts only what must be **absent**. A bug that deleted too much, for example the 16 kHz sidecar or the recordings directory, would keep every test using it green. Since the whole feature is about deleting files, one assertion that the audio artifacts are still there after the cleanup would be worth more than any of the absence checks, and it is the test that would have caught an over-broad delete.

app/MeetingTranscriber/Sources/Settings/OutputSettingsView.swift
```diff
@@ -78,6 +78,9 @@ struct OutputSettingsView: View {
             .accessibilityIdentifier(A11yID.outputFolderSection)
 
             Section("Protocol Generation") {
+                Toggle("Include full transcript in protocol", isOn: $settings.includeFullTranscriptInProtocol)
```
pasrom, line 81: These two toggles have no accessibility identifiers and no test. The project's testing rules ask for one wiring test per control: find it, drive it, assert the write-back to `AppSettings`.

`SettingsViewTests` has the pattern to copy, for instance `testAppsToWatchTogglesBindToSettings` and `testDebugRPCToggleBindsToSetting`.

The wiring test is the part I would insist on. On the identifier, an honest caveat: several existing toggles are located by label predicate rather than by an `A11yID` constant, so if adding constants turns awkward here, following that established pattern is acceptable.

app/MeetingTranscriber/Sources/Settings/OutputSettingsView.swift
```diff
@@ -125,7 +128,7 @@ struct OutputSettingsView: View {
             openAIConfigView
 
         case .none:
-            Text("Only the raw transcript will be saved — no LLM summarization.")
+            Text("No LLM summarization will be generated.")
```
pasrom: Rewording this was right, the old text stopped being true once the raw transcript could be removed. One addition once the guard above is in: with `.none` selected no protocol is ever generated, so the raw transcript will always be kept regardless of the toggle. That is the safe behaviour, but a user who picks no LLM *and* no raw transcript is precisely the person who most wants the opposite, and right now nothing tells them. This caption is the natural place to say it.

app/MeetingTranscriber/Sources/SpeakerNamingSession.swift
```diff
@@ -391,6 +391,13 @@ final class SpeakerNamingSession {
         namingStore.cleanupSidecarFiles(slug: slug)
     }
 
+    /// Removes only the transcript-bearing segment sidecar. The queue uses this
+    /// when the user opts out of permanent raw transcript output; audio remains
```
pasrom, line 395: Thanks for scoping this deliberately and writing it down. The sentence is accurate and it is the right decision: the recording is governed by the recording-retention policy, and widening a delete path to audio would be a much riskier change than this PR should carry.

It is in the wrong place though. The PR description argues that users may not want to retain the raw conversation, and a reader will take that to include the recording. Please lift this sentence into the description, so the promise matches what the feature does: the text is removed, the recording stays.

app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
```diff
@@ -126,10 +126,14 @@ extension PipelineQueue {
     /// Best effort on purpose: a job that cannot write here will report the real
     /// problem when stage 3 tries again and does not swallow it.
     private func saveTranscriptDraft(_ transcript: String, ctx: JobContext, outputDir: URL) {
```
pasrom, line 132: Recording the draft path is a real improvement beyond the stated scope: before this, an errored job left a transcript on disk with no `transcriptPath` on the job at all.

It does open a gap that I am treating as a follow-up rather than a merge condition: cancelling a job mid-flight, or dismissing one that is waiting for speaker naming, removes the job without a terminal transition, so this draft stays on disk indefinitely even with the option off. The file behaviour predates your change, but the promise the setting now makes is what turns it into a broken one. A sentence in the setting's doc comment naming that limitation would be enough for now.

pasrom (changes requested):
Thanks for this, and sorry for the slow response. I have gone through it in detail, and the design underneath is good: capturing the options on the job at enqueue rather than reading live settings during processing is the right call, the optional Codable fields are a sensible migration, deletion runs before the terminal record is written so the automation API never advertises a path that is already gone, and the deterministic `<slug>.txt` filename means a draft and a final write cannot leave an orphan copy behind. The happy-path matrix is genuinely thorough: protocol-only, transcript-only, both, and the naming confirm and skip paths. Both build variants compile and lint is clean. Thanks also for fixing the caption assertion yourself.

One blocking defect, plus eight small conditions. It is worth stating the defect precisely, because the rest is close to mergeable.

**The deletion is keyed to the terminal state and never to whether a protocol actually exists.** That turns the privacy option into a total-content-loss switch on three paths, the worst being provider `.none`, where no protocol is ever produced by design, so every single job deletes the only text it created. Details inline.

The fix is small: gate the deletion on the job having a `protocolPath`. What makes that worth doing carefully is that it flips the failure direction. Today the feature can destroy the only copy; afterwards, every remaining gap is a file surviving that should have been removed, which is a privacy shortfall rather than data loss, and recoverable. That is the right way round.

Merge conditions, all of them small:

1. Gate the deletion on `protocolPath != nil`, and report the retention on the job instead of silently keeping the file.
2. Invert `testWorkflowProtocolFailureCleansRawOutputWhenDisabled`. It currently pins the destructive outcome as intended behaviour, so please make it assert that the transcript survives a failed protocol run.
3. Wrap the deletion in the output directory's security scope, and stop swallowing the segment-file error.
4. Route the restored-job transition in `PipelineQueue+Recovery.swift` (the direct `jobs[idx].state = .done` when naming data is missing) through `updateJobState`. Its own consequence is only a retention gap, which is the class I am deferring below, so to be explicit about why this one blocks: the objection is structural rather than about the outcome. A terminal transition that bypasses the central path also skips the ledger, the terminal record and the reaper, and every future hook hung on that path inherits the same hole. That file is not in this diff, so I could not comment on it inline.
5. Stamp the options on orphan-recovered jobs. That path appends to `jobs` directly instead of going through `enqueue`, so both fields stay nil and get persisted nil.
6. Accessibility identifiers and one wiring test per toggle.
7. Tests: an `.error` reached *after* the draft transcript was written and with raw output disabled (the existing error tests fail before any draft exists and run at default settings, so they do not cover this), the `.none` retention path, decoding a snapshot written before these fields existed, and one proving the recording survives the cleanup.
8. Say in the PR description that text artifacts are removed while the recording stays. You already documented that scoping in the code; it belongs where users will read it.

Not merge conditions, and I will file them separately: cancelling a job or dismissing a naming-pending one leaves the draft transcript behind, and a crash between the transcript write and protocol generation can strand it too. Both fail toward keeping data, so they are follow-ups rather than blockers. A sentence in the setting's doc comment naming that limitation would be honest, since it is the one place the promise is knowingly not kept.

pasrom (approved):
This is a thorough round, thank you. Every blocking point is addressed, and two of them better than I asked for.

The guard does the right thing and the reporting around it is careful: the retention warnings are gated on `transcriptPath != nil`, so they cannot claim a transcript was kept when none was ever written, and the `.error` case gets its own wording rather than being folded into the `.done` one. `stampTranscriptOutputOptions` shared between `enqueue` and orphan recovery is a better answer than the two call sites I asked for. And in the recovery rewrite you collected the job IDs first and transitioned afterwards, instead of mutating `jobs` while iterating it. I had not raised that; it would have been an easy way to introduce a new bug while fixing an old one.

The security scope is on the bookmark-resolved output root rather than on a child path, which is the part that silently fails if you get it wrong. All four requested tests are there and test the real thing: the legacy-snapshot test actually strips the two keys from the encoded JSON rather than just constructing a job without them, and the error test targets a failure after the draft was written, which is the only error path that could destroy anything.

One leftover, not worth holding this for: the PR description still says users may not want to retain the raw conversation, without the qualifier. You did document it properly in the README and the architecture notes ("The recording follows its own retention policy and is not deleted by either setting"), so anyone reading the shipped docs gets the accurate picture. Only the PR page is stale, and that becomes history once this merges.

Merging now.

PR #643, opened 2026-08-21
feat(app): pass meeting timestamps to protocol prompts

niftan5:
## Summary

Local LLMs do not reliably know the current date. This adds authoritative meeting-time context to protocol generation.

- adds `{MEETING_DATE}` (`YYYY-MM-DD`) and `{MEETING_TIME}` (`HH:mm`) prompt variables
- prepends an authoritative metadata block to every protocol request, including
  existing custom prompts
- forwards the captured meeting-start timestamp through both protocol providers
- documents the supported variables

## Verification

- Focused protocol-generator and OpenAI-provider tests passed (145 tests)
- `./scripts/lint.sh`
- `git diff --check upstream/main...HEAD`


app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
```diff
@@ -717,14 +717,20 @@ extension PipelineQueue {
         // path; the fallback just keeps generation working if the job is gone.
         let basename = jobs.first { $0.id == jobID }?.namingSlug
             ?? Self.namingSlug(title: title, jobID: jobID, startTime: Date())
+        let meetingStartTime = jobs.first { $0.id == jobID }.map { job in
+            job.meetingStartTime ?? job.enqueuedAt
```
pasrom: `meetingStartTime` is documented as nil for reimport and orphan-recovery jobs (`PipelineJob.swift:58-66`). Singleton file imports (`PipelineController.swift:273`) and crash recovery (`PipelineQueue+Recovery.swift:161`) both build their jobs without it, so this resolves to `enqueuedAt`, the moment the file was imported or recovered. `buildSystemPrompt` then presents that value as authoritative and anchors every relative time expression to it, which overrides any date the transcript itself states.

Importing a recording from a few weeks ago therefore yields a protocol dated today, with "next Friday" and "by end of month" resolved against the wrong anchor. The template line this PR replaces, `**Date:** [Date from context or today]`, used to leave that inference to the model.

The outer `?? Date()` on the next line has the same effect once the job has left `jobs`, which the existing comment above describes as the job-gone path.

Smallest fix I can see: keep the value optional and branch the wording. With a real start time, emit the block as it stands. With nil, either omit the block or label it as the processing date and tell the model to prefer dates stated in the transcript.

Aside, minor: `jobs` is now scanned twice in adjacent lines where one binding would do, and `meetingStartTime ?? enqueuedAt` becomes the second copy of that rule (line 107 has the first). A computed property on `PipelineJob`, next to the doc comment that already describes the fallback, would single-source it.

app/MeetingTranscriber/Sources/ProtocolGenerator.swift
```diff
@@ -91,20 +109,56 @@ enum ProtocolGenerator {
...
+        promptURL: URL = AppPaths.customPromptFile,
+        timeZone: TimeZone = .autoupdatingCurrent,
+    ) -> String {
+        let metadata = meetingMetadata(for: meetingStartTime, timeZone: timeZone)
+        var prompt = """
+        Meeting metadata:
+        Date: \(metadata.date)
+        Time: \(metadata.time)
+        Language: \(language)
+        The date and time above are authoritative. Interpret relative time expressions
```
pasrom: This is what turns the nil case above from imprecise into harmful. The block does not merely supply a date, it anchors every relative expression to it, so a date the transcript itself states loses against a processing timestamp. That is the right instruction when the value really is the meeting start.

app/MeetingTranscriber/Sources/ProtocolGenerator.swift
```diff
@@ -91,20 +109,56 @@ enum ProtocolGenerator {
...
+        meetingStartTime: Date,
+        promptURL: URL = AppPaths.customPromptFile,
+        timeZone: TimeZone = .autoupdatingCurrent,
+    ) -> String {
+        let metadata = meetingMetadata(for: meetingStartTime, timeZone: timeZone)
+        var prompt = """
+        Meeting metadata:
+        Date: \(metadata.date)
+        Time: \(metadata.time)
+        Language: \(language)
```
pasrom: Worth a second thought: this line now goes into every request, including a fully custom prompt that sets its own output language and never uses `{LANGUAGE}`. Before this change `applyLanguage` was the only path, so such a prompt never saw `AppSettings.protocolLanguage` at all. Now it is stated as a fact at the top of the request. The authority sentence below scopes itself to the date and time, not the language, but a model may still read this line as an output-language hint that contradicts the prompt body. Not obviously wrong, and possibly what you want, but it is a behaviour change for existing custom prompts and the docs you updated do not mention it.

app/MeetingTranscriber/Sources/ProtocolGenerator.swift
```diff
@@ -91,20 +109,56 @@ enum ProtocolGenerator {
...
+        The date and time above are authoritative. Interpret relative time expressions
+        in the transcript relative to this meeting date. Do not rely on the model's
+        assumed current date.
+
+        """ + applyVariables(
```
pasrom: The trailing empty line before the closing delimiter does not do what it looks like. Swift drops the line break that precedes `"""`, so the literal ends with a single `\n` and the prompt runs straight on:

```
...assumed current date.\nYou are a professional meeting minute taker.
```

I confirmed this by running the exact construct. Two empty lines are needed for the blank separator you intended. Neither new assertion notices, because both use `contains`.

app/MeetingTranscriber/Tests/TestHelpers.swift
```diff
@@ -347,7 +347,7 @@ class MockProtocolGen: ProtocolGenerating {
     var capturedDiarized: Bool?
     var shouldThrow = false
 
-    func generate(transcript: String, title: String, diarized: Bool) throws -> String {
+    func generate(transcript: String, title: String, diarized: Bool, meetingStartTime _: Date) throws -> String {
```
pasrom: This mock is the generator every pipeline test injects: `PipelineQueueTests`, `WorkflowIntegrationTests`, `AccidentalNamingAcceptTests`, `PipelineControllerTests`, `AppStateTests`, `ResamplingIntegrationTests` and `WatchLoopE2ETests` all pass `MockProtocolGen` or nil. It is the only `ProtocolGenerating` conformer in the test target, and the real generators are exercised only in their own unit tests, never through the queue. It captures transcript, title and `diarized` but drops `meetingStartTime`, so no test can observe what the pipeline actually forwards. Change the production call site to pass `Date()` instead of the job's time and the suite stays green, which is the exact defect this PR exists to prevent.

About four lines close it: add `var capturedMeetingStartTime: Date?`, assign it here, then assert it in `testAllArtifactsShareOneMeetingStartAnchoredBasename` and in `testBasenameFallsBackToEnqueuedAtWhenNoMeetingStart`. `makeStraightThroughQueue` already returns the mock and both tests currently discard it with `let (q, _)`. The second test already constructs a job with `meetingStartTime: nil`, so it is also the natural place to pin whatever you decide for the nil case.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -22,8 +22,13 @@
 
         // MARK: - ProtocolGenerating
 
-        func generate(transcript: String, title _: String, diarized: Bool) async throws -> String {
-            let prompt = ProtocolGenerator.buildSystemPrompt(diarized: diarized, language: language) + transcript
+        func generate(
+            transcript: String,
+            title _: String,
+            diarized: Bool,
+            meetingStartTime: Date = Date(),
```
pasrom: The single production call goes through `any ProtocolGenerating`, and the protocol requirement carries no default, so this default is unreachable in production. It exists only so the existing test call sites keep compiling. The cost is that any future caller holding the concrete type silently gets `now` as the meeting date, which is the bug this PR is here to remove. Same at `OpenAIProtocolGenerator.swift:62` and at `applyVariables`.

Dropping the defaults turns the omission into a compile error; the test call sites can get a convenience overload in `TestHelpers` instead.

app/MeetingTranscriber/Sources/ProtocolGenerator.swift
```diff
@@ -72,8 +73,25 @@ enum ProtocolGenerator {
 
     /// Load the protocol prompt, preferring a custom file over the built-in default.
```
pasrom: This summary line describes `loadPrompt`, not this function: `applyVariables` does pure string substitution and reads no file. The line was already misplaced before your change, sitting over `applyLanguage`, but this hunk is where it lands on the new substitution entry point, so it is worth moving or dropping while you are here.

app/MeetingTranscriber/Sources/ProtocolGenerator.swift
```diff
@@ -72,8 +73,25 @@ enum ProtocolGenerator {
...
+        timeZone: TimeZone = .autoupdatingCurrent,
+    ) -> String {
+        let metadata = meetingMetadata(for: meetingStartTime, timeZone: timeZone)
+        return prompt
+            .replacingOccurrences(of: "{LANGUAGE}", with: language)
+            .replacingOccurrences(of: "{MEETING_DATE}", with: metadata.date)
+            .replacingOccurrences(of: "{MEETING_TIME}", with: metadata.time)
+    }
+
+    /// Backwards-compatible language-only replacement for callers that do not
```
pasrom: After this change `applyLanguage` has no production caller left; a grep finds the definition and three test call sites, nothing else. The new comment names prompt-editor previews as the reason it stays, but there is no preview: the settings path writes the template and opens it in an external editor. Either delete the function and move its three tests onto `applyVariables`, or say plainly that it is kept for tests.

app/MeetingTranscriber/Sources/ProtocolGenerator.swift
```diff
@@ -91,20 +109,56 @@ enum ProtocolGenerator {
...
+            loadPrompt(from: promptURL),
+            language: language,
+            meetingStartTime: meetingStartTime,
+            timeZone: timeZone,
+        )
         if diarized { prompt += diarizationNote }
         return prompt
     }
 
+    static func meetingMetadata(
```
pasrom, line 138: Two small things here.

The `en_US_POSIX` plus explicit Gregorian pinning already exists as `DateFormatter.filenameStamp` (`DateFormatter+FilenameStamp.swift:9`), used by this very file at line 200 and by `DualSourceRecorder`. Reusing it keeps the pinning in one place. Note that the factory deliberately leaves the timezone at `.current`, so you would still assign `formatter.timeZone = timeZone` afterwards to keep the injectable parameter working.

And this function runs twice per request with identical arguments: once at line 127 and again inside `applyVariables` at line 86, so two `DateFormatter` instances are built to produce two strings. Computing the metadata once in `buildSystemPrompt` and passing it into `applyVariables` removes both the duplication and the second allocation.

app/MeetingTranscriber/Tests/ProtocolGeneratorTests.swift
```diff
@@ -209,25 +209,52 @@ final class ProtocolGeneratorTests: XCTestCase {
...
+            diarized: false,
+            language: "Polish",
+            meetingStartTime: startTime,
+            promptURL: url,
+        )
 
-        // Equality also pins that the injected URL is honoured: reading the shared
-        // user file or the built-in default would not produce this exact string.
-        XCTAssertEqual(prompt, "Protocol language: Polish.")
+        XCTAssertTrue(prompt.contains("Date: 2026-08-21"))
```
pasrom: This test lost its equality assertion and the comment that explained why equality was used: "Equality also pins that the injected URL is honoured: reading the shared user file or the built-in default would not produce this exact string."

With three `contains` checks the test no longer pins that nothing else leaks into the prompt, and it is the only coverage of that composition. Asserting the full expected string, metadata block included, keeps the original property and would additionally have caught the missing blank line noted on `ProtocolGenerator.swift`. Keeping the rationale comment alive would also help the next reader see why equality mattered here.

app/MeetingTranscriber/Tests/OpenAIProtocolGeneratorTests.swift
```diff
@@ -287,6 +287,28 @@ final class OpenAIProtocolGeneratorTests: XCTestCase { // swiftlint:disable:this
         _ = try await gen.generate(transcript: "x", title: "t", diarized: false)
     }
 
+    func testGenerateRequestIncludesAuthoritativeMeetingMetadata() async throws {
+        let startTime = try localDate(2026, 8, 21, 13, 28)
+        let metadata = ProtocolGenerator.meetingMetadata(for: startTime)
```
pasrom: The expected value is produced by the same function under test, so the assertion moves with any defect in it. Change the format to `dd-MM-yyyy`, or drop the `en_US_POSIX` locale so a non-Gregorian system calendar leaks in, and this test still passes. It only falsifies "the argument never arrived".

`ProtocolGeneratorTests` pins the literal `German 2026-08-21 13:28`, and that is the assertion actually holding the format. Pinning literals here too would make this test independent of the code it checks.

pasrom (commented):
Thanks for this, and for the care in the details. The formatter is pinned to `en_US_POSIX` with an explicit Gregorian calendar and an injectable `TimeZone`, and the new `promptURL`/`timeZone` parameters follow the test-isolation pattern the file already uses. That part is exactly right. For live recordings this is a strict improvement: a meeting that runs past midnight now gets its start date instead of the processing date.

I checked the branch locally in a worktree: `swift build` clean, the focused protocol tests green (145 tests), `swift build --build-tests -Xswiftc -DAPPSTORE` clean so the App Store variant still compiles with the changed mock, and `./scripts/lint.sh` reporting 0 violations in 476 files. On the PR itself the core lanes are green: lint, analyze, both build variants, both test variants and coverage. The label-gated E2E and quality lanes report as skipped, so they have not exercised this change.

Two things should change before this merges, both flagged inline:

1. For imported files and crash-recovered recordings `meetingStartTime` is nil by design, so the fallback puts the *processing* date into a block that then tells the model this date is authoritative and that it must not rely on its own assumption. Importing an old voice note is a documented use case, and for that path this is a regression against the previous `[Date from context or today]` wording.
2. Nothing pins that the pipeline forwards the job's meeting time. `MockProtocolGen` is the generator every pipeline test injects, and it discards the new parameter, so replacing the forwarded value with `Date()` at the call site would keep the whole suite green.

The remaining inline notes are small and not blocking.

One item that is not visible in the diff: `CLAUDE.md` (which `AGENTS.md` symlinks to) still documents `{LANGUAGE}` only. Since README and `docs/architecture-macos.md` were updated here, that file deserves the same sentence, because it is what anyone reads before touching the prompt path.

niftan5:
Thanks for the detailed review @pasrom . I addressed all functional review points in 9071062 (on top of the original implementation in b151fe4).

**Unknown meeting times for imports and recovery**
meetingStartTime is now optional through ProtocolGenerating, both providers, and the pipeline. A real recording start remains authoritative; imports, orphan recovery, and legacy jobs preserve nil instead of falling back to enqueuedAt or Date().
I introduced PipelineJob.artifactStartTime as the single fallback rule for output artifact names only. This keeps imported-file names stable while ensuring that their LLM prompt never presents the import/processing time as the meeting time.
For a missing meeting start, the authoritative metadata preamble is omitted entirely. {MEETING_DATE} and {MEETING_TIME} resolve to Unknown, so custom prompts can handle that case explicitly without receiving a misleading anchor.

**Pipeline coverage**
MockProtocolGen now captures meetingStartTime. testAllArtifactsShareOneMeetingStartAnchoredBasename asserts that a recorded start time reaches the generator unchanged, while testBasenameFallsBackToEnqueuedAtWhenNoMeetingStart asserts that the generator receives nil even though artifact naming still uses enqueuedAt.

**Prompt composition and custom prompts**
I removed the automatic Language: ... metadata line, so fully custom prompts no longer receive a potentially conflicting output-language instruction. {LANGUAGE} remains substituted only where the prompt explicitly uses that placeholder.
The metadata separator is now constructed explicitly with \\n\\n. testBuildSystemPromptSubstitutesLanguage again uses exact equality, which pins the full prompt composition, the injected custom prompt, and the intended blank line.

**API and implementation cleanup**
I removed the reachable Date() defaults from the concrete generators and prompt substitution API. Production callers must now supply an explicit optional meeting time; test-only convenience remains in TestHelpers.
I also removed the unused applyLanguage helper, moved the misplaced loadPrompt documentation to the correct function, reuse DateFormatter.filenameStamp with the injected timezone, and compute meeting metadata once per prompt.

**Tests and documentation**
The OpenAI request test now asserts literal 2026-08-21 and 13:28 values rather than deriving its expectations from meetingMetadata. CLAUDE.md now documents {MEETING_DATE} / {MEETING_TIME} and the different behavior for imports and recovery jobs; README and the architecture documentation were updated accordingly.
Local verification passed:
- focused protocol, OpenAI, pipeline, and missing-metadata tests
- swift build --build-tests -Xswiftc -DAPPSTORE
- swift build -c release
- ./scripts/lint.sh
The label-gated E2E and quality lanes were not run locally. 

Could you please re-review?

pasrom (approved):
Thanks for the thorough second round. Both blocking points are properly fixed rather than papered over: `meetingStartTime` is optional all the way through, the authoritative block is emitted only when a real recording start exists, and the pipeline forwarding is now pinned in both directions. Two things you did beyond what was asked are worth calling out: folding the fallback rule into `PipelineJob.artifactStartTime` with the filename-versus-prompt distinction spelled out in the doc comment, and replacing the tuple with a proper `MeetingPromptMetadata` type. The restored equality assertion also pins the block separator, which is a nicer fix than the one I suggested.

Two small things I will follow up on myself, since both are calls about our own default prompt rather than anything wrong in your change: for imports the template now renders a literal `Date: Unknown`, and it would be better to say something that lets the model infer a date the transcript states; and the rationale comment on the equality test is still missing. Neither is a reason to hold this.

PR #650, opened 2026-08-24
fix(app): guide protocol prompts when no meeting start was captured

pasrom:
## What

Follow-up to #643. Since that change, a job with no captured recording start (file imports, crash recovery) substitutes `Unknown` for `{MEETING_DATE}` / `{MEETING_TIME}`, and those placeholders sit inside the built-in template's output structure that the model is told to reproduce verbatim. An imported recording therefore produced a protocol header literally reading `**Date:** Unknown`, with no instruction anywhere on what to do about it.

The placeholders keep resolving to the short scalar `Unknown`. Instead, when no start was captured, the system prompt now opens with guidance in the same preamble slot the authoritative metadata block uses:

> No reliable meeting start time was captured for this recording.
> If the transcript states when this meeting itself took place, use that
> date and time; otherwise leave them as Unknown. Never substitute the
> current date or the processing time.

Design points:

- Scoped to the meeting's own date: a deadline or a past event mentioned in the transcript must not be stamped into the header as the meeting date, and transcripts state clock times constantly without ever stating their own start.
- The common case (transcript states no date) stays deterministic: the header keeps `Unknown`.
- Guidance lives in the preamble, not in the substituted values, so custom prompts that embed the placeholders as short scalars (front matter, table cells) are unaffected and no English instruction lands inside a document whose output language is configurable.

Tests: a regression test now composes the prompt against the built-in default template (both existing tests injected their own) and pins what the header placeholders resolve to plus the guidance; an invariant test asserts the no-start prompt never contains "today", so a rewording cannot quietly reintroduce a current-date fallback. Docs updated at the three sites describing the placeholder resolution, and a dropped rationale comment on the system-prompt equality assertion is restored.

## Unchanged

The authoritative metadata block behaviour from #643 is untouched: a job without a captured start still receives no authoritative time anchor, and nothing reintroduces processing time or the current date as a fallback.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -193,7 +193,7 @@ Use the `/git-workflow` skill. Commit proactively after every logical unit of wo
 **Protocol generation:**
 - `ProtocolGenerating` protocol with two implementations: `ClaudeCLIProtocolGenerator` and `OpenAIProtocolGenerator`.
 - `AppSettings.protocolProvider` enum (`.claudeCLI` / `.openAICompatible` / `.none`) selects the provider. `.none` skips LLM generation and saves the transcript only.
-- `AppSettings.protocolLanguage` string (default `"German"`) is substituted into the prompt as `{LANGUAGE}`. `{MEETING_DATE}` (`YYYY-MM-DD`) and `{MEETING_TIME}` (`HH:mm`) resolve from a captured recording start time, or to `Unknown` for imports and recovery jobs. Only a captured recording start adds the authoritative meeting-metadata preamble, so processing time is never presented as a meeting time.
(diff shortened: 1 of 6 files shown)


pasrom:
Closing this without merging. The premise does not hold up.

The concern was that an imported recording renders `**Date:** Unknown` in the protocol header, framed as a loss against the previous `[Date from context or today]` wording. But that old wording produced *today's date* whenever the transcript stated nothing, which is exactly the defect #643 fixed. `Unknown` only loses in the narrower case where the transcript states the meeting's own date, and buying that back through prompt instructions turned out to cost more than it is worth: an instruction broad enough to find the date is also broad enough to promote a deadline, and it breaks the documented `YYYY-MM-DD` shape.

The more useful finding from working on this: for crash-recovery jobs the app already holds the real recording time. `PipelineQueue+Recovery.swift` reads `.creationDateKey` and builds the job title from a `yyyyMMdd_HHmmss` stem, yet passes no `meetingStartTime`, so the protocol is saved under a timestamped filename while its own header says `Unknown`. That is a small data-flow gap worth fixing on its own terms, and it makes prompt guidance unnecessary for that path. For genuine file imports there is no reliable meeting time to recover, so `Unknown` on main is the honest answer and needs no change.

PR #651, opened 2026-08-24
feat(app): boost transcription with custom vocabulary and terminology rules

### LINKED ISSUE #575: Custom vocabulary is ignored when WhisperKit is the active engine
### What happens

Settings → Transcribe shows the "Custom vocabulary file" field only while Parakeet is
the selected engine. `AppSettings.customVocabularyPath` is persisted either way, but
`EngineController` forwards it to `ParakeetEngine` only, and `WhisperKitEngine` has no
vocabulary handling at all. With WhisperKit active, a configured vocabulary therefore
has no effect, and the field is not even visible to configure.

### Why it matters

WhisperKit is the engine that covers 99+ languages, so it is the one people keep for
anything outside Parakeet's 25. Domain terms (product names, company names, personal
names) are exactly what a general model gets wrong, so the engine with the widest reach
is currently the one that cannot be biased at all.

### Possible approach

Whisper takes prompt context through `DecodingOptions.promptTokens`. There is a prior
attempt in #205; its branch is gone but the code stays reachable at
`refs/pull/205/head`. In outline: read the file into trimmed terms, compose one
"Glossary: term, term, ..." phrase, encode it with the WhisperKit tokenizer, clamp to a
budget well below the 448 token context so prefix and audio still fit, cache it and
refresh only when the path changes, and pass `nil` when the list is empty.

That attempt hooked the settings-to-engine wiring into `AppState.syncLanguageSettings`,
which no longer exists in that shape. The equivalent seam today is `EngineController`,
next to the existing Parakeet assignment. The parse, compose and clamp steps are pure
and testable without loading a model.

### Decide before implementing

Whisper's prompt is a soft hint, not a constraint the way Parakeet's CTC boosting is,
and a long prompt can bleed into the transcript instead of only steering it. #205
asserted the wiring, never the recognition effect. Worth measuring against a fixture
containing known domain terms first, and worth deciding whether the settings field then
becomes engine-independent or stays engine-specific with wording that says how strong
the effect actually is.

niftan5:
Fixes #575

## Problem

Custom vocabulary was configured globally but ignored when WhisperKit was the active transcription engine.

## Primary change

- Pass configured custom vocabulary to WhisperKit as decoding prompt context.

## Supporting improvements

- Persist user-selected vocabulary files with security-scoped bookmarks, so a sandboxed app can access them again after restart.
- Validate vocabulary input and enforce file-size, term-count, and term-length limits before it reaches the recognizer.
- Cache both available and unavailable vocabulary revisions, avoiding repeated file reads during live transcription.
- Add optional canonical terminology rules for spelling variants. These are applied after ASR to saved transcripts and are deliberately separate from the WhisperKit decoding prompt.
- Apply terminology changes to subsequent pipeline jobs without requiring an app restart.
- Use boundary-aware, non-cascading replacements so rules do not alter parts of unrelated words or unexpectedly rewrite each other.

## Testing

- Unit tests for vocabulary parsing, validation, file access, prompt construction, cache invalidation, and terminology normalization.
- Pipeline tests covering post-ASR terminology normalization.
- `swift test --parallel --quiet --skip MenuBarIconSnapshotTests`
- `./scripts/lint.sh`
- `swift build -c release`
- `swift build --build-tests -Xswiftc -DAPPSTORE`

## Notes for reviewers

- The vocabulary quality-regression test is opt-in via `RUN_QUALITY_TESTS=1`, as it needs a real WhisperKit model run.
- Snapshot tests remain excluded because of the existing platform-dependent snapshot mismatch.

### DIFF
--- a/app/MeetingTranscriber/Sources/AppSettings+Vocabulary.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings+Vocabulary.swift
@@ -1,0 +1,110 @@
+import Foundation
+
+enum CustomVocabularyValidation: Equatable {
+    case notConfigured
+    case empty
+    case unavailable
+    case tooLarge
+    case tooManyTerms
+    case termTooLong
+    case ready(termCount: Int)
+
+    var message: String {
+        switch self {
+        case .notConfigured: "No vocabulary file selected."
+        case .empty: "Vocabulary file contains no terms."
+        case .unavailable: "Vocabulary file cannot be read."
+        case .tooLarge: "Vocabulary file is too large."
+        case .tooManyTerms: "Vocabulary file contains too many terms."
+        case .termTooLong: "Vocabulary file contains a term that is too long."
+        case let .ready(termCount): "Vocabulary file can be read: \(termCount) unique term\(termCount == 1 ? "" : "s")."
+        }
+    }
+}
+
+/// Security-scoped vocabulary-file handling. The plain path remains for UI
+/// display and migration from earlier releases; reads must use the resolved URL
+/// so the App Store build retains access after a relaunch.
+extension AppSettings {
+    var customVocabularyFile: URL? {
+        var isStale = false
+        guard let url = resolveCustomVocabularyFile(isStale: &isStale) else { return nil }
+        if isStale,
+           let refreshed = VocabularyFileAccess.withAccess(to: url, makeCustomVocabularyBookmark) {
+            updateCustomVocabularySelection(path: url.path, bookmark: refreshed)
+        }
+        return url
+    }
+
+    func setCustomVocabularyFile(_ url: URL) {
+        let bookmark = VocabularyFileAccess.withAccess(to: url) { scopedURL in
+            makeCustomVocabularyBookmark(for: scopedURL)
+        }
+        updateCustomVocabularySelection(path: url.path, bookmark: bookmark)
+        refreshCustomVocabularyValidation()
+    }
+
+    /// Handles manual path edits from the Settings text field. A bookmark is
+    /// tied to its original URL, so retaining it after a different path was
+    /// typed would silently read the wrong terminology file.
+    func setCustomVocabularyPath(_ path: String) {
+        guard path != customVocabularyPath else { return }
+        updateCustomVocabularySelection(path: path, bookmark: nil)
+        refreshCustomVocabularyValidation()
+    }
+
+    func clearCustomVocabularyFile() {
+        updateCustomVocabularySelection(path: "", bookmark: nil)
+        customVocabularyValidation = .notConfigured
+    }
+
+    func repairStaleCustomVocabularyBookmark() {
+        var isStale = false
+        guard let url = resolveCustomVocabularyFile(isStale: &isStale), isStale,
+              let refreshed = VocabularyFileAccess.withAccess(to: url, makeCustomVocabularyBookmark)
+        else { return }
+        updateCustomVocabularySelection(path: url.path, bookmark: refreshed)
+        refreshCustomVocabularyValidation()
+    }
+
+    func refreshCustomVocabularyValidation() {
+        guard let url = customVocabularyFile else {
+            customVocabularyValidation = customVocabularyPath.isEmpty ? .notConfigured : .unavailable
+            return
+        }
+        customVocabularyValidation = VocabularyFileAccess.withAccess(to: url) { url in
+            guard let revision = WhisperVocabularyPrompt.fileRevision(at: url.path) else { return .unavailable }
+            guard revision.fileSize <= UInt64(WhisperVocabularyPrompt.maximumFileBytes) else { return .tooLarge }
+            guard let contents = try? String(contentsOf: url, encoding: .utf8) else { return .unavailable }
+            let terms = WhisperVocabularyPrompt.terms(from: contents)
+            guard !terms.isEmpty else { return .empty }
+            guard terms.count <= WhisperVocabularyPrompt.maximumTermCount else { return .tooManyTerms }
+            guard terms.allSatisfy({ $0.utf8.count <= WhisperVocabularyPrompt.maximumTermBytes }) else {
+                return .termTooLong
+            }
+            return .ready(termCount: terms.count)
+        }
+    }
+
+    private func resolveCustomVocabularyFile(isStale: inout Bool) -> URL? {
+        if let bookmark = customVocabularyBookmark,
+           let url = try? URL(
+               resolvingBookmarkData: bookmark,
+               options: .withSecurityScope,
+               relativeTo: nil,
+               bookmarkDataIsStale: &isStale,
+           ) {
+            return url
+        }
+        guard !customVocabularyPath.isEmpty else { return nil }
+        return URL(fileURLWithPath: customVocabularyPath)
+    }
+
+    private func makeCustomVocabularyBookmark(for url: URL) -> Data? {
+        try? url.bookmarkData(
+            options: .withSecurityScope,
+            includingResourceValuesForKeys: nil,
+            relativeTo: nil,
+        )
+    }
+}
--- a/app/MeetingTranscriber/Sources/ParakeetEngine.swift
+++ b/app/MeetingTranscriber/Sources/ParakeetEngine.swift
@@ -16,8 +16,21 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
     private(set) var downloadProgress: Double = 0
     private(set) var transcriptionProgress: Double = 0
 
-    /// Path to a custom vocabulary file for CTC boosting. Set from AppSettings before loadModel().
-    var customVocabularyPath: String = ""
+    /// Path to a custom vocabulary file for CTC boosting. A change while the
+    /// model is loaded is applied to the next transcription without restart.
+    var customVocabularyPath: String = "" {
+        didSet {
+            guard customVocabularyPath != oldValue else { return }
+            invalidateVocabularyConfiguration()
+        }
+    }
+
+    var customVocabularyBookmark: Data? {
+        didSet {
+            guard customVocabularyBookmark != oldValue else { return }
+            invalidateVocabularyConfiguration()
+        }
+    }
 
     /// Optional ISO 639-1 language hint. Empty/nil = auto-detect (FluidAudio's
     /// v3 TDT decoder picks the script freely, which can drift Cyrillic ↔ Latin
@@ -45,8 +58,27 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
 
     private var vocabularyBooster: VocabularyBooster?
 
-    /// Tracks the last successfully configured vocabulary path to avoid redundant CTC model downloads.
-    private var currentVocabularyPath: String = ""
+    // The CTC bridge is replaceable only in debug/test builds, where loading
+    // production CTC models would make lifecycle tests impractical.
+    #if DEBUG
+        private var vocabularyPreparationOverride: (([String]) async throws -> Void)?
+        private var fileTranscriptionOverride: ((URL) async throws -> ASRResult)?
+        private var sampleTranscriptionOverride: (([Float]) async throws -> ASRResult)?
+    #endif
+
+    /// The exact vocabulary revision currently being prepared or already
+    /// handled. Unlike a path-only marker it changes when a user edits a file.
+    private var vocabularyPreparationState: ParakeetVocabularyPreparationState?
+    private var vocabularyRefreshGate = ParakeetVocabularyRefreshGate()
+
+    private func invalidateVocabularyConfiguration() {
+        _ = vocabularyRefreshGate.invalidate()
+        vocabularyPreparationState = nil
+        // Do not rescore a new recording with terms from the file that was just
+        // replaced. `ensureVocabularyConfiguration()` prepares the new booster
+        // before the next decode begins.
+        vocabularyBooster = nil
+    }
 
     func loadModel() async {
         await modelLoad.run { [self] in
@@ -64,11 +96,6 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
                 try await manager.loadModels(models)
                 asrManager = manager
                 modelState = .loaded
-
-                // Configure custom vocabulary boosting if a vocabulary file is set
-                if !customVocabularyPath.isEmpty {
-                    try await configureVocabulary(from: customVocabularyPath)
-                }
             } catch {
                 logger.error("Parakeet model load failed: \(error.localizedDescription, privacy: .public)")
                 modelState = .unloaded
@@ -79,6 +106,9 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
 
     private func ensureModel() async throws {
         if asrManager != nil { return }
+        #if DEBUG
+            if fileTranscriptionOverride != nil, sampleTranscriptionOverride != nil { return }
+        #endif
         logger.info("Parakeet: model not loaded, loading…")
         await loadModel()
         guard asrManager != nil else {
@@ -90,13 +120,10 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
 
     func transcribeSegments(audioPath: URL) async throws -> [TimestampedSegment] {
         try await ensureModel()
-        guard let manager = asrManager else {
-            throw TranscriptionError.modelNotLoaded
-        }
+        await ensureVocabularyConfiguration()
 
         transcriptionProgress = 0
-        var decoderState = await TdtDecoderState.make(decoderLayers: manager.decoderLayerCount)
-        var result = try await manager.transcribe(audioPath, decoderState: &decoderState, language: fluidLanguageHint)
+        var result = try await decodeFile(audioPath)
         transcriptionProgress = 1.0
 
         // Apply CTC vocabulary rescoring if configured
@@ -123,14 +150,34 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
     /// can be stateless across VAD segments.
     func transcribeSamples(_ samples: [Float]) async throws -> String {
         try await ensureModel()
-        guard let manager = asrManager else {
-            throw TranscriptionError.modelNotLoaded
-        }
+        let result = try await decodeSamples(samples)
+        return result.text.trimmingCharacters(in: .whitespaces)
+    }
+
+    private func decodeFile(_ audioPath: URL) async throws -> ASRResult {
+        #if DEBUG
+            if let fileTranscriptionOverride {
+                return try await fileTranscriptionOverride(audioPath)
+            }
+        #endif
+        guard let manager = asrManager else { throw TranscriptionError.modelNotLoaded }
         var decoderState = await TdtDecoderState.make(decoderLayers: manager.decoderLayerCount)
-        let result = try await manager.transcribe(
+        return try await manager.transcribe(
+            audioPath, decoderState: &decoderState, language: fluidLanguageHint,
+        )
+    }
+
+    private func decodeSamples(_ samples: [Float]) async throws -> ASRResult {
+        #if DEBUG
+            if let sampleTranscriptionOverride {
+                return try await sampleTranscriptionOverride(samples)
+            }
+        #endif
+        guard let manager = asrManager else { throw TranscriptionError.modelNotLoaded }
+        var decoderState = await TdtDecoderState.make(decoderLayers: manager.decoderLayerCount)
+        return try await manager.transcribe(
             samples, decoderState: &decoderState, language: fluidLanguageHint,
         )
-        return result.text.trimmingCharacters(in: .whitespaces)
     }
 
     /// Run CTC keyword spotting on the audio and rescore the TDT transcript.
@@ -176,46 +223,199 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
 
     // MARK: - Custom Vocabulary
 
-    /// Configure custom vocabulary for CTC boosting (Parakeet only).
-    ///
-    /// Loads a vocabulary file and downloads CTC models for keyword spotting.
-    /// After configuration, `transcribeSegments` will automatically apply CTC-based
-    /// vocabulary rescoring to improve recognition of domain-specific terms.
-    ///
-    /// Skips silently if path is empty. Logs a warning if loading fails.
-    func configureVocabulary(from path: String) async throws {
-        guard !path.isEmpty else {
-            vocabularyBooster = nil
-            currentVocabularyPath = ""
+    #if DEBUG
+        func installVocabularyPreparationForTesting(
+            _ prepare: @escaping ([String]) async throws -> Void,
+        ) {
+            vocabularyPreparationOverride = prepare
+        }
+
+        /// Installs the final ASR decode boundary for focused lifecycle tests. It
+        /// is excluded from release builds; production always uses `AsrManager`.
+        func installTranscriptionForTesting(
+            file: @escaping (URL) async throws -> ASRResult,
+            samples: @escaping ([Float]) async throws -> ASRResult,
+        ) {
+            fileTranscriptionOverride = file
+            sampleTranscriptionOverride = samples
+        }
+    #endif
+
+    /// Ensures the exact current file revision is ready before a decode. This
+    /// makes an edit to an existing file take effect on the next transcription,
+    /// and avoids an unstructured background task racing the decode.
+    func ensureVocabularyConfiguration() async {
+        let configuration = vocabularyConfiguration(for: customVocabularyPath, bookmark: customVocabularyBookmark)
+        guard vocabularyPreparationState?.configuration != configuration else { return }
+        await configureVocabulary(
+            configuration,
+            attempt: vocabularyRefreshGate.current,
+            requiresCurrentSelection: true,
+        )
+    }
+
+    private func vocabularyConfiguration(for path: String, bookmark: Data?) -> ParakeetVocabularyConfiguration {
+        guard let vocabularyURL = VocabularyFileAccess.resolve(path: path, bookmark: bookmark) else {
+            return ParakeetVocabularyConfiguration(path: path, bookmark: bookmark, revision: nil)
+        }
+        let revision = VocabularyFileAccess.withAccess(to: vocabularyURL) { url in
+            WhisperVocabularyPrompt.fileRevision(at: url.path)
+        }
+        return ParakeetVocabularyConfiguration(path: path, bookmark: bookmark, revision: revision)
+    }
+
+    private func configureVocabulary(
+        _ configuration: ParakeetVocabularyConfiguration,
+        attempt: Int,
+        requiresCurrentSelection: Bool,
+    ) async {
+        guard canAdoptVocabulary(
+            configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection,
+        ) else { return }
+        // A new revision must never rescore with the prior revision while CTC
+        // models are written or loaded. Mark it in-flight so concurrent decodes
+        // do not duplicate the same preparation work.
+        vocabularyBooster = nil
+        vocabularyPreparationState = .preparing(configuration)
+
+        guard !configuration.path.isEmpty else {
+            adoptNoVocabulary(configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection)
             return
         }
-        guard path != currentVocabularyPath else { return }
 
-        let vocab: CustomVocabularyContext
-        let ctcModels: CtcModels
+        guard let terms = vocabularyTerms(for: configuration), !terms.isEmpty else {
+            adoptNoVocabulary(configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection)
+            logger.warning("Parakeet: vocabulary file is unavailable or contains no usable terms")
+            return
+        }
+
+        if await prepareVocabularyUsingOverrideIfPresent(
+            terms,
+            configuration: configuration,
+            attempt: attempt,
+            requiresCurrentSelection: requiresCurrentSelection,
+        ) { return }
+
+        // FluidAudio loads from a path. Give it the bounded, deduplicated list
+        // shared with WhisperKit rather than the original raw file, so both
+        // engines honour the validation limits users see in Settings.
         do {
-            (vocab, ctcModels) = try await CustomVocabularyContext.loadWithCtcTokens(
-                from: path,
+            let preparedVocabularyURL = FileManager.default.temporaryDirectory
+                .appendingPathComponent("parakeet-vocabulary-\(UUID().uuidString).txt")
+            defer { try? FileManager.default.removeItem(at: preparedVocabularyURL) }
+            try terms.joined(separator: "\n").write(
+                to: preparedVocabularyURL,
+                atomically: true,
+                encoding: .utf8,
+            )
+            let (vocab, ctcModels) = try await CustomVocabularyContext.loadWithCtcTokens(
+                from: preparedVocabularyURL.path,
                 ctcVariant: .ctc110m,
             )
+            guard canAdoptVocabulary(
+                configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection,
+            ) else { return }
+
+            let spotter = CtcKeywordSpotter(models: ctcModels, blankId: ctcModels.vocabulary.count)
+            let rescorer = try await VocabularyRescorer.create(
+                spotter: spotter,
+                vocabulary: vocab,
+                config: .default,
+                ctcModelDirectory: CtcModels.defaultCacheDirectory(for: ctcModels.variant),
+            )
+            guard canAdoptVocabulary(
+                configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection,
+            ) else { return }
+            vocabularyBooster = VocabularyBooster(context: vocab, spotter: spotter, rescorer: rescorer)
+            vocabularyPreparationState = .ready(configuration)
+            logger.info("Parakeet: custom vocabulary loaded: \(vocab.terms.count) terms")
         } catch {
-            logger.warning("Parakeet: failed to load vocabulary from \(path): \(error.localizedDescription, privacy: .public)")
-            return
+            adoptVocabularyFailure(
+                configuration,
+                attempt: attempt,
+                requiresCurrentSelection: requiresCurrentSelection,
+                error: error,
+            )
         }
+    }
 
-        let blankId = ctcModels.vocabulary.count
-        let spotter = CtcKeywordSpotter(models: ctcModels, blankId: blankId)
+    private func prepareVocabularyUsingOverrideIfPresent(
+        _ terms: [String],
+        configuration: ParakeetVocabularyConfiguration,
+        attempt: Int,
+        requiresCurrentSelection: Bool,
+    ) async -> Bool {
+        #if DEBUG
+            guard let vocabularyPreparationOverride else { return false }
+            do {
+                try await vocabularyPreparationOverride(terms)
+            } catch {
+                adoptVocabularyFailure(
+                    configuration,
+                    attempt: attempt,
+                    requiresCurrentSelection: requiresCurrentSelection,
+                    error: error,
+                )
+                return true
+            }
+            guard canAdoptVocabulary(
+                configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection,
+            ) else { return true }
+            vocabularyBooster = nil
+            vocabularyPreparationState = .ready(configuration)
+            return true
+        #else
+            false
+        #endif
+    }
 
-        let ctcModelDir = CtcModels.defaultCacheDirectory(for: ctcModels.variant)
-        let rescorer = try await VocabularyRescorer.create(
-            spotter: spotter,
-            vocabulary: vocab,
-            config: .default,
-            ctcModelDirectory: ctcModelDir,
-        )
+    // swiftlint:disable:next discouraged_optional_collection
+    private func vocabularyTerms(for configuration: ParakeetVocabularyConfiguration) -> [String]? {
+        guard let vocabularyURL = VocabularyFileAccess.resolve(
+            path: configuration.path, bookmark: configuration.bookmark,
+        ) else { return nil }
+        // swiftlint:disable:next discouraged_optional_collection
+        return VocabularyFileAccess.withAccess(to: vocabularyURL) { url -> [String]? in
+            guard let revision = WhisperVocabularyPrompt.fileRevision(at: url.path) else { return nil }
+            return WhisperVocabularyPrompt.terms(fromFileAt: url.path, revision: revision)
+        }
+    }
 
-        vocabularyBooster = VocabularyBooster(context: vocab, spotter: spotter, rescorer: rescorer)
-        currentVocabularyPath = path
-        logger.info("Parakeet: custom vocabulary loaded: \(vocab.terms.count) terms")
+    private func adoptNoVocabulary(
+        _ configuration: ParakeetVocabularyConfiguration,
+        attempt: Int,
+        requiresCurrentSelection: Bool,
+    ) {
+        guard canAdoptVocabulary(
+            configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection,
+        ) else { return }
+        vocabularyBooster = nil
+        vocabularyPreparationState = .unavailable(configuration)
+    }
+
+    private func adoptVocabularyFailure(
+        _ configuration: ParakeetVocabularyConfiguration,
+        attempt: Int,
+        requiresCurrentSelection: Bool,
+        error: any Error,
+    ) {
+        guard canAdoptVocabulary(
+            configuration, attempt: attempt, requiresCurrentSelection: requiresCurrentSelection,
+        ) else { return }
+        vocabularyBooster = nil
+        vocabularyPreparationState = .failed(configuration)
+        logger.warning("Parakeet: vocabulary preparation failed for \(configuration.path): \(error.localizedDescription, privacy: .public)")
+    }
+
+    private func canAdoptVocabulary(
+        _ configuration: ParakeetVocabularyConfiguration,
+        attempt: Int,
+        requiresCurrentSelection: Bool,
+    ) -> Bool {
+        guard requiresCurrentSelection else { return true }
+        return vocabularyRefreshGate.isCurrent(attempt)
+            && configuration == vocabularyConfiguration(
+                for: customVocabularyPath, bookmark: customVocabularyBookmark,
+            )
     }
 }
--- a/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
@@ -35,6 +35,10 @@ extension PipelineQueue {
         /// Segments cached for diarization reuse (avoids double transcription).
         let cachedSegments: [TimestampedSegment]? // swiftlint:disable:this discouraged_optional_collection
         let isDualSource: Bool
+        /// The exact rules snapshot used for the initial transcription. The
+        /// defensive re-transcription fallback must not pick up Settings edits
+        /// made while this job is being processed.
+        let terminologyNormalizer: TerminologyNormalizer
     }
 
     /// Typed errors thrown by the pipeline stages.
@@ -290,6 +294,10 @@ extension PipelineQueue {
         startElapsedTimer()
         logger.info("[\(ctx.shortID, privacy: .public)] transcription_start title=\(ctx.title, privacy: .private)")
 
+        // Snapshot terminology once per job. A user edit while either track is
+        // decoding must affect the next job, not leave this dual-source result
+        // with differently normalized app and microphone segments.
+        let normalizer = terminologyNormalizer()
         let transcript: String
         // Segments cached for potential diarization reuse (avoids double transcription)
         var cachedSegments: [TimestampedSegment]? // swiftlint:disable:this discouraged_optional_collection
@@ -299,8 +307,9 @@ extension PipelineQueue {
                 ctx, engine: engine, workDir: workDir,
                 appAudioPath: appAudioPath, micAudioPath: micAudioPath,
             )
-            cachedSegments = segments
-            transcript = segments.transcriptText
+            let normalizedSegments = normalize(segments, with: normalizer)
+            cachedSegments = normalizedSegments
+            transcript = normalizedSegments.transcriptText
         } else {
             // Single-source: resample mix to 16kHz
             guard let mixPath = ctx.mixPath else {
@@ -320,7 +329,8 @@ extension PipelineQueue {
             }
 
             // Use transcribeSegments to cache results for diarization
-            var segments = try await engine.transcribeSegments(audioPath: transcriptionPath)
+            let rawSegments = try await engine.transcribeSegments(audioPath: transcriptionPath)
+            var segments = normalize(rawSegments, with: normalizer)
 
             // Remap timestamps back to original timeline if VAD was used
             if let map = vadMap {
@@ -342,7 +352,27 @@ extension PipelineQueue {
             "[\(ctx.shortID, privacy: .public)] transcription_complete segments=\(segCount, privacy: .public) duration=\(totalSecs, privacy: .public)s",
         )
 
-        return TranscriptionOutput(transcript: transcript, cachedSegments: cachedSegments, isDualSource: isDualSource)
+        return TranscriptionOutput(
+            transcript: transcript,
+            cachedSegments: cachedSegments,
+            isDualSource: isDualSource,
+            terminologyNormalizer: normalizer,
+        )
+    }
+
+    private func normalize(
+        _ segments: [TimestampedSegment],
+        with normalizer: TerminologyNormalizer,
+    ) -> [TimestampedSegment] {
+        guard !normalizer.isEmpty else { return segments }
+        return segments.map { segment in
+            TimestampedSegment(
+                start: segment.start,
+                end: segment.end,
+                text: normalizer.normalize(segment.text),
+                speaker: segment.speaker,
+            )
+        }
     }
 
     /// Stage 2 — optional speaker diarization. Returns the transcript with
@@ -570,7 +600,8 @@ extension PipelineQueue {
         } else if transcription.isDualSource {
             return nil
         } else {
-            cachedSegments = try await engine.transcribeSegments(audioPath: mix16k)
+            let rawSegments = try await engine.transcribeSegments(audioPath: mix16k)
+            cachedSegments = normalize(rawSegments, with: transcription.terminologyNormalizer)
         }
         return renderLabeledTranscript(
             run: run, cachedSegments: cachedSegments,
--- a/app/MeetingTranscriber/Sources/PipelineQueue.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue.swift
@@ -64,6 +64,10 @@ class PipelineQueue {
     /// nil disables JSONL logging. AppState injects a real instance for production;
     /// tests leave it nil unless they explicitly want to assert on the log.
     let recognitionStatsLog: RecognitionStatsLog?
+    /// Explicit canonical spelling rules, applied after the ASR engine returns
+    /// text. Kept separate from decoder vocabulary because decoder hints are
+    /// probabilistic while representation rules are deterministic.
+    let terminologyNormalizer: () -> TerminologyNormalizer
 
     /// nil disables per-stage timing capture. AppState injects a real instance;
     /// tests leave it nil unless asserting on the log.
@@ -234,6 +238,7 @@ class PipelineQueue {
         self.snapshotWriter = snapshotWriter
         self.vadConfig = nil
         self.recognitionStatsLog = nil
+        terminologyNormalizer = { TerminologyNormalizer() }
         self.stageTimingLog = stageTimingLog
         self.completedJobLifetime = completedJobLifetime
         self.terminalJobStore = terminalJobStore
@@ -309,6 +314,7 @@ class PipelineQueue {
         snapshotWriter: @escaping @Sendable ([PipelineJob], URL) throws -> Void = PipelineSnapshot.save,
         vadConfig: VADConfig? = nil,
         recognitionStatsLog: RecognitionStatsLog? = nil,
+        terminologyNormalizer: @escaping () -> TerminologyNormalizer = { TerminologyNormalizer() },
         stageTimingLog: StageTimingLog? = nil,
         completedJobLifetime: TimeInterval = 60,
         terminalJobStore: TerminalJobStore? = nil,
@@ -339,6 +345,7 @@ class PipelineQueue {
         self.snapshotWriter = snapshotWriter
         self.vadConfig = vadConfig
         self.recognitionStatsLog = recognitionStatsLog
+        self.terminologyNormalizer = terminologyNormalizer
         self.stageTimingLog = stageTimingLog
         self.completedJobLifetime = completedJobLifetime
         self.terminalJobStore = terminalJobStore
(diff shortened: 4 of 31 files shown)


app/MeetingTranscriber/Sources/WhisperVocabularyPrompt.swift
```diff
@@ -0,0 +1,156 @@
...
+
+// swiftlint:disable discouraged_optional_collection
+
+/// Converts the shared one-term-per-line vocabulary file into a bounded Whisper
+/// decoder prompt. The individual terms stay in file order, so earlier entries
+/// retain priority when the prompt budget is exhausted.
+enum WhisperVocabularyPrompt {
+    /// Keep this well below WhisperKit's own prompt-context limit so a user
+    /// vocabulary cannot crowd out the decoder's control tokens.
+    static let defaultTokenBudget = 96
```
pasrom: Blocker, see the summary for the full trace.

96 tokens plus the `startOfPrev` token is about 43% of WhisperKit's 224-token `maxTokenContext`, which is shared between the prefill and the generated tokens. The per-window generation budget drops from roughly 219 to roughly 122, and a window that hits the cap mid-text advances by the full `segmentSize` in `SegmentSeeker` instead of re-seeking, so its remaining audio is never decoded.

The comment above says the budget is kept "well below WhisperKit's own prompt-context limit so a user vocabulary cannot crowd out the decoder's control tokens". That is true of the control tokens and of WhisperKit's own `suffix(111)` prompt cap, but it says nothing about the generation budget, which is what actually gets halved. Worth rewording either way.

One latent detail if this number is ever raised: WhisperKit truncates the prompt with `Array(promptTokens.suffix(maxPromptLen))` where `maxPromptLen` is 111, so any budget above 111 would silently keep the *last* terms in the file, inverting the file-order priority documented in `tokens(for:...)`.

app/MeetingTranscriber/Sources/TerminologyNormalizer.swift
```diff
@@ -0,0 +1,79 @@
...
+        text.components(separatedBy: .newlines).compactMap { line in
+            let pieces = line.components(separatedBy: "=>")
+            guard pieces.count == 2 else { return nil }
+            let canonical = pieces[0].trimmingCharacters(in: .whitespacesAndNewlines)
+            guard !canonical.isEmpty else { return nil }
+
+            let variants = pieces[1]
+                .components(separatedBy: "|")
+                .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+                .filter { !$0.isEmpty && $0.caseInsensitiveCompare(canonical) != .orderedSame }
```
pasrom: Blocker, plus a second issue on the same line.

**Expansion rules corrupt already-correct text.** This filter drops a variant only when it equals the canonical. A variant contained *in* the canonical survives, and then matches inside correct output:

```
rule: Kubernetes Cluster => Cluster
in  : der Kubernetes Cluster laeuft stabil
out : der Kubernetes Kubernetes Cluster laeuft stabil
```

Neither `nextAvailableLocation` nor the variant-length ordering helps, because there is a single candidate and no protected canonical match. Applying the normalizer twice inserts another copy. `testRulesDoNotCascadeIntoLaterRules` covers rule-to-rule cascade but not a rule matching inside its own canonical.

**Casing-only rules silently become no rules.** `caseInsensitiveCompare` also rejects `Kubernetes => kubernetes`, which leaves `variants` empty, so the `guard` below drops the whole rule. Since matching is already case-insensitive, there is currently no way to express "normalize the casing of this term", which is probably the single most common use of a canonical-spelling feature.

The same silent drop applies to malformed lines: `Aster -> Astor` (wrong arrow) and `A => B => C` both vanish without any feedback, while the vocabulary file right next to this gets a full validation status line. Some counterpart for rules would help, even just "N rules active".

app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
```diff
@@ -304,6 +304,19 @@ extension PipelineQueue {
         return TranscriptionOutput(transcript: transcript, cachedSegments: cachedSegments, isDualSource: isDualSource)
     }
 
+    private func normalize(_ segments: [TimestampedSegment]) -> [TimestampedSegment] {
```
pasrom: Three things about building the normalizer here rather than once per job.

**Rules are not snapshotted per job.** This runs after each track's `await`, so a dual-source recording can normalize its app track under one rule set and its microphone track under another if the user edits the rules while the second track is decoding. The merged transcript is then internally inconsistent. Snapshotting once at the start of the transcription stage would fix it and would also stop recompiling every rule's `NSRegularExpression` on each of the four call sites.

**This is main-actor work with no upper bound.** `PipelineQueue` is `@MainActor`, and `terminologyRulesText` is a free-form `TextEditor` with no size or count validation, in deliberate contrast to the vocabulary file next to it (256 KB / 2000 terms / 512 bytes per term). An hour-long meeting is on the order of 1500 segments, so the cost is rules times segments of `enumerateMatches` on the main actor with nothing capping either factor.

**Live captions are not normalized.** That is defensible and the help text says "Applied to saved transcripts after ASR", so this is only a note: a term shown one way live and another way in the protocol may surprise people.

app/MeetingTranscriber/Sources/PipelineQueue.swift
```diff
@@ -305,6 +310,7 @@ class PipelineQueue {
         snapshotWriter: @escaping @Sendable ([PipelineJob], URL) throws -> Void = PipelineSnapshot.save,
         vadConfig: VADConfig? = nil,
         recognitionStatsLog: RecognitionStatsLog? = nil,
+        terminologyNormalizer: @escaping () -> TerminologyNormalizer = { TerminologyNormalizer() },
```
pasrom, line 317: Both initialisers default to an empty normalizer, and `normalize` short-circuits on `isEmpty`, so every `PipelineQueue` built in tests skips this stage entirely.

I could not find any test that injects a non-empty normalizer and asserts the resulting transcript changed. The consequence is that the four `normalize(...)` call sites in `PipelineQueue+Stages.swift` could be deleted, or `PipelineController` could read the wrong settings property, and the suite would stay green. Unlike the Whisper prompt, this stage is deterministic and cheap to test, so a single test that runs a queue with `Aster => Astor` and asserts the saved transcript is enough.

The PR description lists "Pipeline tests covering post-ASR terminology normalization" under Testing. If that test exists somewhere, it did not make it onto this branch.

app/MeetingTranscriber/Tests/Quality/WhisperKitQualityTests.swift
```diff
@@ -40,6 +40,46 @@ final class WhisperKitQualityTests: XCTestCase {
...
+        let hinted = WhisperKitEngine()
+        hinted.modelVariant = modelVariant
+        hinted.language = "de"
+        hinted.customVocabularyPath = vocabularyFile.path
+        await hinted.loadModel()
+        let hintedSegments = try await hinted.transcribeSegments(audioPath: truth.audioURL)
+        let hintedText = hintedSegments.map(\.text).joined(separator: " ")
+        let hintedWER = WERCalculator.werBreakdown(reference: truth.text, hypothesis: hintedText).wer
+
+        XCTAssertLessThanOrEqual(
```
pasrom, line 89: This assertion cannot fail on "the vocabulary had no effect".

Both runs are greedy decodes of the same fixture. If `promptTokens` never reached the decoder, or if `vocabularyPromptTokens` returned nil, the two transcripts would be identical, `hintedWER == baselineWER`, and the bound holds with room to spare. The bound is one-sided and anchored at equality, so run-to-run noise only widens the pass region. The test does not even assert that the two transcripts differ.

Two further limits: the vocabulary is three common German phrases, roughly 10 tokens, which the baseline already transcribes correctly, so there is no headroom for a positive result and the 96-token budget is never exercised; and the fixture is 28 words in 17s, which cannot reach the context cap under any prompt size, so it also cannot detect the truncation issue from the summary.

The comment above calls this a regression guard, which is fair, and it does have some value: a prompt that collapsed decoding outright would trip it. But as evidence that the feature works it proves nothing.

A test that does falsify "no effect" needs a term the baseline reliably gets wrong plus a two-sided pair: first assert the baseline transcript does *not* contain the term, which guards the premise, then assert the hinted transcript does. That needs a new fixture from the generator scripts with a handful of invented, unambiguously spelled terms, and since prompting is probabilistic, asserting a majority of them corrected is more stable than all of them.

app/MeetingTranscriber/Sources/AppSettings+Vocabulary.swift
```diff
@@ -0,0 +1,103 @@
...
+        guard let url = customVocabularyFile else {
+            customVocabularyValidation = customVocabularyPath.isEmpty ? .notConfigured : .unavailable
+            return
+        }
+        customVocabularyValidation = VocabularyFileAccess.withAccess(to: url) { url in
+            guard let revision = WhisperVocabularyPrompt.fileRevision(at: url.path) else { return .unavailable }
+            guard revision.fileSize <= UInt64(WhisperVocabularyPrompt.maximumFileBytes) else { return .tooLarge }
+            guard let contents = try? String(contentsOf: url, encoding: .utf8) else { return .unavailable }
+            let terms = WhisperVocabularyPrompt.terms(from: contents)
+            guard terms.count <= WhisperVocabularyPrompt.maximumTermCount else { return .tooManyTerms }
```
pasrom, line 81: The validation ceiling and what the engines actually consume are far apart, in both directions.

**WhisperKit.** 2000 terms are accepted, but `defaultTokenBudget` is 96 tokens, which admits roughly 20 to 30 terms. Someone who curates a 500-term glossary sees "Vocabulary file is ready: 500 unique terms." and gets about 5% of it. The status is not wrong about the file, but it reads as a statement about the feature.

**Parakeet.** It does not consume this parsed list at all; `configureVocabulary` hands the raw path to FluidAudio, which loads and tokenizes the file itself. So the limits enforced here are advisory for that engine: a file reported as `.tooManyTerms` or `.termTooLong` is still loaded in full. The dedup gap is the sharpest version: 20000 identical lines is about 200 KB, passes the size check, dedups to `.ready(termCount: 1)`, and Parakeet then carries 20000 entries through its CTC search on every recording.

**Staleness.** The only refresh triggers are `AppSettings.init` and the `.onChange(of: customVocabularyPath)` in the settings view, so if the file is edited, deleted, or access is lost mid-session, the status keeps claiming "ready: N" indefinitely.

One small edge: a file of only blank lines validates as "Vocabulary file is ready: 0 unique terms."

Also worth noting that the dedup here is case-sensitive while `TerminologyNormalizer` treats casing as insignificant, so the same feature uses two different notions of "duplicate".

app/MeetingTranscriber/Sources/AppSettings+Vocabulary.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    func clearCustomVocabularyFile() {
+        customVocabularyPath = ""
+        customVocabularyBookmark = nil
+        customVocabularyValidation = .notConfigured
+    }
+
+    func repairStaleCustomVocabularyBookmark() {
+        var isStale = false
+        guard let url = resolveCustomVocabularyFile(isStale: &isStale), isStale,
```
pasrom, line 63: Minting a security-scoped bookmark needs live access to the resource, and this path resolves the stale bookmark and calls `makeCustomVocabularyBookmark` without a `startAccessingSecurityScopedResource()` bracket around it. Under App Sandbox the `try?` would swallow the throw, the `guard` would fail, and the bookmark would stay stale on every launch. The documented pattern is start access, recreate, stop.

In fairness this mirrors `repairStaleCustomOutputDirBookmark` exactly, so either both work or neither does, and I have no evidence for either. Since the sandboxed build is the only reason these bookmarks exist, it seems worth confirming once rather than doubling the pattern on assumption.

app/MeetingTranscriber/Sources/VocabularyFileAccess.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+/// Resolves persisted sandbox access for terminology files and brackets each
+/// read with the matching security scope. Plain paths remain supported for
+/// existing non-sandbox installations that predate bookmarks.
+enum VocabularyFileAccess {
+    static func resolve(path: String, bookmark: Data?) -> URL? {
+        if let bookmark {
+            var stale = false
```
pasrom, line 9: `stale` is passed to `URL(resolvingBookmarkData:)` and then never read, so this resolve path can never notice that the bookmark needs renewing.

The only renewal is `repairStaleCustomVocabularyBookmark`, which runs once per launch from `AppState.makeDefaultSettings`. Both engines hold their own copy of `customVocabularyBookmark` and call this function directly, so within a session a bookmark that goes stale is used as-is indefinitely and the refreshed data is never written back to settings.

Once it stops resolving entirely, the fallback below returns `URL(fileURLWithPath: path)`, a plain path the sandbox cannot read, and the feature turns itself off while the settings pane still shows the file as configured.

Related, and the reason this matters more than it looks: `customVocabularyPath` is still a freely settable `var` with a persisting `didSet`, while this function prefers the bookmark. `setCustomVocabularyPath` documents exactly why that combination is dangerous, but nothing enforces going through it. Making the stored property `private(set)` and routing all writes through the setter would close that.

app/MeetingTranscriber/Sources/ParakeetEngine.swift
```diff
@@ -191,11 +192,18 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
         }
         guard path != currentVocabularyPath else { return }
 
+        let vocabularyURL = VocabularyFileAccess.resolve(path: path, bookmark: customVocabularyBookmark)
```
pasrom: This is reached only from `loadModel()`, and `ensureModel()` returns early once `asrManager != nil`, so a vocabulary selected after the engine has loaded never reaches Parakeet until the app is restarted.

That staleness predates this PR, but the PR changes what the user is told about it: the new status line says "Vocabulary file is ready: N unique terms", `EngineController` syncs both the path and the new bookmark onto the engine, and the equivalent staleness on the WhisperKit side is now fixed by the per-decode revision check. So the two engines sit behind one shared file picker with different behaviour, and nothing in the UI or the help text says so.

Either reconfiguring on change or a sentence in the engine-specific help text would resolve it.

app/MeetingTranscriber/Sources/WhisperKitEngine.swift
```diff
@@ -122,16 +151,35 @@ final class WhisperKitEngine: TranscribingEngine, StreamingTranscribingEngine {
 
     /// Ensure model is loaded, loading it if necessary.
     private func ensureModel() async throws {
-        if pipe != nil { return }
+        if decodingClient != nil { return }
```
pasrom: This predicate changed from `pipe != nil` to `decodingClient != nil`, and `decodingClient` is `decodingClientOverride ?? pipe`. With an override installed, `ensureModel()` returns without ever loading a model, so `modelState` stays `.unloaded` while transcription succeeds.

That combination is only reachable from tests today, so it is not a runtime bug. It is still a production predicate that a mutable test hook can flip, and `modelState` is the value the automation surface reports and the driver scripts poll to wait for model preload. Worth keeping the production path keyed on `pipe` and letting only the decode call sites resolve through the seam.

More generally, `installDecodingClientForTesting` and `installVocabularyTermsLoaderForTesting` plus their stored properties are production API that exists only for tests, which the repo guidelines ask to avoid. Since the test target uses `@testable import`, `private` plus a narrower injection point would work as well.

The seam does earn something real: the flow tests confirm the engine passes the options it should without loading a model. That is worth keeping in some form, it is only the shape I would push back on.

app/MeetingTranscriber/Sources/WhisperKitEngine.swift
```diff
@@ -213,16 +271,76 @@ final class WhisperKitEngine: TranscribingEngine, StreamingTranscribingEngine {
...
+        return VocabularyFileAccess.withAccess(to: vocabularyURL) { url in
+            vocabularyPromptTokens(for: decodingClient, at: url.path)
+        }
+    }
+
+    private func vocabularyPromptTokens(
+        for decodingClient: any WhisperDecodingClient,
+        at vocabularyPath: String,
+    ) -> [Int]? { // swiftlint:disable:this discouraged_optional_collection
+        guard let revision = WhisperVocabularyPrompt.fileRevision(at: vocabularyPath) else {
```
pasrom, line 385: `fileRevision` runs `FileManager.attributesOfItem` on every call, and this is reached from both `transcribeSegments` and `transcribeSamples`. Live captions drive `transcribeSamples` once per window, so with captions on the app stats the vocabulary file continuously on the main actor for the length of the meeting. The cache spares the content read but never the stat.

On a cache miss the cost is larger: `WhisperVocabularyPrompt.tokens` only breaks on `promptTokens.count == tokenBudget`, an exact hit, so with a large file whose budget is never exactly filled it calls `tokenizer.encode` once per term, up to 2000 times synchronously, in the middle of a transcription.

Related to the summary's first point, the live path also pays the prompt in decoder passes: the prompt is force-fed one token per inference, so a full budget adds roughly 97 serial decoder predictions to every partial. I have not measured where that becomes visible, but it is the path with the tightest latency budget in the app.

A cheaper shape would be to compute the tokens when the setting changes rather than per decode, since `didSet` already invalidates on all three inputs.

app/MeetingTranscriber/Sources/Settings/TranscriptionSettingsView.swift
```diff
@@ -51,20 +51,44 @@ struct TranscriptionSettingsView: View {
...
+                Text(settings.customVocabularyValidation.message)
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
+                    .onChange(of: settings.customVocabularyPath) { _, _ in
+                        settings.refreshCustomVocabularyValidation()
+                    }
+
+                VStack(alignment: .leading, spacing: 4) {
+                    Text("Canonical terminology")
+                    TextEditor(text: $settings.terminologyRulesText)
```
pasrom, line 95: Per the GUI testing section of `CLAUDE.md`, each new control wants an `.accessibilityIdentifier` from the shared `A11yID` namespace plus one ViewInspector wiring test that drives it and asserts the `AppSettings` write-back. `A11yID.swift` currently has no vocabulary or terminology constants, and neither this `TextEditor` nor the rewired vocabulary `TextField` has a test.

The `TextField` is the one I would prioritise: its binding now routes through `setCustomVocabularyPath`, which clears the bookmark as a side effect, and nothing pins that behaviour.

Separately, `testVocabularyHelpExplainsEachEngineAndLiveCaptionLimit` compares `vocabularyHelpText(for:)` against a literal copy of the same string. It cannot fail unless someone edits both, and it would stay green if the `.help` modifier were removed from the view entirely. Asserting through the view that the help text is attached would test something.

pasrom (changes requested):
Thanks for this, and sorry for the long review. The plumbing is genuinely good: the prompt does reach the decoder (`usePrefillPrompt` defaults to true and `promptTokens` is prefilled per window), the security scope correctly brackets every read, the token cache is main-actor confined and invalidates on all three inputs, and the terminology stage does not disturb any downstream timestamps. I checked all of that against the pinned WhisperKit 1.1.0 rather than the stale checkout in `.build`.

Two findings I would like resolved before merge, both silent: nothing fails, nothing is logged, and every gate stays green.

**1. The 96-token prompt costs about 43% of the decoder context, and a window that runs out is not retried.**

`Constants.maxTokenContext` is `Int(448 / 2)` = 224, and it covers the prefill *and* the generated tokens: the loop stops at `currentTokens.count >= maxTokenContext - 1`, where `currentTokens` starts as `decoderInputs.initialPrompt`. Without a vocabulary the prefill is 4 tokens, so roughly 219 tokens are available per 30s window. With a full budget it is `startOfPrev` + 96 + 4 = 101, leaving roughly 122.

What happens at the cap is the part that matters. The sampled token is not appended (the append sits in the non-completed branch), `finalize` adds EOT, and `filteredTokens` spans SOT..EOT, so the token stream ends `[text, text, EOT]`. That makes `isTimestampToken.suffix(3) == [false, false, false]`, i.e. `noTimestampEnding`, and `SegmentSeeker` then takes `else { seek += segmentSize }` and advances the full window. The audio between the truncation point and the window end is never decoded.

Fixture density from this repo: `two_speakers_de` is 28 words in 17.0s = 49 words per 30s, roughly 80 tokens, which never caps. `four_speakers_en_ami` is 333 words in 93.8s = 106 words per 30s, roughly 140 tokens, which sits between 122 and 219. So the prompt moves dense recordings from "never caps" to "caps routinely", and it does so on the most content-heavy stretches.

To be fair to the PR: this truncation behaviour is WhisperKit's and exists without this change. What the change does is halve the budget and make it reachable. So the fix is likely not architectural: a much smaller default budget plus a measurement would do it. What I would want to see is not just a cap-hit count but an assertion that prompting introduces no truncated window tails and no extra deletions on a dense fixture.

**2. An expansion rule rewrites text that was already correct.**

`parse` rejects a variant only when it is case-insensitively *equal* to the canonical, so a variant *contained in* its canonical survives. With `Kubernetes Cluster => Cluster`, the standalone word inside an already-correct phrase is word-bounded on both sides and matches:

```
in : der Kubernetes Cluster laeuft stabil
out: der Kubernetes Kubernetes Cluster laeuft stabil
```

It also compounds on re-application. I reproduced this by porting `parse`/`normalize` line by line; the UI documents the format as `Canonical spelling => spoken variant`, so this is a normal expansion rule, not malformed input. Rejecting every contained variant would be the wrong fix since it kills useful expansion rules; protecting existing canonical occurrences (for example by making the canonical the first alternative in the rule's regex and treating that match as a no-op) keeps them working.

**On testing.** The suite is thorough on the pure types, but nothing currently falsifies "the feature does nothing". The real-model test asserts only that WER did not get worse, and the terminology stage is never exercised through a `PipelineQueue`. The PR description mentions pipeline tests covering post-ASR terminology normalization; I could not find one on this branch, so that item may have been lost in a rebase. Details are in the inline comments.

The remaining inline comments are smaller: a validation status that promises more than either engine delivers, a Parakeet path that ignores mid-session changes, a stale-bookmark repair that probably cannot succeed under sandbox, a test seam that changes a production predicate, and the missing accessibility identifiers the repo's GUI testing rules ask for.

niftan5:
Thank you very much for the detailed review and for taking the time to trace the edge cases so carefully. The feedback was very helpful, especially around the shared Whisper prompt/generation budget, terminology rule semantics, lifecycle handling, and the distinction between “wired up” and meaningfully tested behaviour.

I have addressed the points as follows:

### WhisperKit prompt budget and quality coverage

- Reduced the default vocabulary prompt budget from 96 to 32 tokens.

  WhisperKit 1.1.0 shares its 224-token decoder context between prompt prefill and generated speech. The smaller bound deliberately leaves at least 180 tokens for generation after control tokens, so a vocabulary hint cannot crowd out dense speech windows.

- Added a dense real-meeting quality scenario for the vocabulary path.

  The test compares prompted and unprompted decoding on the four-speaker fixture and checks:
  - prompted WER does not regress beyond the allowed tolerance;
  - the prompt does not introduce additional deletions;
  - decoding reaches the recording tail;
  - the scenario really uses the full configured vocabulary-token budget.

- Kept the vocabulary format bounded and deterministic:
  - file-size, term-count, and term-length limits;
  - case-insensitive duplicate removal while preserving the first occurrence and file order;
  - complete terms only — no token-level truncation;
  - earlier terms retain priority when the prompt budget is full.

### Parakeet vocabulary lifecycle

- Reworked Parakeet’s vocabulary refresh to be revision-aware rather than path-only.

  An edit to an already selected vocabulary file now creates a new configuration identity based on path, bookmark, and file revision. This means the updated terms are prepared before the next decode rather than leaving an old CTC booster active.

- Removed the unstructured background refresh race.

  A settings change invalidates the existing booster immediately. The next transcription awaits preparation of the current vocabulary configuration. Generation and revision checks prevent an older asynchronous CTC preparation from being adopted after the user has selected or edited a newer file.

- Kept vocabulary failures local to the optional enhancement.

  If CTC vocabulary setup fails, the loaded ASR model remains available. A vocabulary problem must not turn an otherwise usable Parakeet model back into an unloaded engine state.

- Parakeet continues to use the same validated, bounded and deduplicated term list as WhisperKit before handing it to FluidAudio.

### Terminology normalisation

- Tightened the canonicalisation semantics.

  Rules use the form:

  `Canonical spelling => spoken variant | another variant`

  Matching remains case-insensitive and limited to complete words or phrases. The canonical form is considered before variants, so an already correct canonical phrase is protected from a shorter contained variant. Casing-only corrections are supported as well.

- Kept rule processing bounded and observable.

  Invalid lines are ignored with active/ignored-rule feedback, and text/rule/term sizes are bounded before regex compilation.

- Made duplicate handling locale-stable.

  Vocabulary and terminology duplicate detection now both use the same POSIX case-folding policy rather than depending on the local macOS UI locale.

- Preserved the terminology snapshot for the complete job.

  The same normalizer used for the original transcription is now carried into the defensive re-transcription fallback, so a settings edit during processing cannot produce mixed terminology results within one job.

### Bookmarking, UI, and testability

- Retained the security-scoped bookmark handling and made the vocabulary path/bookmark pair flow through one settings mutation path.

- The UI now explains the WhisperKit limitation explicitly:
  - it is a soft decoder hint, not guaranteed correction;
  - it uses a 32-token budget;
  - earlier entries have priority once the budget is full.

- Added UI coverage for the actual `.help(...)` attachment, the visible capacity note, vocabulary path write-back, and stable accessibility identifiers.

- Kept the WhisperKit decoding seam focused on tests while ensuring production loaded-state behaviour is based on the real `pipe`.

### Verification

I ran:

- `./scripts/pre-push.sh`
- `./scripts/pre-push.sh --with-appstore`
- the focused vocabulary, terminology, pipeline, and settings regression tests
- lint and release builds

Both Homebrew and App Store parity checks passed.

Thank you again for the careful review. Please let me know if you see another edge case or if you would prefer any part of the lifecycle handling to be structured differently.

niftan5:
Follow-up to the **Codecov** report:

I have rebased the branch onto the current `main` and added focused coverage for the remaining vocabulary boundaries:

- Parakeet vocabulary preparation is now exercised without downloading the production CTC models. The tests verify validated, deduplicated term hand-off, error propagation, and refresh after an existing vocabulary file changes.
- Vocabulary validation now covers excessive term counts and overlong terms.
- Terminology normalization now covers its text-size, rule-count, and term-size limits.

The test boundary only replaces the model-dependent CTC preparation during unit tests. The production configuration, refresh, and rescoring paths remain unchanged.

I used the Codecov report as guidance to add behavioural coverage, without changing coverage thresholds or excluding source files.

Verified locally:

- full Swift test suite (excluding the visual snapshot test)
- SwiftFormat and SwiftLint
- release builds for Homebrew and App Store variants

app/MeetingTranscriber/Sources/WhisperVocabularyPrompt.swift
```diff
@@ -0,0 +1,160 @@
...
+    ) -> [Int] {
+        guard tokenBudget > 0 else { return [] }
+
+        var promptTokens: [Int] = []
+        for term in terms.prefix(maximumTermCount) {
+            if promptTokens.count == tokenBudget { break }
+
+            let termTokens = tokenize(" " + term)
+            guard !termTokens.isEmpty,
+                  termTokens.allSatisfy({ $0 < specialTokenBegin }),
```
pasrom: Blocker, full trace in the summary.

This guard rejects every term in production. `tokenize` is `tokenizer.encode(text:)`, which adds special tokens, so `termTokens` is always `[50258, 50364, ...content..., 50257]` while `specialTokenBegin` is 50257. `allSatisfy { $0 < 50257 }` can therefore never hold, `promptTokens` stays empty, and the caller turns that into `nil`.

Two ways out. Encode without the wrapper tokens, if the tokenizer type exposes the `addSpecialTokens: false` overload through `WhisperTokenizer`. Or keep `encode(text:)` and filter instead of reject, which is what WhisperKit itself does with the same values (`Array(promptTokens.suffix(maxPromptLen)).filter { $0 < tokenizer.specialTokens.specialTokenBegin }`). Filtering also keeps the intent of the guard, which is that no special token reaches the prompt.

Whichever you pick, the budget accounting changes with it, since terms will suddenly consume their real token counts for the first time. Worth re-running the dense-fixture quality test right after, because it is the only thing here that can see the difference.

app/MeetingTranscriber/Sources/WhisperKitEngine.swift
```diff
@@ -268,16 +332,81 @@ final class WhisperKitEngine: TranscribingEngine, StreamingTranscribingEngine {
...
+        guard case let .loaded(terms) = loadResult,
+              let tokenizer = decodingClient.tokenizer
+        else {
+            vocabularyPromptCache.storeNoPrompt(for: key)
+            return nil
+        }
+
+        let tokens = WhisperVocabularyPrompt.tokens(
+            for: terms,
+            tokenize: tokenizer.encode(text:),
```
pasrom, line 420: This is the call site that makes the guard above unsatisfiable. `WhisperTokenizerWrapper.encode(text:)` forwards to the underlying `encode(text:)`, which defaults `addSpecialTokens` to true and runs the model's `TemplateProcessing` post-processor.

Nothing is wrong with the line in isolation, it is the pairing with the reject-on-any-special-token guard that produces the no-op. Fixing either side is enough, but this is the cheaper place if the `addSpecialTokens:` overload is reachable from the `WhisperTokenizer` protocol.

Also worth a thought: `vocabularyPromptTokenCount` is currently the only production signal that would reveal this, and only inside a gated test. A single `logger.info` with the term and token counts when a prompt is built, or a debug-level line when a term is rejected, would have turned this into a five-minute diagnosis.

app/MeetingTranscriber/Sources/TerminologyNormalizer.swift
```diff
@@ -0,0 +1,153 @@
...
+                .sorted { lhs, rhs in lhs.count > rhs.count }
+            guard !variants.isEmpty else {
+                ignoredLineCount += 1
+                continue
+            }
+
+            // Canonical comes first. A canonical phrase therefore occupies its
+            // entire range before a shorter expansion can match inside it; the
+            // replacement is a no-op except for a requested casing correction.
+            let alternatives = ([canonical] + variants)
```
pasrom: Blocker, and it is the mirror of the bug this line fixes.

Prepending the canonical unconditionally protects a correct canonical phrase, which was the point and it works. But `NSRegularExpression` alternation is leftmost-first, so when the canonical is itself a word-bounded prefix of a longer variant, the canonical wins the match and the longer variant is never tried.

```
rule: Kubernetes => Kubernetes Cluster
in  : der Kubernetes Cluster laeuft
out : der Kubernetes Cluster laeuft     (unchanged, rule inert)
```

"Kubernetes" matches at the same start, the following space satisfies `(?![\p{L}\p{N}])`, and replacing it with the canonical is a no-op. Neither the variant sort nor `nextAvailableLocation` can help, because the engine never emits the longer match.

Sorting canonical and variants together longest-first, rather than prepending, handles both directions:

```swift
let alternatives = ([canonical] + variants)
    .sorted { $0.count > $1.count }
    .map(NSRegularExpression.escapedPattern(for:))
    .joined(separator: "|")
```

I checked that ordering against four cases and all behave: contraction works, the already-correct expansion case stays protected, plain expansion still fires, and casing-only still corrects. Note the `variants` sort above then becomes redundant.

Worth a regression test per direction. `TerminologyNormalizerTests` currently covers whole-word matching, multi-word variants, malformed lines, non-cascading and the new bounds, but no contraction case.

app/MeetingTranscriber/Sources/ParakeetEngine.swift
```diff
@@ -90,6 +122,7 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
 
     func transcribeSegments(audioPath: URL) async throws -> [TimestampedSegment] {
         try await ensureModel()
+        try await ensureVocabularyConfiguration()
```
pasrom: This `try` is uncaught, and `loadModel()` treats the very same call as optional (`do { try await ensureVocabularyConfiguration() } catch { logger.warning(...) }`) with a comment explaining why. The decode path should say the same thing.

Two paths inside actually throw: the temp-file write of the prepared term list, and the rethrow after `VocabularyRescorer.create`. Neither is about the audio.

Consequence in batch: `processNext` catches and marks the job `.error`, and because the throw happens before `engine.transcribeSegments`, there is no transcript to fall back on. A full meeting recording is lost to a failed optional enhancement. That is worse than the pre-PR behaviour, where a vocabulary problem could only cost the boosting.

Wrapping this in the same do/catch as `loadModel()` is enough. The recording then transcribes unboosted, which is exactly what the vocabulary is documented to be: optional.

app/MeetingTranscriber/Sources/ParakeetEngine.swift
```diff
@@ -123,6 +156,7 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
     /// can be stateless across VAD segments.
     func transcribeSamples(_ samples: [Float]) async throws -> String {
         try await ensureModel()
+        try await ensureVocabularyConfiguration()
```
pasrom: This await can go away entirely, which would also remove the two problems above from the live path at no functional cost.

Live Parakeet does not apply the booster. There is no rescoring call in this method, and the Settings help text says so explicitly ("Live captions do not use CTC vocabulary rescoring"). So a live session currently pays for a CTC model download, two CoreML loads, the retry churn described below, and the throw exposure, for a booster it then does not consult.

The user-visible version: someone selects a vocabulary file mid-meeting, the next caption window blocks on a model download with no progress indication, and the captions it eventually produces are identical to what they would have been. The inline load also sidesteps `ModelWarmupQueue`, which exists to stop exactly this kind of concurrent model loading.

app/MeetingTranscriber/Sources/ParakeetEngine.swift
```diff
@@ -184,38 +218,179 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
...
         let ctcModels: CtcModels
         do {
             (vocab, ctcModels) = try await CustomVocabularyContext.loadWithCtcTokens(
-                from: path,
+                from: preparedVocabularyURL.path,
                 ctcVariant: .ctc110m,
             )
         } catch {
-            logger.warning("Parakeet: failed to load vocabulary from \(path): \(error.localizedDescription, privacy: .public)")
+            logger.warning("Parakeet: failed to load vocabulary from \(configuration.path): \(error.localizedDescription, privacy: .public)")
```
pasrom: Two problems in this catch, and the second one produces wrong output rather than slow output.

**No memoization.** It logs and returns without touching `configuredVocabulary`, so `ensureVocabularyConfiguration`'s `configuration != configuredVocabulary` guard is still true next time and the whole preparation is attempted again. On every `transcribeSegments`, and on every `transcribeSamples`, which is once per live-caption window. FluidAudio calls `downloadAndLoad` each attempt: a complete cache still means two fresh CoreML instantiations plus a tokenizer load, and an incomplete one means another network fetch. I understand the retry is deliberate, but it needs a backoff or an attempt cap per configuration.

**A superseded booster stays installed.** `invalidateVocabularyConfiguration` clears `vocabularyBooster`, but it only runs from the `didSet` of the path and the bookmark. Editing the selected file in place changes neither, so no `didSet` fires. `ensureVocabularyConfiguration` then notices the new revision, preparation fails here, and the previous revision's booster is still in place for the decode that follows. The recording is rescored with the terms the user just removed, and nothing says so.

Clearing the booster before attempting a new configuration, or on this failure path, would make the failure mode "no boosting" rather than "boosting with stale terms".

app/MeetingTranscriber/Sources/ParakeetEngine.swift
```diff
@@ -184,38 +218,179 @@ final class ParakeetEngine: TranscribingEngine, StreamingTranscribingEngine {
...
+    func installVocabularyPreparationForTesting(
+        _ prepare: @escaping ([String]) async throws -> Void,
+    ) {
+        vocabularyPreparationOverride = prepare
+    }
+
+    /// Ensures the exact current file revision is ready before a decode. This
+    /// makes an edit to an existing file take effect on the next transcription,
+    /// and avoids an unstructured background task racing the decode.
+    func ensureVocabularyConfiguration() async throws {
```
pasrom: Small one, on the new tests rather than against them.

`ensureVocabularyConfiguration` went from `private` to internal purely so `ParakeetVocabularyPreparationTests` can call it, and `installVocabularyPreparationForTesting` is a second production hook that exists only for tests. This is the same pattern we resolved on the WhisperKit side, reappearing here. Since the test target uses `@testable import`, driving this through `customVocabularyPath` and a public entry point would keep the seam narrower.

On the tests themselves: `testVocabularyPreparationReceivesValidatedDeduplicatedTerms` and the revision-refresh test are good, they pin real behaviour. `testVocabularyPreparationFailureReachesTheCaller` is worth a second look though, because it pins propagation out of `configureVocabulary`, and the fix for the decode-path issue above is to stop that propagation from reaching the caller there. The test will still pass, since it drives the configure entry point directly, but the two intentions read as contradictory side by side and a future reader will have to work out which one is load-bearing.

Note also that the override short-circuits before the temp-file write and the CTC load, so the three new tests never exercise the two paths that actually throw in production.

pasrom (changes requested):
Thanks for the thorough rework. I went through it against the code rather than the description, and nine of the twelve threads are genuinely resolved, so I have marked those resolved: the terminology snapshot, the pipeline test, the falsifying quality assertion, the validation ceiling and the Parakeet bypass, the bookmark minting inside the security scope, the resolve fallback and `private(set)` path, the mid-session Parakeet refresh, the production `pipe` predicate, and the accessibility identifiers with real wiring tests. The new boundary tests are a good addition too.

Three threads stay open and there are two new blockers. One of them is mine to own.

**1. The WhisperKit vocabulary prompt is a no-op in production, and it always was.**

This is in the code I reviewed the first time and I missed it. The chain:

`WhisperKitEngine` passes `tokenize: tokenizer.encode(text:)`. `WhisperTokenizerWrapper.encode(text:)` forwards to the underlying tokenizer, whose `encode(text:)` is defined as `encode(text:addSpecialTokens: true)`. The whisper-large-v3 `tokenizer.json` carries a `TemplateProcessing` post-processor with the single template `<|startoftranscript|> <|notimestamps|> $A <|endoftext|>`. Those ids are 50258, 50364 and 50257, and `specialTokenBegin` is `convertTokenToId("<|endoftext|>")`, i.e. 50257.

So every term encodes as `[50258, 50364, ...content..., 50257]`, and the guard `termTokens.allSatisfy({ $0 < specialTokenBegin })` rejects all of them, always. `promptTokens` ends up empty, becomes `nil`, and `DecodingOptions.promptTokens` is never set. No log line, and Settings still reports the file as ready with a 32-token hint.

Nothing catches it because every unit test injects a mock tokenizer emitting small ids. The one assertion that would is your new `XCTAssertEqual(hinted.vocabularyPromptTokenCount, WhisperVocabularyPrompt.defaultTokenBudget)`, and it sits behind `RUN_QUALITY_TESTS=1`, so it has never run. It should currently fail with 0 against 32, which is a good way to confirm this before and after the fix.

A consequence worth stating: the context-budget concern from my first review was never reachable in production, precisely because the prompt was always nil. Lowering the budget to 32 and adding the dense-fixture coverage test is the right change, but it stays unproven until the prompt actually reaches the decoder and that lane runs once. That is why I am leaving the budget thread open rather than resolving it.

**2. The terminology fix closed the expansion hole and opened the mirror one.**

Putting the canonical unconditionally first fixes `Kubernetes Cluster => Cluster` on already-correct text, which I verified. But the alternation is leftmost-first, so when the canonical is a boundary-terminated prefix of a longer variant, the canonical matches first and the longer variant is never considered. `Kubernetes => Kubernetes Cluster` on "der Kubernetes Cluster laeuft" matches only "Kubernetes", replaces it with itself, and the contraction silently does nothing.

Sorting the canonical together with the variants longest-first instead of prepending it satisfies both directions. I checked expansion, contraction, the already-correct case and casing-only against that ordering and all four behave.

The rest is the Parakeet lifecycle, inline below. The direction is right; the issue is that an optional enhancement can now fail a whole job, retry without bound, and in one case rescore with terms the user just deleted.

app/MeetingTranscriber/Sources/WhisperVocabularyPrompt.swift
pasrom: Leaving this open. Lowering the budget to 32 and adding the dense-fixture coverage test is exactly the right change, and I would resolve it on that basis alone. The reason I cannot yet is that the prompt never reaches the decoder at all, so neither the old 96 nor the new 32 has ever been exercised in production, and the test that would show it has not run.

Once the tokenizer issue is fixed, one run of the quality lane settles this thread in either direction. If `vocabularyPromptTokenCount` reports 32 and the deletion and tail assertions hold on the dense fixture, this is done and I will close it.

app/MeetingTranscriber/Sources/TerminologyNormalizer.swift
pasrom: Leaving this open, because the fix is half of the story. The duplication this thread was about is genuinely gone and the casing-only case now works, both of which I verified. But the canonical-first ordering makes a contraction rule inert, which is the same defect pointed the other way. Details and a suggested ordering are in the new review, anchored on the line that builds the alternation.

app/MeetingTranscriber/Sources/WhisperKitEngine.swift
pasrom: Leaving this open, though it is much better than it was. The per-decode `attributesOfItem` is unchanged, so a live session still stats the vocabulary file once per caption window on the main actor. The prefill cost is down from roughly 97 extra decoder predictions per partial to roughly 33 with the smaller budget, which I no longer consider blocking, but it is a reduction rather than a fix.

Worth noting that the same shape now exists on the Parakeet side, where `ensureVocabularyConfiguration` resolves the bookmark and stats the file on every decode as well. There the cheaper answer is to drop the live call entirely, see the new review.

app/MeetingTranscriber/Sources/TerminologyNormalizer.swift
pasrom: Tightening my own wording above, because the scope matters for the fix.

"rule inert" in that code block is a notch too strong. The rule still corrects the casing of the canonical prefix, so `"kubernetes Cluster"` does become `"Kubernetes Cluster"`, exactly as the comment at lines 131-133 intends. What never happens is the contraction itself.

The defect is also narrower than my comment reads. It bites only when the canonical is a word-bounded prefix of a longer variant, where the boundary is a space, a hyphen, or any other non letter-or-digit character. A canonical that is short but not a prefix is fine today: `K8s => Kubernetes Cluster` contracts correctly as written.

So the regression test that belongs next to it is specifically the prefix case, something like `Kubernetes => Kubernetes Cluster` asserting that `"der Kubernetes Cluster laeuft"` becomes `"der Kubernetes laeuft"`. That is the one the current ordering fails and the longest-first ordering passes.

app/MeetingTranscriber/Tests/CustomVocabularyTests.swift
```diff
@@ -39,19 +39,120 @@ final class CustomVocabularyTests: XCTestCase {
...
+            .appendingPathComponent("vocabulary-empty-\(UUID().uuidString).txt")
+        try "\n  \n".write(to: file, atomically: true, encoding: .utf8)
+        defer { try? FileManager.default.removeItem(at: file) }
+
+        settings.setCustomVocabularyFile(file)
+
+        XCTAssertEqual(settings.customVocabularyValidation, .empty)
+    }
+
+    func testVocabularyValidationRejectsTooManyTerms() throws {
```
pasrom: Good that the limits are pinned now. One gap while you are in here: these two validation tests, and `testOverlongTermsAreReportedAsIgnored` next door, only pin the reject side of each bound.

Concretely: change `<=` to `<` in the `maximumTermCount` guard, and a file with exactly the advertised maximum number of terms is rejected while every test added in this commit stays green. The user-visible promise is the boundary itself, so it is the accept side that carries the contract.

`testRuleLimitReportsFurtherRulesAsIgnored` is the shape to copy. It pins both halves at once, the 200 active rules and the 201st being ignored, so an off-by-one in either direction turns it red.

Also untested: the overlong-canonical branch of the rule parse. Only the overlong-variant branch has a case.

niftan5:
Thank you again for the very careful review and for tracing the WhisperKit tokenizer behaviour back to the pinned implementation. That was particularly valuable here: it showed that the original implementation was not merely weakly conditioned, but effectively a production no-op because `encode(text:)` adds wrapper/control tokens which the prompt builder rejected.

I revisited the WhisperKit approach rather than treating the smaller prompt budget as sufficient on its own.

### WhisperKit vocabulary prompt

The current implementation now removes tokenizer wrapper/control tokens before accounting for the vocabulary budget. This means that, when enabled, the vocabulary produces actual content tokens for the real WhisperKit decoder rather than silently becoming an empty prompt.

The real-model test was run locally with `RUN_QUALITY_TESTS=1`; it verifies that the enabled prompt reaches the real decoder with a non-zero content-token count within the configured 32-token budget.

However, the dense-fixture measurements also showed an important limitation: even small static glossary prompts can reduce transcript completeness on WhisperKit. I therefore changed the product behaviour rather than weakening the normal WhisperKit quality guard:

- The WhisperKit vocabulary prompt is now explicitly experimental and disabled by default.
- While disabled, WhisperKit does not read or tokenize the vocabulary file.
- Users can opt in through `Use custom vocabulary prompt (experimental)`.
- The setting is persisted, reactively propagated to the engine, covered by UI interaction tests, and explained with a tooltip.
- The UI explicitly states that the feature may reduce transcription completeness and recommends Parakeet for vocabulary boosting.
- The option applies consistently to saved transcription and to WhisperKit-backed live transcription.

This keeps the normal WhisperKit path unprompted and protected by its established quality baselines, while still giving users a deliberate way to try vocabulary conditioning where their terminology is more important than the known completeness trade-off. If WhisperKit later provides a better conditioning API, the implementation can be replaced behind the same setting.

### Terminology and Parakeet follow-ups

The remaining review follow-ups are also addressed:

- Terminology alternatives are ordered as one longest-first list containing both canonical spelling and variants. This covers expansion, contraction, already-correct canonical text, and casing-only normalization.
- Parakeet vocabulary boosting is now best-effort: preparation failures do not fail a batch transcription, a failed configuration does not retry on every decode, and a changed file revision can be prepared again.
- Parakeet CTC preparation is excluded from model preload and live sample transcription.
- Literal boundary tests now pin the accepted and rejected limits for vocabulary terms, terminology rules, canonical terms, and variants.

### Verification

I verified the updated branch with:

- full Swift test suite: 2,930 tests;
- focused vocabulary, terminology, engine-sync, and settings interaction tests;
- the gated real-model WhisperKit decoder test;
- SwiftFormat and SwiftLint;
- `./scripts/pre-push.sh --with-appstore` for both Homebrew and App Store release builds.

Thank you again for the detailed guidance. I would be very grateful for another look at the revised WhisperKit opt-in approach and the remaining lifecycle changes.

pasrom (dismissed):
Approving. Everything raised across both reviews is addressed, and I checked each one against the code rather than the description.

- The prompt now reaches the decoder: `filter { $0 < specialTokenBegin }` strips the tokenizer's wrapper tokens instead of rejecting the term, which is what WhisperKit does internally with the same values.
- Terminology matches longest-first across canonical and variants, so both directions work. The stable tiebreak on configured order is a nicer touch than I suggested, since it keeps the compiled pattern deterministic.
- Parakeet vocabulary preparation is genuinely best effort now: non-throwing at the decode site, a terminal `.failed` state per file revision so a broken CTC setup is not retried per window, the superseded booster cleared before the new attempt, and the live path no longer waits on a booster it never consults.
- The limit tests pin both sides of every bound, including the canonical and variant branches, and the term ceiling at 256 is a much more honest number next to a 32-token budget.
- The debug-gated seams keep the test hooks out of release builds.

Making the WhisperKit prompt an explicit experimental opt-in, off by default, is a better answer to the context-budget question than a number would have been. It puts the trade-off in front of the person who chooses it instead of hiding it behind a default.

Two things I am not blocking on, both worth a look before or after merge.

**The completeness trade-off is now shipped unmeasured.** The dense-fixture test was replaced rather than extended, so the WER, added-deletions and tail-coverage assertions are gone. The reasoning in the new test's comment is sound as a gating decision: an opted-in user accepts the trade-off, so failing CI on it would be wrong. But nobody has measured how large it is, and the wording follows from that. "May reduce transcription completeness" reads like degraded wording, whereas the actual mechanism is that a window which exhausts the decoder context advances a full 30 seconds and never decodes its tail, so whole sentences can be missing with nothing indicating it. Keeping those three assertions as a reported measurement rather than a gate, or running the lane once and putting a real figure in the help text, would let someone make that choice informed. Right now the toggle asks them to accept a risk whose size is unknown to everyone including us.

**Some unrelated documentation was removed.** `AppSettings.swift` lost the rationale on `parakeetLanguage` (why the default is empty, and the reference to the issue where a concrete default hurt non-German users) and the explanation on `activeEngineLanguageOrNil` of how it selects the live-caption backend. Neither is recoverable from the code, and this PR otherwise only adds two properties to that file. Worth restoring in a follow-up.

Note that no CI run exists on this head yet, so the approval is on the code as read. Worth waiting for green before merging.

Thanks for the patience through four rounds of this. The lifecycle handling in particular ended up in a much better place than where it started.

niftan5:
Thank you again for the careful follow-up feedback — I’ve addressed both points in two focused commits.

1. **Make the WhisperKit trade-off measurable and visible**

   The experimental WhisperKit prompt now discloses the measured dense-fixture result directly in its help text: the 25-content-token prompt (within the 32-token budget) increased WER from 29% to 77% and deletions from 73 to 241 in the four-speaker English evaluation. The visible caption also makes clear that dense audio can omit whole sentences and recommends Parakeet for vocabulary boosting.

I also changed the gated production-model test to compare the unprompted and prompted runs and report:
- WER
- deletion count
- end-of-audio coverage
- the prompt’s actual content-token count

These are deliberately reported measurements rather than a CI threshold for the explicit experimental opt-in. The normal unprompted WhisperKit quality baselines remain the release-relevant regression guard.

2. **Restore the AppSettings rationale**

I restored the detailed documentation for Parakeet’s language hint and the active-engine language routing. To stay below the file-length limit without removing the rationale, I moved the derived language properties into `AppSettings+Computed.swift`; behavior is unchanged and the existing settings tests continue to cover the routing.

Validation completed locally:
- full Swift test suite
- SwiftFormat and SwiftLint
- Homebrew release build
- App Store release build

Thanks again — the feedback made the experimental nature and the retained design rationale much clearer.

pasrom:
One last thing before merge, about the commit history rather than the code. The approval stands either way.

This repo is rebase-only:

```
allow_merge_commit: false   allow_squash_merge: false   allow_rebase_merge: true
```

So all eleven commits land on `main` verbatim. There is no squash to absorb the review round afterwards, which means what is in the list now is what `git log` and `git blame` show permanently.

Three things I would like to avoid putting there:

1. **Four `fix` commits repair code that never existed outside this PR.** `harden vocabulary review paths`, `harden vocabulary refresh lifecycle`, `match longer terminology phrases first`, `make Parakeet vocabulary boosting best effort`. On `main` those read as fixes for defects no user ever saw.
2. **Two `feat` commits for the same feature, one word apart in the title.** `add WhisperKit vocabulary prompts` and `add experimental WhisperKit vocabulary prompt`. The second changes the first's semantics to opt-in and repairs its central defect, so it is a fixup of the first in everything but the name. A reader cannot tell which is authoritative.
3. **`restore transcription setting rationale` undoes what the experimental-prompt commit did.** That commit shortened the doc comments on `parakeetLanguage` and `activeEngineLanguageOrNil` to stay under the `file_length` limit; the later one solves the same constraint properly by extracting to `AppSettings+Computed.swift` and restores the text. In a folded history the extraction happens once, in the commit that makes it necessary, and nothing is ever shortened.

Also worth noting: `cover vocabulary preparation boundaries` is labelled `test` but changes `ParakeetEngine.swift` by 42 lines to add a seam, and `restore transcription setting rationale` is labelled `docs` but adds a source file and moves code.

## A shape that would work

Four commits, along seams the code already has:

**`feat(app): validate and share a custom vocabulary file`** — `VocabularyFileAccess`, `AppSettings+Vocabulary`, the path field and validation status line, the file and limit half of `WhisperVocabularyPrompt`, and their tests.

**`feat(app): apply custom vocabulary to Parakeet CTC boosting`** — `ParakeetEngine`, `ParakeetVocabularyConfiguration`, the Parakeet branch of `EngineController`, and the preparation and configuration tests.

**`feat(app): add experimental WhisperKit vocabulary prompt`** — `WhisperDecodingClient`, the token half of `WhisperVocabularyPrompt`, the engine path, the opt-in toggle, the `AppSettings+Computed` extraction with the comments intact, and the quality test. Introduced directly in its final opt-in form with the measured help text, so the intermediate state never appears.

**`feat(app): add canonical terminology normalization`** — `TerminologyNormalizer`, the pipeline wiring, the rules editor and its tests. This one is genuinely independent: `TerminologyNormalizer.swift` has no reference to `WhisperVocabularyPrompt`, `VocabularyFileAccess` or `customVocabulary`, and neither does the pipeline wiring.

## Mechanics, because interactive rebase will not do it

Eight files carry more than one of those themes: `AppSettings.swift`, `AppSettings+Computed.swift`, `A11yID.swift`, `TranscriptionSettingsView.swift`, `EngineController.swift`, `WhisperVocabularyPrompt.swift`, `CustomVocabularyTests.swift`, `EngineSettingsRuntimeSyncTests.swift`. Reordering and squashing cannot split one commit's content across two targets, and commits 2 to 4 each touch terminology and Parakeet and WhisperKit, so a reorder mostly produces conflicts.

Rebuilding from the final tree is simpler than it sounds:

```bash
git switch -c fold-651 origin/main
git checkout <current-head> -- <whole files for commit 1>
git restore --source=<current-head> -p -- <the shared files>
git commit -m "feat(app): validate and share a custom vocabulary file"
# repeat for the other three
```

Build after each commit so no intermediate state is broken, and finish with the only check that actually proves nothing was lost:

```bash
git diff <current-head> fold-651   # must be empty
```

## If that is more than you want to do

Folding the four `fix` commits and the last three into the two `feat` commits, without re-slicing any file, gets you to two or three commits. The themes stay a little mixed, but the review round disappears from the history, which is the part that matters. I would take that happily.

## Three unrelated changes worth separating while you are in there

- `scripts/generate_quality_fixtures.sh` gains `export LC_ALL=C`, which has nothing to do with vocabulary.
- The `AppSettings+Computed.swift` extraction is a refactor and would read better as its own `refactor(app)` commit, or inside the commit that forces it.
- The terminology normalization is effectively a second feature. Zero code coupling to the vocabulary work, its own file, pipeline stage, settings, limits and tests, and roughly a third of the diff, in a PR titled for the WhisperKit prompt. Not worth splitting now after this many rounds, but worth keeping in mind next time, and your own PR description already describes it as deliberately separate.

None of this is a criticism of the work, which ended up in good shape. It is only that rebase-only makes commit structure permanent, so it is worth a pass before the merge button.

niftan5:
Hi @pasrom,
thank you again for the detailed feedback and for explaining why the commit history matters in a rebase-only repository. I am still learning how to prepare PRs well in that workflow, and your concrete suggestion was very helpful.
I have now rebuilt the branch history and rebased it onto the current main. The implementation itself has not changed as part of this rewrite: I compared the previous and current five-commit series commit by commit, and each patch is unchanged. The vocabulary validation, Parakeet CTC boosting, experimental WhisperKit prompt, and canonical terminology normalization are now clearly separated; the locale-stability adjustment for quality fixtures is isolated in its own small commit.
I also reran the relevant local verification on the rebased branch: the 2,924 non-snapshot Swift tests, lint/format checks, and both Homebrew and App Store release builds passed.
The implementation and history rewrite are now complete. GitHub has created the fresh workflow runs for the rewritten branch; as this PR comes from my fork, they are currently awaiting maintainer approval before they can start.
Thank you again for your time and the careful review.

pasrom (approved):
Re-approving after the fold. The history now reads as four coherent feature commits plus the separated locale fix, and I verified the fold changed nothing: across all 31 files this PR touches, the diff between the pre-fold and post-fold heads is empty, so the content is identical and the rebase onto the newer base is the only difference.

CI is green on this head.

One thing for the record, which I am not asking you to fix at this point. `WhisperVocabularyPrompt.swift` is added in the third commit, but `AppSettings+Vocabulary.swift` already references it in the first, and `ParakeetEngine.swift` and `ParakeetVocabularyConfiguration.swift` in the second, so those two commits do not build on their own. The name is not declared in any form in either tree. That is my fault rather than yours: the plan I gave you said to split that file across the first and third commit and was too casual about it. The end state is correct, and the only cost is that `git bisect` cannot build across those two commits. Worth moving the file and its tests into the first commit if you happen to touch the branch again, but not worth another round on its own.

Thanks for working through six rounds of this. The lifecycle handling and the honest measurement in the help text both ended up somewhere better than where we started.

Issue #608, opened 2026-08-10
Add Slack Huddles to “Apps to Watch”


GiggleSamurai:
### App version

v0.7.0

## Feature request

Please add **Slack Huddles** to the macOS **Settings → General → Apps to Watch** list so Meeting Transcriber can automatically detect Slack meetings the same way it detects Teams, Zoom, Webex, FaceTime, etc.

## Current behavior

Slack is not available in the **Apps to Watch** list. A Slack Huddle can be recorded manually by selecting Slack, but it is not automatically detected.

## Expected behavior

- Add a **Slack** toggle under **Apps to Watch**.
- Detect when a Slack Huddle starts or becomes active.
- Use the same recording prompt / automatic recording behavior as supported meeting apps.
- Handle recording appropriately when the Huddle ends.

## Why this is useful

Slack Huddles are commonly used as lightweight meetings, especially for remote teams. Automatic detection would make the workflow consistent with Teams and Zoom.

Related issue: #422 discusses Slack manual recording echo behavior, but this request is specifically for **automatic Slack Huddle detection / Apps to Watch support**.

### What would you like it to do?

Add Slack Huddles to the Apps to Watch list so Meeting Transcriber can automatically detect when a Slack Huddle starts and handle recording the same way it does for Teams, Zoom, Webex, and other supported meeting apps.

Issue #562, opened 2026-07-31
Zoom detection in v0.5.1 fails because Zoom's assertion is named "Describe Activity Type"


duncanwood:
Documenting the root cause, since the fix landed in `v0.6.0-rc6` without the reason being written down — easy to regress.

**v0.5.1 cannot detect Zoom on any Mac.** `PowerAssertionDetector.matchAssertion` requires the assertion *name* to contain a keyword:

```swift
guard pattern.processNames.contains(processName) else { return false }
let lowerAssert = assertName.lowercased()
return pattern.keywords.contains { lowerAssert.contains($0.lowercased()) }
```

with `keywords: ["zoom"]`. But the only assertion Zoom publishes during a live meeting is:

```
pid 33847(zoom.us): NoDisplaySleepAssertion named: "Describe Activity Type"
```

`"describe activity type".contains("zoom")` is false, so the process name matches and the keyword never can. `"Describe Activity Type"` is the placeholder string from Apple's `IOPMAssertionCreateWithName` documentation — Zoom shipped the doc placeholder as its assertion name.

Verified on macOS 26.5.2 / Zoom 6.x by replicating `matchAssertion` against live assertions: `NO MATCH (process name OK, but AssertName lacks the required keyword)`.

`v0.6.0-rc6` fixes this via the `assertionTypes` fallback matching `NoDisplaySleepAssertion`, which is correct. Two suggestions:

1. Worth a note in the release notes that Zoom detection is broken in 0.5.1 — it presents as a permissions problem, and I spent a while on Screen Recording, TCC, and the Homebrew quarantine flag before reading the source.
2. The `MeetingDetector` docstring says *"Requires Screen Recording permission"*, and `AppState` wires up `PowerAssertionDetector` instead — so the window-title regexes and that permission are not actually load-bearing for detection. The docstring sends debugging in the wrong direction.

pasrom:
Both parts are now in stable. v0.6.0 (released today) carries the assertion-type fallback from rc6, so Zoom's placeholder assertion name no longer defeats detection. The documentation you flagged was wrong and is corrected in the same release (commit c225057f): the MeetingDetector docstring and the docs now state that auto-detection runs on IOKit power assertions and needs no Screen Recording, and that the grant only sharpens the meeting title and gates the fallback path for the audio capture permission, so a detection failure is never a Screen Recording problem. Thanks for writing the root cause down. Closing this once you confirm nothing is left, or close it yourself if you agree it is covered.

[...]

PR #653, opened 2026-08-24
fix(app): detect Webex calls via Cisco's "On a call" assertion

### LINKED ISSUE #652: Webex desktop client is never auto-detected (keyword-only pattern; Cisco names the assertion "On a call") — same failure mode as #562
## Summary

The Webex desktop client is not auto-detected on macOS. Recording has to be
started manually, which also means there is no automatic stop when the call
ends. Both symptoms have the same root cause: Webex's call-scoped power
assertion does not contain the keyword currently required by the detector.

Environment used to reproduce the issue:

- MeetingTranscriber 0.7.0
- macOS 26.x on Apple Silicon
- Webex 46.7.0.35472

## Why the current pattern cannot fire

`PowerAssertionDetector.defaultPatterns` contains:

```swift
AssertionPattern(
    appName: "Webex",
    processNames: ["Webex", "Cisco Webex Meetings", "Meeting Center"],
    keywords: ["webex"],          // assertionTypes: [] — no fallback
),
```

Webex is therefore matched only if the assertion name contains `webex`.

Zoom had the same failure mode in #562 (the assertion carried Apple's
placeholder `Describe Activity Type`) and received an `assertionTypes`
fallback in v0.6.0-rc6 via #475. Webex was deliberately excluded there
because of the risk of a display-sleep assertion being held while no call is
active.

There is no second detection path for the native Webex client:

- `MicInputDetector.defaultPatterns` does not include Webex.
- `MeetingDetector` contains Webex window-title patterns, but it is not wired
  into `WatchingController+Detectors.swift`, which composes only
  `PowerAssertionDetector` and `MicInputDetector`.
- Browser meeting detection does not apply to the native desktop client.

## Measurement: idle Webex client

With Webex running but no meeting active, `pmset -g assertions` showed no
assertion owned by Webex. Repeated checks after a call produced the same
result. The relevant assertions therefore appear to be call-scoped rather
than client-scoped for this Webex version.

## Measurement: active Webex meeting

During an active call, Webex held these two assertions:

```text
PreventUserIdleSystemSleep  named: "io avilable"
PreventUserIdleDisplaySleep named: "On a call"
```

The typo `io avilable` is produced by Webex.

The process name matches the configured `processNames`, but neither assertion
name contains `webex`, and the pattern has no `assertionTypes` fallback.
Detection is therefore impossible with the current pattern. MeetingTranscriber
remained running during the reproduction and did not emit a detection event.

## Suggested fix

```diff
 AssertionPattern(
     appName: "Webex",
     processNames: ["Webex", "Cisco Webex Meetings", "Meeting Center"],
-    keywords: ["webex"],
+    keywords: ["webex", "on a call"],
 ),
```

This mirrors the existing Teams approach: Teams matches `call in progress`,
while Webex needs `on a call`. It remains bound to Webex process names, so an
unrelated application using the same generic assertion name cannot trigger
the pattern.

`io avilable` is deliberately not used as a keyword because it is a typo that
may change in a future Webex release. The display-sleep assertion is also the
stronger signal of the two.

One remaining question is whether `On a call` is localized in non-English
Webex installations. If it is, an `assertionTypes` fallback may be more
robust; the idle measurements above suggest that this could still be safe for
the tested Webex version.

## Broader suggestion

This issue and #608 (Slack Huddles) show the same structural weakness:
detection depends on vendor-controlled assertion strings that may change at
any release.

The `pmset -g assertions` data also exposes call-related audio activity through
`coreaudiod` assertions. They identify the consuming process through
`kIOPMAssertionOnBehalfOfPID` (`Created for PID:`) and distinguish `audio-in`
from `audio-out` in the resources field. `PowerAssertionDetector` already
retrieves these dictionaries through `IOPMCopyAssertionsByProcess`; it does
not currently inspect those fields.

A detector using the `coreaudiod` owner, `OnBehalfOfPID`, and `audio-in` could
avoid vendor-specific strings without requiring audio capture, a new API, or
an additional entitlement. It would also distinguish meetings, which use
audio input, from ordinary playback, which is output-only.

That is a larger follow-up change. PR #653 contains the narrow keyword fix and
regression tests for this issue.

## Diagnostics note

`logUnmatchedWatchedAssertions`, the log entry that would expose this mismatch,
is emitted at `logger.info`. macOS does not persist INFO messages by default,
so the entry is normally visible only in a live `log stream` or after enabling
INFO persistence for the subsystem. Emitting this diagnostic at `.default`,
or implementing the persistent diagnostic log described in
`docs/plans/2026-05-05-persistent-diagnostic-log.md`, would make future
detection gaps easier to diagnose.

## Verification

- Reproduced with the native Webex desktop client while idle and during a call.
- Confirmed that the call assertions disappear when the call ends.
- Confirmed the current match logic cannot accept either assertion name.
- Added regression coverage in PR #653, including a process-name gating test
  for the generic `On a call` keyword.

DevUrandomFoo:
Fixes #652.

## What

The Webex desktop client is never auto-detected. Its `AssertionPattern` is
keyword-only on `"webex"`, but Cisco does not put the product name in the
assertion name. Measured with Webex 46.7.0.35472 on macOS 26.5.2 during an
instant meeting:

```
pid 870(Webex): [0x000402e000019b1b] 00:00:39 PreventUserIdleSystemSleep  named: "io avilable"
pid 870(Webex): [0x000402e200059b1c] 00:00:37 PreventUserIdleDisplaySleep named: "On a call"
```

`matches()` passes the bound `processNames` gate and then fails both remaining
tests — no `"webex"` in the name, and `assertionTypes` is empty — so detection
is impossible by construction. This is the #562 failure mode one app over
(Zoom carried Apple's `"Describe Activity Type"` placeholder).

It also explains a second symptom the reporter saw: with auto-detection dead
every recording is hand-started, and `ManualRecordingMonitorPolicy` only ends on
`.stopPidExited` / `.stopMaxDurationExceeded` — `WatchLoopEndPolicy.endGrace`
applies to auto-detected meetings only. One root cause, two symptoms.

## Why a keyword and not the `assertionTypes` fallback

#475 gave Zoom an `assertionTypes` fallback but deliberately kept Webex
keyword-only:

> Teams and Webex stay keyword-only, because Teams' WebView holds a
> display-sleep "Video Wake Lock" even with no call in progress.

Adding a keyword keeps that property, so the concern does not arise. The
pattern is also *bound*: `matches()` requires `processNames.contains(processName)`
first, so no foreign process can trigger it. On the measured machine
Amphetamine holds both display-sleep assertion types permanently and still
cannot false-positive — pinned by a counter-test.

`"io avilable"` is deliberately **not** used as a keyword: it is a Cisco typo a
future release will plausibly fix, and `PreventUserIdleSystemSleep` is the
weaker of the two signals.

## Tests

Four tests in `PowerAssertionDetectorPatternsTests`:

| test | pins |
|---|---|
| `testWebexCallWithCiscoAssertionNameIsDetected` | the measured assertion is detected |
| `testWebexKeywordIsBoundToTheWebexProcess` | `"On a call"` from Amphetamine / Chrome / an unknown process does **not** fire |
| `testWebexWithUnrelatedAssertionNameIsNotDetected` | a Webex client holding an unrelated assertion is not a meeting |
| `testWebexCallIsIgnoredWhenWebexIsNotWatched` | the "Apps to Watch" toggle still gates it |

Verified in that order, on the reporter's machine (Xcode 26.6, Swift 6.3.3,
macOS 26.5.2, Apple Silicon):

- **red first, deliberately** — with the tests added but the source unpatched:
  `Executed 18 tests, with 1 failure`, and the failure is
  `testWebexCallWithCiscoAssertionNameIsDetected`. The three guard tests pass
  before and after, since they pin behaviour the change must not alter.
- with the change: `Executed 78 tests, with 0 failures` across all
  `PowerAssertionDetector*` suites.
- wider regression (`--filter "Detector|Watching|Assertion|Settings|Pattern|Monitor"`):
  `Executed 484 tests, with 0 failures`.

`./scripts/lint.sh` was not run (SwiftLint/SwiftFormat unavailable in that
environment); the diff was checked by hand against `.swiftlint.yml` and
`.swiftformat` — no line exceeds 160 columns, the test file is 368 lines
(`file_length` 600), the class body 361 (`type_body_length` 400), and trailing
commas follow `--commas always`.

## Not verified

Whether `"On a call"` is localized in a non-English Webex UI. If it is, the
`assertionTypes` fallback becomes the more robust option after all — and a
measurement in #652 shows it would be safe for Webex: an **idle** Webex client
holds no assertion at all, unlike Teams' WebView.

## Follow-up, not in this PR

#652 also describes a detector that would remove this whole class of bug: the
same `IOPMCopyAssertionsByProcess` output already carries `coreaudiod`
assertions attributed to the consuming process via `kIOPMAssertionOnBehalfOfPID`
with `audio-in` / `audio-out` in the resources field. Keying on that is
vendor-string-independent, needs no new API or entitlement, and would cover
#608 (Slack Huddles) too. Happy to prepare it separately if you want it.

One small thing in the same area: `logUnmatchedWatchedAssertions` — the one line
that would have diagnosed this from the outside — is emitted at `logger.info`,
which macOS does not persist, so it is invisible after the fact. Emitting it at
`.default` would make future detection-gap reports self-diagnosing.

### DIFF
--- a/app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
+++ b/app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
@@ -298,7 +298,7 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
     /// the desktop client can never be auto-detected, which in turn means no
     /// auto-stop, because a hand-started recording only ends on
     /// `.stopPidExited` / `.stopMaxDurationExceeded`.
-    func testWebexCallWithCiscoAssertionNameIsDetected() {
+    func testWebexCallWithCiscoAssertionNameIsDetected() throws {
         let detector = PowerAssertionDetector(
             patterns: PowerAssertionDetector.patterns(watching: ["Webex"]),
             confirmationCount: 1,
@@ -307,9 +307,42 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
         detector.assertionProvider = {
             PowerAssertionFixture.assertionDict(processName: "Webex", assertName: "On a call")
         }
+        let meeting = try XCTUnwrap(
+            detector.checkOnce(),
+            "Cisco names the call assertion \"On a call\", not \"webex\"",
+        )
+        XCTAssertEqual(meeting.pattern.appName, "Webex")
+        // Auto-stop is the other half of the fix, and liveness takes a
+        // different route: `isMeetingActive` filters by `identifies` before it
+        // re-applies `matches`. Correct today because Webex is `.shared`, so
+        // both paths end in the same static `matches` — pinned here so a later
+        // split of the two routes has to fail out loud.
+        XCTAssertTrue(
+            detector.isMeetingActive(meeting),
+            "the measured assertion must also keep the meeting alive, or auto-stop fires at the end grace",
+        )
+    }
+
+    /// The measured pair, verbatim: one pid held both assertions at the same
+    /// time, and only the second one matches. The unmatched sibling must not
+    /// hide it — and the fixture has to be able to express the shape at all,
+    /// which is what the count pins.
+    func testWebexCallIsDetectedBesideTheUnmatchedSecondAssertion() {
+        let measured = PowerAssertionFixture.typedAssertions([
+            (pid: 870, processName: "Webex", assertName: "io avilable", assertType: PowerAssertionFixture.systemSleep),
+            (pid: 870, processName: "Webex", assertName: "On a call", assertType: PowerAssertionFixture.displaySleep),
+        ])
+        XCTAssertEqual(measured[870]?.count, 2, "one process holds several assertions at once; the fixture must keep both")
+
+        let detector = PowerAssertionDetector(
+            patterns: PowerAssertionDetector.patterns(watching: ["Webex"]),
+            confirmationCount: 1,
+        )
+        detector.windowListProvider = { [] }
+        detector.assertionProvider = { measured }
         XCTAssertEqual(
             detector.checkOnce()?.pattern.appName, "Webex",
-            "Cisco names the call assertion \"On a call\", not \"webex\"",
+            "\"io avilable\" comes first and matches nothing; \"On a call\" must still be found",
         )
     }
 
--- a/app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
+++ b/app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
@@ -282,4 +282,92 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
         _ = off.checkOnce()
         XCTAssertNil(off.checkOnce(), "browser toggle off → a Chrome WebRTC call must be ignored")
     }
+
+    // MARK: - Webex desktop client — Cisco's assertion name is not "webex"
+
+    /// Measured on macOS 26.5.2 with Webex 46.7.0.35472 during an instant
+    /// meeting: the client holds two assertions, and neither name contains
+    /// "webex".
+    ///
+    ///     pid 870(Webex): PreventUserIdleSystemSleep  named: "io avilable"
+    ///     pid 870(Webex): PreventUserIdleDisplaySleep named: "On a call"
+    ///
+    /// Same failure mode as Zoom in #562 (Apple's "Describe Activity Type"
+    /// placeholder): `processNames` passes the bound gate, then both remaining
+    /// tests fail — no "webex" in the name, and `assertionTypes` is empty. So
+    /// the desktop client can never be auto-detected, which in turn means no
+    /// auto-stop, because a hand-started recording only ends on
+    /// `.stopPidExited` / `.stopMaxDurationExceeded`.
+    func testWebexCallWithCiscoAssertionNameIsDetected() {
+        let detector = PowerAssertionDetector(
+            patterns: PowerAssertionDetector.patterns(watching: ["Webex"]),
+            confirmationCount: 1,
+        )
+        detector.windowListProvider = { [] }
+        detector.assertionProvider = {
+            PowerAssertionFixture.assertionDict(processName: "Webex", assertName: "On a call")
+        }
+        XCTAssertEqual(
+            detector.checkOnce()?.pattern.appName, "Webex",
+            "Cisco names the call assertion \"On a call\", not \"webex\"",
+        )
+    }
+
+    /// The "on a call" keyword must stay bound to the Webex client: `matches()`
+    /// checks `processNames.contains` before testing the name, so a foreign
+    /// process holding an identically named assertion cannot fire. Amphetamine
+    /// holds both display-sleep assertion types permanently on the measured
+    /// machine.
+    ///
+    /// That gate is NOT what makes a keyword preferable to the `assertionTypes`
+    /// fallback #475 gave Zoom: it runs ahead of both arms, so it excludes
+    /// foreign processes either way. The keyword earns its keep against
+    /// assertions Webex ITSELF holds outside a call, which is the #475 concern
+    /// and what `testWebexWithUnrelatedAssertionNameIsNotDetected` pins.
+    func testWebexKeywordIsBoundToTheWebexProcess() {
+        for process in ["Amphetamine", "Google Chrome", PowerAssertionFixture.unknownBrowser] {
+            let detector = PowerAssertionDetector(
+                patterns: PowerAssertionDetector.patterns(watching: ["Webex"]),
+                confirmationCount: 1,
+            )
+            detector.windowListProvider = { [] }
+            detector.assertionProvider = {
+                PowerAssertionFixture.assertionDict(processName: process, assertName: "On a call")
+            }
+            XCTAssertNil(
+                detector.checkOnce(),
+                "\(process): the Webex keyword must not fire for a foreign process",
+            )
+        }
+    }
+
+    /// A Webex client holding an unrelated assertion must still not count as a
+    /// meeting — the keyword gate, not the process name, decides. Measured: an
+    /// idle Webex client holds no assertion at all, so this is the guard against
+    /// widening the pattern to plain display-sleep later.
+    func testWebexWithUnrelatedAssertionNameIsNotDetected() {
+        let detector = PowerAssertionDetector(
+            patterns: PowerAssertionDetector.patterns(watching: ["Webex"]),
+            confirmationCount: 1,
+        )
+        detector.windowListProvider = { [] }
+        detector.assertionProvider = {
+            PowerAssertionFixture.assertionDict(processName: "Webex", assertName: "Playing audio")
+        }
+        XCTAssertNil(detector.checkOnce(), "an unrelated Webex assertion must not be a meeting")
+    }
+
+    /// The Webex toggle still gates the new keyword: with Webex unchecked, the
+    /// measured assertion must be ignored.
+    func testWebexCallIsIgnoredWhenWebexIsNotWatched() {
+        let detector = PowerAssertionDetector(
+            patterns: PowerAssertionDetector.patterns(watching: ["Zoom"]),
+            confirmationCount: 1,
+        )
+        detector.windowListProvider = { [] }
+        detector.assertionProvider = {
+            PowerAssertionFixture.assertionDict(processName: "Webex", assertName: "On a call")
+        }
+        XCTAssertNil(detector.checkOnce(), "Webex off → a Webex call must not be detected")
+    }
 }
--- a/app/MeetingTranscriber/Sources/PowerAssertionDetector.swift
+++ b/app/MeetingTranscriber/Sources/PowerAssertionDetector.swift
@@ -131,7 +131,10 @@ class PowerAssertionDetector: MeetingDetecting {
         AssertionPattern(
             appName: "Webex",
             processNames: ["Webex", "Cisco Webex Meetings", "Meeting Center"],
-            keywords: ["webex"],
+            // Cisco names the call assertion "On a call", not "webex" (measured:
+            // Webex 46.7.0.35472 on macOS 26.5.2). Keyword-only like Teams, so
+            // the #475 concern does not arise; bound to processNames above.
+            keywords: ["webex", "on a call"],
         ),
         AssertionPattern(
             appName: AppMeetingPattern.simulator.appName,
--- a/app/MeetingTranscriber/Tests/PowerAssertionFixture.swift
+++ b/app/MeetingTranscriber/Tests/PowerAssertionFixture.swift
@@ -13,19 +13,36 @@ enum PowerAssertionFixture {
     /// implementation can pass a test that uses it.
     static let unknownBrowser = "Fjordfox"
 
+    /// The two assertion types a meeting client is observed to hold. Only
+    /// patterns that set `assertionTypes` read this field, so `assertions`
+    /// pins the display-sleep one and a test spells both out when it
+    /// reproduces a measured mixture.
+    static let displaySleep = "PreventUserIdleDisplaySleep"
+    static let systemSleep = "PreventUserIdleSystemSleep"
+
     /// Assertions from one or more processes in a single poll. Variadic so a
     /// test can pin which browser a shared-keying implementation would have
     /// picked arbitrarily.
     static func assertions(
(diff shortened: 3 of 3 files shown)


app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
```diff
@@ -282,4 +282,87 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
...
+            "Cisco names the call assertion \"On a call\", not \"webex\"",
+        )
+    }
+
+    /// The "on a call" keyword must stay bound to the Webex client. This is the
+    /// counter-test to the one above and the reason a keyword is preferable to
+    /// the `assertionTypes` fallback #475 gave Zoom: `matches()` checks
+    /// `processNames.contains` first, so a foreign process holding an
+    /// identically named assertion cannot fire. Amphetamine holds both
+    /// display-sleep assertion types permanently on the measured machine.
```
pasrom: This rationale does not hold, and it is the one thing I want fixed before merge.

`matches()` applies the `processNames.contains` gate ahead of *both* arms, the keyword test and the `assertionTypes` test alike. Amphetamine is therefore excluded either way, so the gate is not what distinguishes a keyword from the `assertionTypes` fallback. The test is still worth keeping, it just pins something other than what the comment claims.

This matters more than a comment normally would, because the wrong reason points at the unsafe change: a reader who believes the process gate is what makes the keyword safe will conclude `assertionTypes` is equally safe, and that is exactly the mutation your next test catches. I checked, and it does catch it: adding `assertionTypes: ["PreventUserIdleDisplaySleep"]` here turns `testWebexWithUnrelatedAssertionNameIsNotDetected` red, and only that one.

The argument that does hold is the #475 one you already quote in the PR description: an `assertionTypes` fallback fires on any display-sleep assertion the app holds, including outside a call, and no process gate helps there because the process is Webex either way.

Note also that the comment on `testWebexWithUnrelatedAssertionNameIsNotDetected` below says "the keyword gate, not the process name, decides", which is the accurate version of the same thought. As it stands the two comments contradict each other.

The suggestion below passes both `swiftformat --lint` and `swiftlint --strict` unchanged.

```suggestion
    /// The "on a call" keyword must stay bound to the Webex client: `matches()`
    /// checks `processNames.contains` before testing the name, so a foreign
    /// process holding an identically named assertion cannot fire. Amphetamine
    /// holds both display-sleep assertion types permanently on the measured
    /// machine.
    ///
    /// That gate is NOT what makes a keyword preferable to the `assertionTypes`
    /// fallback #475 gave Zoom: it runs ahead of both arms, so it excludes
    /// foreign processes either way. The keyword earns its keep against
    /// assertions Webex ITSELF holds outside a call, which is the #475 concern
    /// and what `testWebexWithUnrelatedAssertionNameIsNotDetected` pins.
```

app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
```diff
@@ -282,4 +282,87 @@ final class PowerAssertionDetectorPatternsTests: XCTestCase {
...
+    ///
+    ///     pid 870(Webex): PreventUserIdleSystemSleep  named: "io avilable"
+    ///     pid 870(Webex): PreventUserIdleDisplaySleep named: "On a call"
+    ///
+    /// Same failure mode as Zoom in #562 (Apple's "Describe Activity Type"
+    /// placeholder): `processNames` passes the bound gate, then both remaining
+    /// tests fail — no "webex" in the name, and `assertionTypes` is empty. So
+    /// the desktop client can never be auto-detected, which in turn means no
+    /// auto-stop, because a hand-started recording only ends on
+    /// `.stopPidExited` / `.stopMaxDurationExceeded`.
```
pasrom, line 300: Optional, not blocking. This is where the auto-stop half of the fix is claimed, and all four new tests exercise `checkOnce()` only. Nothing asserts that `isMeetingActive()` returns true for the measured `"On a call"` shape, and liveness runs an extra `identifies` filter before it re-applies `matches`.

It is correct today: Webex is a `.shared` pattern, so `identifies` compares the app name and the same static `matches` decides both paths, which is also why a keyword regression cannot make the two disagree. So this is one cheap assertion for the sake of the claim, not a gap I would hold the PR for.

pasrom (changes requested):
Thanks for this, and in particular for the issue behind it. The measurement, the idle-versus-in-call control, and the explicit "here is why the current pattern cannot fire" trace made this reviewable in minutes rather than hours.

The production change is right, and I verified your numbers rather than taking them on trust:

- with the patch, `swift test --filter "PowerAssertionDetector"` gives 78 tests, 0 failures
- red-first reproduces exactly as you describe: reverting the one-liner gives 18 tests with 1 failure, and the failure is `testWebexCallWithCiscoAssertionNameIsDetected`
- wider run, `--filter "Detector|Assertion|Pattern"`: 172 tests, 0 failures
- one mutation you did not claim, worth knowing: adding `assertionTypes: ["PreventUserIdleDisplaySleep"]` to the Webex pattern turns `testWebexWithUnrelatedAssertionNameIsNotDetected` red, and only that test. Your guard is load-bearing against precisely the widening a later reader would be tempted by.

I am requesting changes on one thing only, and it is a comment. It is inline below, with a suggestion you can take as-is: the stated reason why a keyword beats the `assertionTypes` fallback is not the reason that actually holds, and the wrong version points a future reader at the unsafe change.

### Optional, and I am equally happy to take these as follow-ups

1. `PowerAssertionFixture.assertions` does `out[entry.pid] = [[...]]`, so two entries for the same PID overwrite rather than accumulate. The consequence is that the shape you actually measured, PID 870 holding `"io avilable"` and `"On a call"` at once, is the one shape the fixture cannot express, and your own doc comment documents it. Making the helper append and adding a single test with the verbatim measured pair would close that.

2. Nothing pins `isMeetingActive()`, noted inline. It is correct today, so this is one cheap assertion rather than a hole.

### A note on rollout rather than on your patch

`watchWebex` has defaulted to true since the toggle existed, and Webex does not require a recording consent prompt, so for every install with the default settings this flips a population from "never records" to "records automatically, without a prompt", roughly six seconds in. That is exactly what the toggle has always promised, and Teams and Zoom already behave this way, so it is not an argument against the change. It does mean this wants a line in the release notes, and it makes your open question about localization worth more than it looks: if `"On a call"` is localized, those users stay silently undetected, which is the pre-patch status quo and no worse, just invisible again.

I agree with your reasoning against the `assertionTypes` route, and I would not take it even with the idle measurement in hand. One build's idle behaviour is a thin base for widening a match to a whole assertion type.

### Your two follow-up offers

Yes to both, separately from this PR.

The `coreaudiod` and `kIOPMAssertionOnBehalfOfPID` detector is the interesting one: vendor-string-independent, no new entitlement, and the `audio-in` versus `audio-out` split is a genuinely better meeting signal than "something keeps the display awake". If you want to prepare it, please open it as its own PR against a fresh issue so it can be judged on its own evidence.

Raising `logUnmatchedWatchedAssertions` to `.default` is a good catch and small enough to stand alone. It is the log line that exists to make this class of bug self-diagnosing, and at `info` it does not survive to be read.

app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
DevUrandomFoo: Applied verbatim in 5e6d153 — thank you, and you are right that this was worth blocking on. The wrong reason pointed at the unsafe change, which makes it a correctness problem in the comment, not a wording preference. Your mutation check is the part I had not done: I claimed the keyword was safer without pinning which test would notice if someone disagreed.

The contradiction with `testWebexWithUnrelatedAssertionNameIsNotDetected` is resolved in the same commit — the two comments now name that test explicitly as the thing that pins the #475 concern, so the argument lives in one place instead of two half-versions.

Comment only, no behaviour change.

app/MeetingTranscriber/Tests/PowerAssertionDetectorPatternsTests.swift
DevUrandomFoo: Done in 7fc8407, together with the fixture point from your review body.

- `testWebexCallWithCiscoAssertionNameIsDetected` now unwraps `checkOnce()` and asserts `isMeetingActive(meeting)`, with the `identifies`-before-`matches` detour spelled out as the reason the assertion exists — so a later split of the two routes fails out loud instead of silently dropping auto-stop.
- `PowerAssertionFixture` accumulates per pid instead of overwriting, and `typedAssertions` takes the assertion type per entry. That makes the measured shape expressible at all: one pid holding `"io avilable"` (system sleep) and `"On a call"` (display sleep) at the same time. The variadic helper keeps its signature, so no existing call site moved.
- `testWebexCallIsDetectedBesideTheUnmatchedSecondAssertion` reproduces that pair with the unmatched sibling first, so the match cannot be found by luck, and pins the fixture's own shape by count — under the old helper that count is 1 rather than 2, which is what makes the fixture change testable rather than assumed.

One thing on your side: the workflow runs for 7fc8407 are sitting at `action_required`, so CI has not started on the new commits yet.

pasrom (approved):
All three points addressed, and the two I had only offered as follow-ups came back done as well. Thanks.

I mutation-tested the new half rather than reading it, because tests written in response to a review are exactly where nobody looks twice:

- reverting `typedAssertions` to overwrite per pid turns `testWebexCallIsDetectedBesideTheUnmatchedSecondAssertion` red at the count assertion
- taking the keyword match out of `isMeetingActive`, which is the split of the two routes your comment names, turns the new `XCTAssertTrue` red and nothing else: 19 tests, 1 failure

So both additions are load-bearing rather than decorative. The extra test for the unmatched sibling assertion was a good call and not something I had asked for.

One thing I checked before trusting the fixture change: no existing caller of `assertions()` passes two entries under the same pid, so switching from overwrite to append leaves every existing suite behaviourally unchanged.

Locally: SwiftFormat clean, SwiftLint `--strict` clean, 173 tests green across `Detector|Assertion|Pattern`.

On your two follow-up offers, both still welcome as separate PRs. The `coreaudiod` and `kIOPMAssertionOnBehalfOfPID` detector is the one I am most interested in.

pasrom:
Heads up: I rewrote the branch before merging, so your local copy is now behind a force-push. Sorry for doing that without asking first, and here is exactly what changed and why.

Two things happened:

1. `5e6d153` is folded into the first commit. It corrected a doc comment that the first commit had introduced in this same PR, and this repo merges by rebase, so both commits land on `main` verbatim. Without the fold, `main` would permanently carry "here is the rationale" followed by "that rationale was wrong". After the fold the comment is simply right from the start.
2. The branch is rebased onto current `main`, which was seven commits ahead. Branch protection requires up to date, so this had to happen anyway, and doing it in the same operation costs one CI run instead of two.

What I did not touch: `7fc8407` is unchanged as its own commit. It arguably does two things, the fixture change and the `isMeetingActive` pin, and by this repo's atomic-commit convention that is a candidate for a split. That is a judgement about your division of the work, not a defect, so I left it to you. If you would rather split it, say so and I will hold the merge.

Your authorship is intact on both commits, and I did not edit your commit messages. The body of the first one describes the process gate as an additional property rather than as the reason a keyword beats `assertionTypes`, so it did not carry the error the test comment did.

Proof that the fold changed no content: `git diff` from the old merge base to the old tip and `git diff` from current `main` to the new tip are byte identical, 190 lines both. I also ran the first commit on its own, 172 tests green, so the history stays bisectable, and the new tip at 173 tests green with SwiftFormat and SwiftLint clean.

The new tip is `422022b`. To realign your local branch:

```
git fetch origin && git reset --hard origin/fix/webex-on-a-call-assertion
```

pasrom:
Merged, thank you. Both commits are on `main` with your authorship intact.

To close the loop on what I wrote above: I said I would hold the merge if you wanted to split `7fc8407`, and then merged without leaving you a real window to answer. That was inconsistent of me, so to be explicit rather than let it stand: the split is still welcome as a follow-up PR if you think the fixture change and the auto-stop pin belong apart. Nothing about this merge forecloses that.

The issue behind this was the most useful bug report this project has had in a while. The idle-versus-in-call control is what made the fix reviewable without me having to reproduce anything, and pre-empting the "why not assertionTypes" question saved an entire round trip.

Both follow-ups you offered are still open. The `coreaudiod` and `kIOPMAssertionOnBehalfOfPID` detector in particular: if that works, it retires a whole class of vendor-string bugs rather than the one we just fixed, and it would cover the Slack Huddles gap too. Open it against its own issue whenever suits you.

PR #661, opened 2026-08-27
feat(app): hide live caption overlay without stopping transcription

marcus:
## Why

Live transcription today is one switch that both starts the streaming ASR pipeline and shows the click-through caption bar. The bar is useful for a lot of people, but it is large, sits on every Space, and is awkward during a screen share. Other people consume captions through the Local Automation API (`GET /state` → `liveCaptions`) and just want the stream, not the on-screen overlay.

## What

This adds Settings → Transcribe → "Show caption overlay" (default on), nested under "Show partial transcripts during recording". Turning the overlay off hides `LiveCaptionsOverlay` only. The existing master toggle still starts and stops the pipeline. `LiveCaptionsState` and `/state.liveCaptions` keep updating. Overlay visibility can change mid-recording (UI only, no controller rebuild).

## Compatibility

The overlay stays on by default, so current users keep the bar. The master toggle still owns the pipeline. `LiveCaptionsGate.strategy` and the coordinator keep the same inputs.

## Test plan

- [x] `AppSettingsTests.testDefaultValues`: overlay defaults on (missing UserDefaults key → true)
- [x] `LiveCaptionsGateTests`: `overlayVisible` is false when `overlayEnabled` is false even if captions are available and recording; true when all three are true; false when not recording
- [x] `SettingsInteractionTests`: tap "Show caption overlay" writes back to `AppSettings`; toggle is disabled when live transcription is off
- [x] `RPCSettingsStateTests`: `recording.liveCaptionsOverlayEnabled` on the settings allowlist, default snapshot, and flipped-value snapshot
- [ ] Manual: enable live transcription, start a recording, turn the overlay off. Confirm the bar hides and `/state.liveCaptions` still moves if the Local Automation API is on.


app/MeetingTranscriber/Sources/AppState.swift
```diff
@@ -450,14 +450,17 @@ final class AppState {
...
+    /// toggle is on. Overlay visibility is UI-only; the pipeline still runs.
     var shouldShowLiveCaptions: Bool {
-        LiveCaptionsGate.captionsAvailable(
+        LiveCaptionsGate.overlayVisible(
             liveEnabled: settings.liveTranscriptionEnabled,
             engineLanguage: settings.activeEngineLanguageOrNil,
             engineSupportsLive: settings.transcriptionEngine.supportsLiveTranscription,
-        ) && watching.isRecording
+            isRecording: watching.isRecording,
+            overlayEnabled: settings.liveCaptionsOverlayEnabled,
```
pasrom: This is the only line that carries the new setting into the window, and nothing exercises it at any layer.

Reproduction: change it to `overlayEnabled: true` and run `swift test --parallel`. 3045 tests, exit 0, nothing red. What that lets through is silent: the toggle flips, `/state` reports the new value, and the bar stays on screen.

The pin for this already exists in the repo. `/state.windows` carries every window that has a scene identifier, and the caption panel sets one (`LiveCaptionsWindowController`: `panel.identifier = NSUserInterfaceItemIdentifier("live-captions")`). `scripts/e2e-live-captions.sh` already starts a real recording and polls `/state` until `liveCaptions.recentFinals` grows, and it already seeds settings through `write_dev_default` with a `snapshot_default` / `restore_bool_default` pair. So this is four lines in that script:

```bash
# next to the existing SAVED_LIVE_TRANS snapshot
SAVED_OVERLAY=$(snapshot_default "$_STANDARD_PLIST" liveCaptionsOverlayEnabled)

# next to the other write_dev_default calls
write_dev_default "$BUNDLE_ID" liveCaptionsOverlayEnabled false bool

# once recentFinals has grown, i.e. captions are demonstrably flowing
rpc /state | jq -e '[.windows[] | select(.id == "live-captions") | .isVisible] | any | not' >/dev/null \
    || fail "caption overlay is on screen although liveCaptionsOverlayEnabled is false"

# in cleanup
restore_bool_default "$_STANDARD_PLIST" liveCaptionsOverlayEnabled "$SAVED_OVERLAY"
```

That pins this line (nail it to `true` and the panel gets created and shown, so the assertion fires) and, in the same run, the "captions keep flowing while the overlay is hidden" claim from the description, which nothing checks today either.

If a live lane is more than you want to take on in this PR, a cheaper partial is a unit test on `AppState.shouldShowLiveCaptions` with a recording watch loop, asserting it follows `settings.liveCaptionsOverlayEnabled`. That leaves `.onChange` to `orderOut` unproven, but it does catch this line.

app/MeetingTranscriber/Sources/AppSettings.swift
```diff
@@ -216,6 +216,14 @@ final class AppSettings {
         didSet { defaults.set(liveTranscriptionEnabled, forKey: "liveTranscriptionEnabled") }
     }
 
+    /// Overlay-window visibility only. Live transcription still runs when the
+    /// master toggle is on; this setting only hides `LiveCaptionsOverlay`.
+    /// Default on, so existing users keep seeing the bar. Missing UserDefaults
+    /// key reads as true (`defaults.object(forKey:) as? Bool ?? true`).
+    var liveCaptionsOverlayEnabled: Bool {
+        didSet { defaults.set(liveCaptionsOverlayEnabled, forKey: "liveCaptionsOverlayEnabled") }
```
pasrom: Untested in the same way: change the key here to anything else and the suite stays green. I ran that mutation against `AppSettingsTests`, `RPCSettingsStateTests`, `SettingsInteractionTests` and `LiveCaptionsGateTests`, 91 tests, all still passing.

The key is correct as written, and `AppSettingsTests.testDefaultValues` pins the missing-key default. What is missing is the round trip, which this file already does for comparable preferences (`testWatchBrowserMeetingsPersists`, `testDebugRPCEnabledPersistence`, `testCheckForUpdatesPersistence`): write the value, build a fresh `AppSettings` over the same defaults, read it back. Without it, a wrong key means the preference silently resets to on at every launch and nothing notices.

app/MeetingTranscriber/Tests/SettingsInteractionTests.swift
```diff
@@ -127,6 +127,41 @@ final class SettingsInteractionTests: XCTestCase {
...
+        settings.liveTranscriptionEnabled = true
+        XCTAssertTrue(settings.liveCaptionsOverlayEnabled, "precondition: overlay defaults on")
+        let view = TranscriptionSettingsView(
+            settings: settings,
+            whisperKitEngine: WhisperKitEngine(),
+            parakeetEngine: ParakeetEngine(),
+        )
+        let toggle = try view.inspect().find(ViewType.Toggle.self) { toggle in
+            try toggle.labelView().text().string() == "Show caption overlay"
+        }
```
pasrom: `A11yID.liveCaptionsOverlayToggle` is added in this PR and currently has no consumer at all: not here, and not in the `/ui/press` allowlist. Locating by the label string instead means a reworded label fails this test for the wrong reason, while a renamed or deleted identifier fails nothing.

Every other Toggle test in this file (`testIncludeFullTranscriptToggleWritesBackToSettings`, `testSaveRawTranscriptToggleWritesBackToSettings`) and in `EchoCancellationSettingTests` goes through the identifier, which is also what the project's GUI testing rules ask for.

I ran the replacement below: green as written, and red with `Search did not find a match` once `.accessibilityIdentifier(A11yID.liveCaptionsOverlayToggle)` is removed from the view. `./scripts/lint.sh` is clean on it.

```suggestion
        let toggle = try view.inspect()
            .find(viewWithAccessibilityIdentifier: A11yID.liveCaptionsOverlayToggle)
            .find(ViewType.Toggle.self)
```

app/MeetingTranscriber/Tests/SettingsInteractionTests.swift
```diff
@@ -127,6 +127,41 @@ final class SettingsInteractionTests: XCTestCase {
...
+        let settings = makeSettings()
+        settings.liveTranscriptionEnabled = false
+        let view = TranscriptionSettingsView(
+            settings: settings,
+            whisperKitEngine: WhisperKitEngine(),
+            parakeetEngine: ParakeetEngine(),
+        )
+        let toggle = try view.inspect().find(ViewType.Toggle.self) { toggle in
+            try toggle.labelView().text().string() == "Show caption overlay"
+        }
```
pasrom: Same here. Note that `EchoCancellationSettingTests.testTheDedupRowIsDisabledWhileCancellationIsOn` asserts `isDisabled()` on the identifier-bearing view directly rather than on the `Toggle` inside it, which is what the suggestion does. Verified green, and red once the identifier is removed.

```suggestion
        let toggle = try view.inspect()
            .find(viewWithAccessibilityIdentifier: A11yID.liveCaptionsOverlayToggle)
```

app/MeetingTranscriber/Sources/Settings/TranscriptionSettingsView.swift
```diff
@@ -161,6 +163,17 @@ struct TranscriptionSettingsView: View {
...
     }
 
+    /// Nested under the master live-transcription toggle. Disabled when the
+    /// parent is off so the overlay cannot be flipped independently of the
+    /// pipeline. Visibility only: the coordinator still arms when the master
+    /// toggle is on.
+    private var captionOverlayToggle: some View {
+        Toggle("Show caption overlay", isOn: $settings.liveCaptionsOverlayEnabled)
+            .disabled(!settings.liveTranscriptionEnabled)
+            .padding(.leading)
```
pasrom: Minor, and a genuine question rather than a complaint: this is the only `.padding(.leading)` in the Settings views. The other controls that depend on a sibling setting express that with `.disabled` alone and no indent (`AudioSettingsView` line 96, `OutputSettingsView` line 73, `SpeakersSettingsView` line 196).

Indenting a dependent row is arguably nicer than the status quo. It is just worth being a deliberate choice, since it starts a convention the other four places would eventually want to follow.

app/MeetingTranscriber/Sources/Settings/TranscriptionSettingsView.swift
```diff
@@ -227,8 +240,9 @@ struct TranscriptionSettingsView: View {
         // future engine returns `supportsLiveTranscription == false`, reintroduce
         // a conditional "unsupported" message gated on that + `englishStreaming`.
         "Captions appear in a click-through overlay at the bottom of "
-            + "the screen during recording. Hold ⌥ (Option) to drag "
-            + "it; the position is remembered across sessions. "
+            + "the screen during recording. Turn off \"Show caption overlay\" "
```
pasrom: Product question rather than a defect, and the only thing here I would want your opinion on.

Consider: overlay off, then master off, then master on. The nested toggle comes back enabled but still off, because the preference is remembered. So the user has just switched on something labelled "Show partial transcripts during recording", possibly confirming the 0.6 GB download on the way, and sees no transcripts at all. This footnote explains it, but a footnote is doing a lot of work there.

Remembering the sub-preference is defensible. The alternatives are to reset the overlay to on whenever the master is switched on, or to reword the master so it stops promising something visible. I lean towards rewording, but it is your call.

app/MeetingTranscriber/Sources/AppState.swift
pasrom, line 462: This is the only line that carries the new setting into the window, and nothing exercises it at any layer.

Reproduction: change it to `overlayEnabled: true` and run `swift test --parallel`. 3045 tests, exit 0, nothing red. What that lets through is silent: the toggle flips, `/state` reports the new value, and the bar stays on screen.

**The cheap fix, and the one I am actually asking for:** a unit test on `AppState.shouldShowLiveCaptions` with a driven watch loop. The header comment in `LiveTranscriptionGatingTests` says the recording cases "would need a real driven WatchLoop and are deferred to live-E2E", which is out of date. It is roughly 25 lines: `ManualRecordingTests.makeLoop` has the recipe (a `WatchLoop` built with a `MockRecorder` factory, `startManualRecording`, wait for `.recording`), then assign `state.watching.watchLoop = loop` the way `AppStateTests` does and assert that `shouldShowLiveCaptions` follows `settings.liveCaptionsOverlayEnabled`. That goes red under the mutation above, and leaves only `.onChange` to `orderOut` unproven.

**The stronger pin exists too, but neither you nor CI can run it on this PR.** `/state.windows` carries every window that has an identifier, and the caption panel sets one (`LiveCaptionsWindowController`: `panel.identifier = NSUserInterfaceItemIdentifier("live-captions")`), so `scripts/e2e-live-captions.sh` could assert on it directly. But `e2e-app.yml` only runs for same-repo branches, so a fork PR never triggers that lane. I would rather add this myself on a same-repo branch after your change lands than ask you for something you cannot verify. Recorded here so it does not get lost:

```bash
# next to the existing SAVED_LIVE_TRANS snapshot
SAVED_OVERLAY=$(snapshot_default "$_STANDARD_PLIST" liveCaptionsOverlayEnabled)

# next to the other write_dev_default calls
write_dev_default "$BUNDLE_ID" liveCaptionsOverlayEnabled false bool

# once recentFinals has grown, i.e. captions are demonstrably flowing.
# Capture first: rpc swallows curl failures into empty output, and jq -e
# on empty input exits 0, so piping straight in could pass vacuously.
state=$(rpc /state)
[ -n "$state" ] || fail "no /state response"
jq -e '[.windows[] | select(.id == "live-captions") | .isVisible] | any | not' >/dev/null <<<"$state" \
    || fail "caption overlay is on screen although liveCaptionsOverlayEnabled is false"

# in cleanup
restore_bool_default "$_STANDARD_PLIST" liveCaptionsOverlayEnabled "$SAVED_OVERLAY"
```

Worth being explicit about its limit: with the setting off from launch the panel is never created, so that assertion passes by absence. It pins this mutation, not "the overlay works in general". The complete version pairs it with a run that asserts `isVisible == true` with the setting on.

app/MeetingTranscriber/Sources/AppSettings.swift
pasrom, line 224: Untested in the same way: change the key here to anything else and the suite stays green. I ran that mutation against `AppSettingsTests`, `RPCSettingsStateTests`, `SettingsInteractionTests` and `LiveCaptionsGateTests`, 91 tests, all still passing, and then against the full suite with the same result.

The key is correct as written, and `AppSettingsTests.testDefaultValues` pins the missing-key default. What is missing is the round trip, which this file already does for comparable preferences: write the value, build a fresh `AppSettings(defaults:)`, read it back. `testProtocolProviderPersistence` and `testWatchBrowserMeetingsPersists` are the pattern. Without it, a wrong key means the preference silently resets to on at every launch and nothing notices.

app/MeetingTranscriber/Tests/SettingsInteractionTests.swift
pasrom: `A11yID.liveCaptionsOverlayToggle` is added in this PR and currently has no consumer at all: not here, and not in the `/ui/press` allowlist. Locating by the label string instead means a reworded label fails this test for the wrong reason, while a renamed or deleted identifier fails nothing.

The other two Toggle tests in this file (`testIncludeFullTranscriptToggleWritesBackToSettings`, `testSaveRawTranscriptToggleWritesBackToSettings`) and the ones in `EchoCancellationSettingTests` all go through the identifier, which is also what the project's GUI testing rules ask for.

I ran the replacement below: green as written, and red with `Search did not find a match` once `.accessibilityIdentifier(A11yID.liveCaptionsOverlayToggle)` is removed from the view. `./scripts/lint.sh` is clean on it.

```suggestion
        let toggle = try view.inspect()
            .find(viewWithAccessibilityIdentifier: A11yID.liveCaptionsOverlayToggle)
            .find(ViewType.Toggle.self)
```

app/MeetingTranscriber/Tests/SettingsInteractionTests.swift
pasrom: Same here. Note that `EchoCancellationSettingTests.testTheDedupRowIsDisabledWhileCancellationIsOn` asserts `isDisabled()` on the identifier-bearing view rather than on the `Toggle` inside it, which is what the suggestion does. Verified green, red once the identifier is removed, and still red if `.disabled(...)` is nailed to `false`, so it is not a vacuous assertion.

```suggestion
        let toggle = try view.inspect()
            .find(viewWithAccessibilityIdentifier: A11yID.liveCaptionsOverlayToggle)
```

app/MeetingTranscriber/Sources/Settings/TranscriptionSettingsView.swift
pasrom: Optional, and a genuine question rather than a complaint: this is the only `.padding(.leading)` in `Sources`. The closest analogue is `AudioSettingsView` line 96, a toggle disabled by its sibling toggle, and it carries no indent. `OutputSettingsView` line 73 and `SpeakersSettingsView` line 196 are Reset buttons disabled by state, so looser comparisons.

Indenting a dependent row is arguably nicer than the status quo. It is just worth being a deliberate choice, since it starts a convention those places would eventually want to follow.

app/MeetingTranscriber/Sources/Settings/TranscriptionSettingsView.swift
pasrom: Optional, and a product question rather than a defect.

Consider: overlay off, then master off, then master on. The nested toggle comes back enabled but still off, because the preference is remembered. So the user has just switched on something labelled "Show partial transcripts during recording", possibly confirming the 0.6 GB download on the way, and sees no transcripts at all. This footnote explains it, but a footnote is doing a lot of work there.

Remembering the sub-preference is defensible. The alternatives are to reset the overlay to on whenever the master is switched on, or to reword the master so it stops promising something visible. I lean towards rewording, but it is your call.

pasrom (commented):
Thanks for this. The split is the right one and the implementation is careful: outside the toggle's own binding, the new flag is read only in `AppState.shouldShowLiveCaptions` and in the RPC projection, so the pipeline gate really is untouched. All four claims in the description are true of the code as written.

One thing should change before merge. The rest is smaller, and I have marked which is which.

### Blocking: the one line that carries the setting into the window is uncovered

Replace `overlayEnabled: settings.liveCaptionsOverlayEnabled` in `AppState.swift` with `overlayEnabled: true` and run `swift test --parallel`: 3045 tests, exit 0, nothing red. The toggle still writes UserDefaults, `/state.settings.recording.liveCaptionsOverlayEnabled` still flips, and the caption bar stays on screen.

The new `LiveCaptionsGate` tests are good and they do earn their keep: drop `&& overlayEnabled` from `overlayVisible` and exactly one of them goes red. The gap is one level up. Nothing exercises `AppSettings` to `AppState` to `.onChange` to `orderOut`, and `LiveTranscriptionGatingTests` only covers the `watchLoop == nil` branch, where the answer is false whatever the overlay flag says.

There is a cheap way to close this that needs no live lane. Recipe in the line comment.

### Requested, not blocking: persistence has the same shape

Change the key string in the `didSet` to something wrong and the full suite stays green too.

### Optional

The indentation note and the product question in the line comments. Neither needs to happen for this to merge.

### What I checked, so you do not have to

- **Mid-recording propagation works.** `AppSettings` is `@Observable` and the new property is plain stored state; `AppState` is `@Observable` and reads it directly. The read is a call argument in `shouldShowLiveCaptions`, so it is evaluated, and therefore tracked, on every evaluation, with no short circuit ahead of it. `LiveTranscriptionCoordinatorTests.testSettingsChangeDuringRecordingDefersRebuildUntilIdle` shows the same `@Observable` change delivery for a mid-recording `AppSettings` change, though through the coordinator's own `withObservationTracking` rather than SwiftUI's `.onChange`.
- **Hide and re-show lose nothing.** `hide()` is `orderOut` with the panel and its hosting view retained, `show()` goes through `positionAtSavedOrDefault`, and click-through, window level and Spaces behaviour are set once in `ensurePanel` and never reset.
- **No rebuild, no model re-warm.** `LiveTranscriptionCoordinator` never reads the overlay flag, so flipping it triggers no settings-change handling.
- **The `false` for the new field in `RPCStateSnapshot.Recording.empty` is fine.** It follows the zero-value convention already used there (`perChannelIndicatorEnabled`, `micName` and `asymmetricSilenceWarningSeconds` are all zeroed despite non-zero defaults), and the sentinel is only reachable once `AppState` has been deallocated. Noting it only so nobody raises it again.

marcus:
Addressed the review on this branch (rebased onto current `main`):

**Blocking**
- Added `LiveTranscriptionGatingTests.testShouldShowFollowsOverlayToggleWhileRecording`: driven `WatchLoop` via `makeTestWatchLoop` + `startManualRecording`, asserts `AppState.shouldShowLiveCaptions` follows `settings.liveCaptionsOverlayEnabled` (red if `overlayEnabled: true` is hardcoded). Updated the outdated E2E-only header comment.
- Did **not** touch `e2e-live-captions.sh` (as noted — fork PR cannot run that lane).

**Requested**
- `AppSettingsTests.testLiveCaptionsOverlayEnabledPersists`: write → fresh `AppSettings(defaults:)` → read back (wrong key goes red).

**Line comments**
- `SettingsInteractionTests` finds the overlay toggle via `A11yID.liveCaptionsOverlayToggle` (disabled assert on the identifier-bearing view, EchoCancellationSettingTests pattern). Identifier remains on the Toggle.
- Removed `.padding(.leading)` from `captionOverlayToggle` to match AudioSettingsView.
- Reworded master toggle to “Enable live transcription during recording” and adjusted the footnote so it no longer promises on-screen captions when the overlay can stay off. Overlay preference is still remembered; master-on does not reset it.

Focused tests: AppSettings (persistence), LiveCaptionsGateTests, LiveTranscriptionGatingTests, SettingsInteraction overlay cases — 20/20 passed.

pasrom:
Verified on `741b2095`: all three mutations now go red. Hardcoding `overlayEnabled: true` fails `testShouldShowFollowsOverlayToggleWhileRecording`, a wrong `didSet` key fails `testLiveCaptionsOverlayEnabledPersists`, and removing the identifier fails both overlay tests with `Search did not find a match`. The gap is genuinely closed. Good call on the master-label reword too: it fixes the problem at the root, which makes remembering the sub-preference the right behaviour rather than a compromise.

CI has finished. `lint` is the only red check, everything else that ran is green. Two small things:

**1.** The lint failure is a file that was already full:

```
AppSettingsTests.swift:604:1: error: File Length Violation: File should contain 600 lines or less: currently contains 604
```

596 lines on `main`, so the eight added lines tipped it over. Moving `testLiveCaptionsOverlayEnabledPersists` out on its own gets it to 597; the default assertion can stay with the other defaults. `EchoDedupSettingTests` and `EchoCancellationSettingTests` are the existing one-file-per-setting shape.

**2.** `CLAUDE.md` line 136 and `docs/architecture-macos.md` line 498 still say "Show partial transcripts during recording", in paragraphs this PR already edits.

The `/state.windows` assertion in `scripts/e2e-live-captions.sh` stays with me as agreed; I will add it on a same-repo branch once this lands.

pasrom:
Pushed two commits to your branch so this can make the next release rather than wait on another round trip. Both are mechanical, and I did not touch anything else:

- `5cf2cc64` moves `testLiveCaptionsOverlayEnabledPersists` into its own `LiveCaptionsOverlaySettingTests.swift`, following the `EchoDedupSettingTests` shape. `AppSettingsTests.swift` is back to 597 lines and lint passes. The default assertion stayed in `testDefaultValues` with the other defaults. I re-ran the mutation after the move: a wrong `didSet` key still turns the test red on both assertions, so it kept its power.
- `5a38e1fa` updates `CLAUDE.md` and `docs/architecture-macos.md` to the new toggle label.

Say the word if you would rather have done it yourself and I will drop them.

pasrom (approved):

PR #668, opened 2026-09-02
feat(app): add About settings tab, relocate version info and Updates

iHateCode:
## What changed and why

Settings gets an About tab, last in the sidebar. It carries the version, bundle
identifier, build date and ffmpeg availability, plus a link to the project repo —
the first time that link exists anywhere in the app.

Those identity rows used to sit buried in Advanced, and the Updates section lived
in General, separated from the version display it relates to. Both now live in
About. Update checking itself is untouched; the section moved wholesale.

Two smaller things came out of building it:

The GitHub link needed the owner and repo name, which `GitHubReleaseProvider`
already held privately for update checks. Rather than write the strings a second
time, those are now internal statics with one source of truth — a repo move would
otherwise go stale in the link while the update checker kept working.

Driving the new tab through the local automation API turned up a real gap. A
Settings sidebar row is a `List(selection:)` row: it accepts a plain accessibility
press, reports success, and doesn't actually select itself. The server already
supported a real click, but `mt-cli ui-press` had no way to ask for one, so it
gained `--via ax|click`. Without that, `test_rpc.sh` would have been asserting
against a tab that never opened.

## Also in here

One unrelated commit adds a note to `CLAUDE.md` warning against shortening a test
run with a pipe. `swift test | tail -60` exits 0 even when the suite is red, because
the shell reports the last command in the pipeline — so a red run reads as a clean
pass. Bundled rather than split out because it is four lines of documentation.

## Checklist

- [x] Rebased on `main`, no merge commits
- [x] `./scripts/lint.sh` clean
- [x] `swift test --parallel` green


iHateCode:
**Re: Codecov coverage drop**

The uncovered lines are the Updates section's interactive states (button press, spinner, status labels, download button) plus one `switch` case in `SettingsView`. Neither was covered by view-level tests before this PR — this project tests that state logic in `UpdateChecker`'s own unit tests, not in the view, and none of the other Settings tabs have their `switch` case covered by a view test either (the live RPC script covers that instead). The About tab is just following the same pattern every other tab already has, not introducing a new gap.

This response was written with AI assistance, but I've read it fully and stand behind it. Happy to take pushback on any of this, and I'm glad to add whatever tests you'd want to see.

app/MeetingTranscriber/Tests/DebugRPCServerUIPressTests.swift
```diff
@@ -61,6 +61,21 @@
...
+        func testAboutTabRowCanBePressed() {
+            let target = FakePressTarget(identifier: A11yID.settingsTabAbout)
+            let root = FakePressTarget(identifier: "settings", children: [target])
+
+            let outcome = DebugRPCServer.performPress(
+                identifier: A11yID.settingsTabAbout, in: root, maxDepth: 10, via: .click,
+            )
+
+            XCTAssertEqual(outcome, .pressed(true))
+            XCTAssertEqual(target.clickCallCount, 1)
```
pasrom, line 76: This is the one I would most want fixed, because it looks like it guards the allowlist entry and does not.

`performPress` never consults `uiPressAllowedIdentifiers`, and the allowlist test further down asserts only `recordOnlyToggle`. So removing `A11yID.settingsTabAbout` from the production set leaves CI green while the real request returns 403.

One line closes it:

```swift
XCTAssertTrue(DebugRPCServer.isIdentifierAllowedForUIPress(A11yID.settingsTabAbout))
```

Speakers has the same unit-level gap, but the smoke lane exercises it live over curl. About has no such lane, so this entry is currently covered by nothing that runs in CI.

pasrom (commented):
Thanks for this, and sorry for the wait. Two independent reviews went over it and neither found a correctness or security problem. The About tab, the relocation of the identity rows and the Updates section, and the shared owner and repo statics all hold up.

Checked rather than assumed: the Updates section moved without behavioural change (same bindings, `appState.updateChecker` passed through, `startPeriodicChecks` still in the app `.task`, no `APPSTORE` guard on either side); the provider refactor is behaviour preserving, with `releases/latest` and `releases` intact; and on `/ui/press`, `route()` runs Origin, Host and bearer first while `uiPressResponse` checks both allowlists before `performPress(via:)`, the click uses the matched element's own frame midpoint rather than anything from the request, and `AboutSettingsView` declares no identifiers, so the GitHub link and the download button stay unreachable to a token holder. `via` already existed on the server, so the CLI flag adds no capability.

I also ran `scripts/test_rpc.sh` against your branch on real hardware, including your new About step. It passes: the click selects the tab and the tree shows the About pane. Then I ran it again with your branch rebased onto current `main`, and it still passes.

Four line comments below, plus one thing that has no line to hang on:

**`docs/architecture-macos.md`** still lists Updates under General and About under Advanced, in two tables, and has no `AboutSettingsView.swift` row. After this move, anyone using that table to find the code lands in the wrong file.

## Not your fault, listed so the diff does not look worse than it is

- Your branch is **19 commits behind `main`**, and almost all of that landed after you pushed. You were current at the time. `merge-tree` is clean, so the rebase is mechanical.
- `scripts/test_rpc.sh` could not run at all on a machine holding both an Apple Development and a Developer ID certificate: it picked the wrong one and died at `503: no window` before ever reaching your step. Fixed on `main` now (#694), so you do not need to work around it.
- Its "Cache invalidation roundtrip" section was failing on `main` too while that signing bug was present.

tools/mt-cli/Tests/UIPressTests.swift
```diff
@@ -0,0 +1,27 @@
...
+/// `--via` breaks the About tab lives in `scripts/test_rpc.sh`.
+final class UIPressTests: XCTestCase {
+    func testViaDefaultsToNilSoOlderCallsKeepUsingAXPress() throws {
+        let command = try UIPress.parse(["someIdentifier"])
+        XCTAssertNil(command.via)
+    }
+
+    func testViaClickParses() throws {
+        let command = try UIPress.parse(["someIdentifier", "--via", "click"])
+        XCTAssertEqual(command.via, .click)
```
pasrom, line 16: Same shape as the allowlist one: these pin that the flag **parses**, not that `run()` puts the `via` field into the request body.

That matters more than usual here, because no workflow runs `swift test` in `tools/mt-cli`, and the smoke lane drives the server with curl rather than the CLI. So deleting the line in `MTCLI.swift` that adds the field would leave every automated check green while About selection silently stops working, which is exactly the failure your `--via` flag exists to prevent.

CLAUDE.md
```diff
@@ -58,6 +58,10 @@ Single-source: Audio/Video → 16kHz mono (AVAudioFile → AVAsset → ffmpeg fa
 ./scripts/run_app.sh
 
 # Swift tests (parallel — ~1.4× faster than sequential)
+# Never pipe a test run into `tail`/`head`/`grep` to shorten it: the shell reports
+# the *last* command's status, so `swift test | tail -60` exits 0 on a red suite and
+# the truncation hides which test failed. Redirect to a file and read that instead
+# (`swift test --parallel > /tmp/run.log 2>&1`), or check ${PIPESTATUS[0]}.
```
pasrom: The warning is right and worth having. I fell for exactly this today: I wrote a status line reading `PIPESTATUS[0]`, got an empty value back, and read straight past it.

The remedy is the problem. `PIPESTATUS` is bash-only and expands to an empty string under zsh, which is the default macOS shell and the one an agent runs commands in. The zsh spelling is `pipestatus[1]`, which is equally non-portable the other way.

Measured in both shells:

```
false | tail        exit 0 in both
PIPESTATUS[0]       1 in bash, empty in zsh
pipestatus[1]       empty in bash, 1 in zsh
set -o pipefail     exit 1 in both
```

The redirect advice in the same note is portable and correct. Suggest replacing the `PIPESTATUS` clause with `set -o pipefail`.

scripts/test_rpc.sh
```diff
@@ -190,6 +190,19 @@ out=$("$MT_CLI_BIN" ui-press recordOnlyToggle --window speaker-naming 2>&1 || tr
...
+# Press the About tab and confirm its detail pane actually opened. "GitHub" is
+# the marker (not "About") because the sidebar row's own label is always in
+# the tree regardless of which pane is selected — "GitHub" only exists inside
+# the About detail content, so it only appears once that pane is showing.
+# --via click: a sidebar row is a List(selection:) row, which accepts a plain
+# ax press and reports it ran without actually selecting itself (see
+# DebugRPCServer+UIPress.swift's uiPressAllowedIdentifiers comment).
+"$MT_CLI_BIN" ui-press settings-tab-about --window settings --via click >/dev/null \
+    || fail "ui-press settings-tab-about failed"
+tree=$("$MT_CLI_BIN" ui-tree --window settings)
```
pasrom: Small race: the click posts mouse events and returns, and the tree is read on the very next line. On a loaded machine the queued click may not have been handled yet, so this can observe the General pane and fail spuriously.

`e2e-settings-smoke.sh` already has the shape for this, polling for its marker rather than reading once. Worth borrowing here.

For what it is worth, the step itself is correct: I ran this script against your branch and the About pane came up as asserted.

app/MeetingTranscriber/Tests/AboutSettingsViewTests.swift
```diff
@@ -0,0 +1,101 @@
...
+        let body = try makeAbout().inspect()
+        XCTAssertNoThrow(try body.find(text: "ffmpeg"))
+    }
+
+    func testAboutSectionShowsGitHubLink() throws {
+        let body = try makeAbout().inspect()
+        XCTAssertNoThrow(try body.find(text: "GitHub"))
+    }
+
+    func testUpdatesSectionShownWhenUpdateCheckerPresent() throws {
```
pasrom, line 75: Pre-existing and not introduced by you, but this is the natural place to close it: no test in this file drives a binding or an action, so deleting the manual-check button or wiring a toggle to the wrong setting would still pass. The house rule in `CLAUDE.md` is one wiring test per control.

The one thing this PR did change is that the retarget dropped the explicit `Check Now` assertion the old tests had, so that control now has no coverage at all.

On coverage generally: your reply on the Codecov comment is correct. Those Updates branches were not covered at view level before either, and the state logic is tested in `UpdateCheckerTests`.

pasrom (approved):
All five line comments are addressed, and the docs point too. Read in the code rather than taken from the commit messages:

- `isIdentifierAllowedForUIPress(A11yID.settingsTabAbout)` is now asserted, so removing the allowlist entry turns CI red instead of leaving it green while the real request 403s.
- `requestPayload` is extracted and `run()` posts exactly that value, so the new `--via` test pins the request body rather than a parallel copy of it. That is the one that could have gone wrong quietly, and it did not.
- The `PIPESTATUS` clause is gone in favour of the redirect plus `set -o pipefail`, which behaves the same in bash and zsh.
- `test_rpc.sh` polls for the About marker instead of reading the tree once.
- `testCheckNowButtonStartsAnUpdateCheck` drives the real button.
- `docs/architecture-macos.md` carries the `AboutSettingsView.swift` row and lists the About tab.

Two commits in the new push were written after my review, so I read those as well. Promoting `owner` and `repo` to statics leaves the URL identical, and the `force_unwrapping` suppression sits on a static built from those two constants, 130 characters against a warning threshold of 160. Both fine.

On the Codecov comment: your reading was right, and I checked it rather than taking it on trust. The 23 lines in `AboutSettingsView` are the Updates section's four render states plus the download button, moved verbatim out of `GeneralSettingsView`, and none of them were covered in the old location either. The line in `UpdateChecker` is the real network path in `fetchJSON`. The one in `SettingsView` is the `case .about` arm, which `test_rpc.sh` exercises live. There is a real gap behind that number, the precedence of error over available update over up to date, but it predates this PR and belongs in a pure value rather than being enumerated through the view, so it is not a condition on this change.

Thanks for the patience with the wait, and for doing the rebase.

PR #670, opened 2026-09-02
feat(app): remove the far end from the microphone track before transcription

pasrom:
Wires the echo canceller into the pipeline. It has been on main since it was vendored, with its model shipping in the bundle, and until now nothing has ever called it. **Off by default.**

## What it does

On a dual-source recording the detector already reports whether the loudspeaker output is coming back through the microphone. On a recording it calls affected, and only there, the far end is removed from the microphone track before anything reads it.

Removing it from the audio rather than from the text is the point. The transcript dedup that shipped earlier could only drop lines; it could not keep a bled-in voice out of the per-track diarization, and through that out of a stored speaker centroid.

## Five things worth reviewing

**The seam became file to file, and aligned.** At 16 kHz Float32 a 90 minute meeting is around 345 MB per track, and the array-shaped call held microphone, reference and output at once. It also declared that sample `i` of both files is simultaneous, which no caller can satisfy: the recorder starts its two captures independently and every neighbour already uses the delta — the detector centres its lag search on it, the mixer prepends silence by it, the merge shifts segment times by it. Fed a reference that does not line up, a canceller is not weaker, it is looking at the wrong audio.

**The microphone track is rewritten in place.** Every later stage opens `mic_16k.wav` by convention, so replacing the file is what makes transcription, diarization and the speaker embeddings all see the cleaned track without a second path threaded through the pipeline. The adoption is three renames inside one directory, each of which either happened or did not, with the step that undoes it written out: `replaceItemAt` documents nothing about where the original is when it throws, and a recovery path whose whole job is to still have the recording must not rest on that.

**Measure first, then repair.** The detector runs on the raw tracks. A cancelled track no longer correlates with the app track, so a detector run afterwards would report every repaired recording as clean and take the embedding quarantine off exactly the audio that needed it.

**The run checks itself, by a difference.** This model sometimes removes essentially nothing while reporting success, for a whole file, and nothing about such a run is visible from outside. The check compares the windows where the reference carried signal against the windows where it did not. Those are the control, not noise to discard: a healthy run leaves them within a fraction of a decibel of untouched, so requiring the echo windows to drop that much further is what separates cancellation from attenuation. A level cannot: a run that simply halves the microphone reports about six decibels wherever the far end plays and leaves the echo exactly as audible as it was. A one-sided floor on the control catches the other direction, a run that pushes the two groups apart by amplifying the quiet ones.

**Only a confirmed run is adopted.** A missing model, a throw, a run that removed nothing, a run that damaged the control, and a run nobody could judge all leave the track as recorded and say which of those happened. The remedy is then resolved from the outcome, so the transcript dedup takes over rather than standing down for a cancellation that did not occur.

## What does not change

The embedding quarantine stays on an affected recording even when cancellation succeeded. Lifting a safety measure on a repair nobody has watched in the field is not something this PR is prepared to argue for.

**The saved recording is not cleaned**, and the setting text says so. What is cleaned is the 16 kHz track transcription and diarization consume; the persisted audio stays what the devices captured, and the mix beside it still comes from the raw originals through the older RMS gate. Whether that should change is a bigger question than this feature, since it decides what the saved artifact of a meeting is.

## Settings

One new switch in Audio, off by default, read live so it takes effect without restarting watching. The transcript-dedup row below it is disabled while it is on: the two do not compose, cancellation wins, and a switch that is on and does nothing is a failure this project has shipped before.

`AppSettings` crosses the file-length cap with the new property and is suppressed rather than split. The honest split is to move its three enums out, and other open PRs are editing that file.

## Testing

`swift test --parallel` with the model present, `scripts/lint.sh`, `swiftlint analyze --strict`, and `scripts/pre-push.sh` release parity, all green locally, plus the four E2E lanes on the self-hosted runner. Every commit was verified to build on its own.

Six mutations were checked, each turning exactly one test red: resetting the streaming context on a block boundary; the previous reference read, which lost what followed the padding; the two-step adoption, which left no microphone track at all; removing the restore after a failed adoption; the previous self-check rule, which confirmed both a uniformly attenuated run and one with no control group; and removing the disabled state from the dedup row.

## Not in this PR

The E2E lane that exercises cancellation itself, and the removal of the transcript dedup this replaces. Both want this landed first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,6 +196,16 @@ jobs:
         with:
           cache-key-suffix: ${{ matrix.variant }}
           swift-flags: ${{ matrix.swift-flags }}
+      # Without this the echo-cancellation tests skip, which is how this
+      # feature reached a review with single-digit patch coverage on the two
+      # files that do the work. The gap is worse than the number: the library
+      # checks a model against a SHA-256 list compiled into it, so a model the
+      # linked library rejects fails only at load time — compiling stays green,
+      # the tests skip, and both builds merely copy the file. `appstore.yml`
+      # already downloads the same model through `build_release.sh`, so this
+      # adds no network dependency that CI did not have.
+      - name: Fetch the echo-cancellation model
+        run: echo "MEETINGTRANSCRIBER_LOCALVQE_MODEL=$(./scripts/fetch-localvqe-model.sh)" >> "$GITHUB_ENV"
       # 18 min step-timeout on the actual tests: typical hot-cache run is
       # 5–7 min, so this is ~3× margin while still failing fast as
       # `timed_out` (not `cancelled`) when a single test hangs.
--- a/app/MeetingTranscriber/Sources/A11yID.swift
+++ b/app/MeetingTranscriber/Sources/A11yID.swift
@@ -39,6 +39,7 @@ enum A11yID {
     static let saveRawTranscriptToggle = "saveRawTranscriptToggle"
     static let outputFolderSection = "outputFolderSection"
     static let vadSection = "vadSection"
+    static let echoCancellationToggle = "echoCancellationToggle"
     static let echoDedupToggle = "echoDedupToggle"
     static let diarizationSection = "diarizationSection"
     static let liveTranscriptionSection = "liveTranscriptionSection"
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -1,3 +1,11 @@
+// swiftlint:disable file_length
+//
+// A settings registry grows by one documented property per feature, so the
+// 600-line cap measures how many features the app has rather than whether this
+// file does too much. Suppressed rather than split because the honest split is
+// to move the three enums below into their own file, and two open pull requests
+// are editing this one; that move is worth doing on a quiet tree, not folded
+// into a feature branch.
 import SwiftUI
 
 // SwiftFormat strips redundant raw values matching their case name, which then
@@ -296,6 +304,14 @@ final class AppSettings {
         didSet { defaults.set(vadThreshold, forKey: "vadThreshold") }
     }
 
+    /// Remove the far end from the microphone audio before transcription
+    /// (`EchoRemedy.cancellation`). Off by default: it has been measured on an
+    /// archive with synthetic loudspeaker echo, never in the field, and it wins
+    /// over the transcript dedup wherever both are on.
+    var echoCancellationEnabled: Bool {
+        didSet { defaults.set(echoCancellationEnabled, forKey: "echoCancellationEnabled") }
+    }
+
     /// Leave the far end out of the transcript when the microphone only picked
     /// it up from the loudspeaker.
     ///
@@ -522,6 +538,7 @@ final class AppSettings {
         diarize = defaults.object(forKey: "diarize") as? Bool ?? true
         vadEnabled = defaults.object(forKey: "vadEnabled") as? Bool ?? false
         vadThreshold = defaults.object(forKey: "vadThreshold") as? Float ?? 0.5
+        echoCancellationEnabled = defaults.object(forKey: "echoCancellationEnabled") as? Bool ?? false
         echoDedupEnabled = defaults.object(forKey: "echoDedupEnabled") as? Bool ?? false
         diarizerMode = (defaults.string(forKey: "diarizerMode")
             .flatMap(DiarizerMode.init(rawValue:))) ?? .offline
--- a/app/MeetingTranscriber/Sources/EchoCancellationSelfCheck.swift
+++ b/app/MeetingTranscriber/Sources/EchoCancellationSelfCheck.swift
@@ -1,0 +1,124 @@
+import Foundation
+
+/// Whether a cancellation run actually removed anything, decided from the
+/// canceller's own per-window measurements.
+///
+/// This exists because "it removed nothing" is a real outcome of this model and
+/// an invisible one. On a minority of recordings it removes essentially nothing
+/// once local speech is present, for the whole file rather than just the
+/// overlapping parts, while removing tens of decibels from the same echo
+/// presented on its own. Nothing about such a run looks wrong from the outside:
+/// it completes, it writes a full-length track, and every gate stays green.
+///
+/// The separation below was measured on recordings that cannot be shared, so
+/// the reasoning is written out here rather than pointed at: what a reader
+/// needs is why these two constants have these shapes, not a citation they
+/// cannot check.
+///
+/// Pure, and separate from the canceller, so the threshold can be chosen from
+/// data without touching the code that produces the audio. That split is the
+/// same one `EchoReductionVerdict` makes for the bundle probe.
+enum EchoCancellationSelfCheck {
+    /// What the run achieved. Three cases and not a `Bool?`, for the same
+    /// reason `EchoVerdict` is not one: *not enough evidence* is not *no*, and
+    /// a caller that collapses them either warns about a recording nobody
+    /// measured or stays quiet about one that failed.
+    enum Effect: Equatable {
+        /// The reduction cleared the threshold. The echo is gone.
+        case removed
+        /// The run completed and took essentially nothing off. The measured
+        /// silent failure.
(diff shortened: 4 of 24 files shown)


pasrom:
Followed up on the coverage report. The number was pointing at something real, and it was not "these lines are hard to test".

**No workflow has ever set the model path**, so every model-gated test skips in CI. Measured on this branch, both ways:

| file | without the model | with it |
|---|---|---|
| `LocalVQECanceller+File.swift` | 41 % | 96 % |
| `LocalVQECanceller.swift` | 20 % | 95 % |

The coverage number is the symptom. What is underneath it is that the linked library checks a model against a SHA-256 list compiled into it, so a library bump that rejects the pinned model fails only when something loads it. Nothing else looks: compiling is green because the C interface is unchanged, the tests are green because they skip, both builds are green because they only copy the file. That reaches a release. `appstore.yml` already fetches the same file through `build_release.sh`, so pointing the test job at it adds no network dependency CI did not have.

Two gaps were mine rather than CI's:

- `WindowAccumulator` was extracted as a value type precisely so it could be tested without the C library, and then only ever reached through the model-gated file tests. It now has its own, including the invariant the doc comment claims and nothing was holding: the block size must not change what the report says.
- The stage had a test named for the missing-model branch that did not exercise it. Injecting a canceller with a bad path is a canceller that throws once it runs, not the absence of one; the two report different things. The factory returns an optional now, so the real branch is reachable. Cancellation is covered too, and it is the one failure that has to stay quiet.

Left uncovered on purpose, and said so at the code: the production model resolution. Every line of it reads the process environment and the main bundle, so a test would be asserting about the machine it ran on. The decision it delegates to is pure and already has tests.

pasrom:
Another pass. Where the remaining lines went, and where they stayed.

| file | before | now |
|---|---|---|
| `LocalVQECanceller.swift` | 96.1 % | **100 %** |
| `PipelineQueue+EchoCancellation.swift` | 86.7 % | **98.3 %** |
| `LocalVQECanceller+File.swift` | 95.6 % | **97.2 %** |
| `EchoRemedy` / `EchoCancellationSelfCheck` | 100 % | 100 % |

Three of them turned out to be worth more than the number.

**The whole chain had no end-to-end test.** Measure, resolve the remedy, replace the microphone track, stand the dedup down, say what happened: four assertions in three files, and nothing checked they compose in the order the pipeline runs them. The three uncovered lines in `PipelineQueue+Stages.swift` were exactly that call. The discriminating part is the dedup standing down, and it is not a tautology: it is the same fixture and the same mock segments that make the dedup suppress exactly one line in an existing test, so under cancellation the transcript must keep both.

**The library-refusal path was untested,** and it is the one that matters most of the three failure modes: without the throw, a mid-recording refusal leaves a short file that looks exactly like a finished one, and the pipeline transcribes the truncated track with nothing noticing. Provoked with a frame length the library rejects, verified first that it refuses cleanly rather than crashing.

**The production model resolution read ambient state inline,** so its one decision was reachable only by arranging the machine the test ran on. Its three reads are default arguments now, the same way `LocalVQEModel.resolve` already takes its `fileExists`. Production is unchanged.

Also covered: a cancelled run leaves no staging file beside the destination.

**What stays uncovered, and now says why at the code:** three `AVAudioPCMBuffer` guards. They are unreachable as their two callers ask (both rule out a zero capacity first, and the format is the one the file handed back), and they stay in rather than becoming force-unwraps, because a crash is a worse way to learn the assumption stopped holding.

`swift test --parallel`, `scripts/lint.sh` green locally.

pasrom:
Two independent reviews, and between them six defects. All addressed; six commits pushed. What each one was, and what it turned into.

**The self-check proved attenuation, not removal.** Both reviewers landed on this from different directions, and the sharpest form is: a run that simply halves the microphone reports about six decibels in every window carrying echo, passes, and leaves the echo exactly as audible next to the local voice as it was. The check now asks for a *difference* between the windows carrying echo and the windows carrying none. Those were the control all along and were being discarded as "no evidence" — a healthy run leaves them within a fraction of a decibel of untouched, which is what gives the difference its room. It separates as cleanly as the level did and rejects what the level waved through.

**The reference was never aligned with the microphone.** The seam declared sample `i` of both files simultaneous. No caller can satisfy that: the recorder starts its two captures independently, and every neighbour already uses the delta — the detector centres its lag search on it, the mixer prepends silence by it, the merge shifts segment times by it. The canceller alone got two unaligned paths with the value sitting at the call site. Fed a reference that does not line up, it is not a weaker canceller, it is looking at the wrong audio. The seam aligns now, and the test asserts the pair through the self-check: aligned, confirmed; unaligned, not.

**Cancellation was swallowed.** `CancellationError` became "the feature declined" and the run continued. The job is already gone from the queue by then, so its state updates and warnings go nowhere, while both transcription passes and the stage that saves the transcript and moves the recordings still run. A user who cancelled a meeting got a finished job's artifacts and no job to explain them. Rethrown.

**An unconfirmed run was adopted anyway.** The original was deleted and the model's output moved into place before the self-check was consulted, while the warning beside it said the track had been left as recorded. The report is in hand before the swap, so the ordering was free. Anything short of a confirmed removal now leaves the recording as it was.

**A failed cancellation took the dedup down with it.** The remedy was resolved from the settings before the run. With both switches on and no model, the user got neither remedy on every affected recording. The dedup stands down for one reason — the far end is no longer in the audio its measure was calibrated on — and when cancellation does not happen, that reason is gone. Two questions now, two functions: what the settings ask for, and what the recording got.

**The setting did nothing until watching was restarted.** The queue copies its settings when it is built, and it is built once. For a feature that ships off by default, the first thing anyone does with it was the thing that did not work. Read through a closure now. Its neighbours are still copies; that is the same latent problem for each, and widening it is not this change's to make.

**The setting text overclaimed.** It said "from the recording"; what is cleaned is the track transcription and diarization consume. The saved recording is the one the devices captured, and the cleaned intermediate is deleted once naming resolves. Whether that should change is a bigger question than this feature — it decides what the saved artifact of a meeting is — but a setting describing the outcome a user would prefer while delivering another one is not defensible, so the text now says which audio it touches and which it does not.

`swift test --parallel`, `scripts/lint.sh`, `swiftlint analyze --strict` and `scripts/pre-push.sh` release parity all green locally.

pasrom:
Second round of review on the six fix commits. Both reviewers ran again; between them, three more defects, two of them blockers. Fixed in three commits.

**The negative reference lead lost what followed it.** Both found it, independently, with the same mechanism. The padding branch read a whole block from the reference and trimmed it to fit the zeros in front; the trimmed frames were gone, because the file had already advanced past them. The head of the recording was padded correctly and every block after it arrived early by exactly the amount the padding was meant to correct. Both signs occur: `micDelay` is the microphone's first frame minus the app track's, the clamp passes negatives through, and the mixer carries an explicit branch for them. My only alignment test used a positive lead, so the whole sign was untested. The new test turns red against the previous read.

What kept it from being worse is worth stating: an unaligned run fails the self-check, so the track was never adopted. The cost was a recording the feature could have repaired being reported as unconfirmed.

**Damaging the control group was rewarded.** One reviewer checked this case and cleared it; the other found the adjacent one that gets through. A run barely clearing the level test with the quiet windows amplified twenty decibels reads as a large difference and was confirmed — a run that is not cancelling anything, and is making the recording worse precisely where the local speaker is alone.

The floor is one-sided, and measured that way. Healthy runs do attenuate the control a little, occasionally by around ten decibels, which the difference already handles by shrinking. None of them amplified it at all. A symmetric bound would have rejected the healthy tail; this one sits just below zero, where no healthy run has been.

**Two smaller ones.** Moving the judge before the swap also moved the swap out of the block that caught its errors, so a failed rename failed the job — upside down, since the original is still there and usable. And a propagated cancellation left its job ID in the cancelled set, which is what tells a later generic error apart from a cancellation; nothing threw cancellation out of a stage until this feature did, which is what made that leak reachable.

Everything else in both reports came back clean or as notes: the shared threshold constant is safe because the control barely moves on a healthy run (with the caveat now written at the constant: a model that denoised broadly would break that, and would break it by refusing to confirm runs that worked); the stubs still discriminate and each new self-check test turns red against the old rule; the rethrow reaches the right handler with nothing half-done; the live-setting closure has no cycle and no cross-actor read.

`swift test --parallel` and `scripts/lint.sh` green locally.

PR #675, opened 2026-09-04
fix(app): keep offline diarization off Metal

bastiank:
## Summary

- route FluidAudio offline diarization through CPU + Neural Engine
- avoid the Metal/MPSGraph path whose in-process abort bypasses Swift error handling
- pin the compute-unit policy with a regression test

## Evidence

A symbolicated crash ended in `OfflineEmbeddingExtractor.runEmbeddingModel` above Metal/MPSGraph/Core ML. The process received SIGABRT, so pipeline retry logic cannot contain it. CPU + Neural Engine retains hardware acceleration without invoking that GPU path.

## Validation

- `git diff --check`
- local Swift tests unavailable: repository requires Swift 6.2; this host has Swift 6.1
- CI is the build/test authority for this change

### DIFF
--- a/app/MeetingTranscriber/Sources/FluidDiarizer.swift
+++ b/app/MeetingTranscriber/Sources/FluidDiarizer.swift
@@ -1,3 +1,4 @@
+import CoreML
 import FluidAudio
 import Foundation
 import os.log
@@ -283,6 +284,17 @@ struct FluidOfflineProcessor: OfflineDiarizationProcessing {
         return config
     }
 
+    /// Keep FluidAudio's offline inference off Metal. Core ML's `.all`
+    /// routing can send the WeSpeaker embedding model through MPSGraph, whose
+    /// in-process abort cannot be caught or retried by the pipeline. The CPU +
+    /// Neural Engine route retains hardware acceleration without invoking the
+    /// crash-prone GPU path.
+    static func makeModelConfiguration() -> MLModelConfiguration {
+        let configuration = MLModelConfiguration()
+        configuration.computeUnits = .cpuAndNeuralEngine
+        return configuration
+    }
+
     mutating func prepare(numSpeakers: Int?) async throws {
         guard manager == nil || numSpeakers != currentNumSpeakers else { return }
 
@@ -296,7 +308,7 @@ struct FluidOfflineProcessor: OfflineDiarizationProcessing {
                 "FluidAudio offline tuning (\(flag)): clusterThreshold=\(t.clusterThreshold) warmStartFa=\(t.warmStartFa) warmStartFb=\(t.warmStartFb) minSegmentDurationSeconds=\(t.minSegmentDurationSeconds) excludeOverlap=\(t.excludeOverlap)",
             )
         let newManager = OfflineDiarizerManager(config: config)
-        try await newManager.prepareModels()
+        try await newManager.prepareModels(configuration: Self.makeModelConfiguration())
         manager = newManager
         currentNumSpeakers = numSpeakers
         logger.info("FluidAudio offline models ready")
--- a/app/MeetingTranscriber/Tests/FluidDiarizerTuningTests.swift
+++ b/app/MeetingTranscriber/Tests/FluidDiarizerTuningTests.swift
@@ -1,3 +1,4 @@
+import CoreML
 import FluidAudio
 @testable import MeetingTranscriber
 import XCTest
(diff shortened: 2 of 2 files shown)


pasrom:
Thanks for tracking this down, and for keeping the patch this small. The direction is right and I want it in. Two things I need from you; the measuring I have done.

**The change is mechanically correct.** I checked it against FluidAudio v0.15.6, which is what `main` pins. `OfflineDiarizerManager.prepareModels(directory:configuration:forceRedownload:)` exists, and `OfflineDiarizerModels.load` reads exactly one thing off the configuration, `configuration?.computeUnits ?? .all`. So this flips segmentation, embedding and PldaRho from `.all` to `.cpuAndNeuralEngine`; FBank is hardcoded `.cpuOnly` upstream either way.

Worth adding to the rationale, because it is stronger than the argument in the description: FluidAudio's own ANE profiling of this pipeline (`Documentation/ANE_Profiler.md`) measures the offline embedding model at 93% ANE, 0% GPU, 7% CPU under `.cpuAndNeuralEngine`, and `runEmbeddingModel` already calls `prefetchToNeuralEngine()` on its inputs. That path is written for the ANE.

**Measured, since CI cannot do it here.** The unit lanes never touch these models, and the quality lane is restricted to same-repo branches and runs on VMs without an ANE, so it would exercise neither routing. I ran the three offline quality fixtures locally on an M3 Max, macOS 26.6.2, comparing your branch against the same branch with only the `configuration:` argument removed, twice per routing, reporting the warm pass. FluidAudio logs the routing it actually used, and the runs are cleanly separated.

| Fixture | DER `.all` | DER `.cpuAndNeuralEngine` | Wall `.all` | Wall `.cpuAndNeuralEngine` |
|---|---|---|---|---|
| two_speakers_de | 0.569670 | 0.569670 | 0.340 s | 0.257 s |
| three_speakers_de | 0.551700 | 0.551700 | 0.460 s | 0.348 s |
| four_speakers_en_ami | 0.361107 | 0.361107 | 1.250 s | 0.837 s |

Identical DER to every digit reported, and 24% to 33% faster. One caveat for expectations: on the cold pass the AMI fixture was slower under your routing (1.612 s against 1.307 s), which is the one-time program compile for a new compute-unit selection. One machine, one OS, warm models, so this bounds the accuracy and speed risk, it does not prove the abort is gone.

**First ask: the crash report, or at minimum the faulting frames.**

Two upstream crash families point opposite ways, and the frames tell them apart:

* FluidInference/FluidAudio#843 (closed; a Kokoro TTS graph on an iPadOS 27 beta, not this model): SIGABRT with Metal and MetalPerformanceShadersGraph frames above Espresso. Upstream's fix was to route the aborting stages off the GPU, after which Metal is never entered. That is the mechanism your description names, and the one this patch removes.
* FluidInference/FluidAudio#878 (open): `EXC_BAD_ACCESS` / SIGBUS inside `libBNNS`, no Metal frames, in this offline diarizer, reported against macOS 14 under the compute units this PR selects. That reporter's runner matrix crashed on every macos-14 run and none on 15 or 26, and they saw it intermittently on physical Sonoma hardware. If your frames are BNNS rather than MPSGraph, this patch does not address it, and the fix upstream points at is an OS update.

Your description reads like the first family. The frames settle it, and I would rather read them than infer them.

**Second ask: a test that guards the regression.**

`testModelConfigurationAvoidsMetalInference` asserts that the factory returns what the factory sets. Delete `configuration: Self.makeModelConfiguration()` from the call site in `prepare()` and it stays green, which is exactly the edit that would put the crash back. Codecov says the same from outside: the whole `prepare()` body is unhit, that call included. No lane that runs on this PR executes it; the unit tests inject a mock offline processor, and the two lanes that drive the real one, the quality job and the self-hosted E2E job, are opted in by env var and restricted to same-repo branches.

A seam is enough: a stored closure `(OfflineDiarizerManager, MLModelConfiguration) async throws -> Void` defaulting to `{ try await $0.prepareModels(configuration: $1) }`, with `prepare()` passing `Self.makeModelConfiguration()` through it. The spy takes the real `OfflineDiarizerManager` (its init loads nothing) and just records `computeUnits` without calling `prepareModels`, so no model download. Put it in `FluidDiarizerTuningTests`, which is not main-actor isolated; from a `@MainActor` test you would have to drive it through `FluidDiarizer(mode: .offline, offlineProcessor:)` instead. Acceptance is the usual one: revert the fix and the test goes red.

**Small things**

* Please note in the doc comment that only `computeUnits` is read downstream, and that it is honoured only from FluidAudio 0.15.5 onwards. In 0.15.4 the value was hardcoded `.all` and the argument was silently ignored, so a dependency downgrade past that point would revert the fix with no diagnostic.
* Not asking you to widen the scope; the three other `.all` load sites in the app are my call, separately. Your change targets the path your trace names.

PR #613, opened 2026-08-12
docs: document echo mitigation, shared vocabulary, and /v1 watch/record

pasrom:
## Summary

CLAUDE.md, README.md and docs/architecture-macos.md had fallen behind the code. The architecture doc's file tables were missing 39 source files that have landed since (the echo-mitigation stack around `LocalVQECanceller`, the shared-vocabulary files, the `/v1/watch` and `/v1/record` control surfaces, the manual-recording types, `ChannelFaultMonitor`, `AppLauncher`) and still listed the deleted `MicRecorder.swift`; CLAUDE.md still described custom vocabulary as a Parakeet-only feature and said nothing about terminology normalization, the channel-fault notification split, the transcript-retention settings, or the `/v1/watch` and `/v1/record` routes.

What the docs now say:

- **Custom vocabulary and terminology:** one shared vocabulary file (`AppSettings.customVocabularyPath`, resolved through `VocabularyFileAccess`) feeds both engines: Parakeet via CTC boosting, WhisperKit via the experimental, off-by-default decoder-prompt hint (`WhisperVocabularyPrompt`). `TerminologyNormalizer` is documented as a separate, engine-independent post-ASR pass. Covered in CLAUDE.md, the README feature list and settings table, and a new architecture-doc section.
- **Echo mitigation:** the architecture doc gains a section summarizing detection (`EchoBleedDetector`), the two non-composing remedies decided by `EchoRemedy` (acoustic cancellation via the vendored `CLocalVQE` library vs. transcript dedup), and the embedding quarantine, and points at CLAUDE.md's Architecture Notes for the full rationale. The README gains an "Echo bleed handling" bullet and lists echo cancellation in the Audio settings row.
- **Channel fault vs. channel quiet:** CLAUDE.md now explains why `ChannelFaultMonitor` (buffer ages, drives the "Capture Channel Silent" notification) is deliberately separate from `ChannelHealthMonitor` (levels, drives the menu-bar tint), per issue #614.
- **Automation API:** the `/v1` documentation in CLAUDE.md and the architecture doc now includes the idempotent `GET`/`POST /v1/watch` and `/v1/record` resources behind Stream Deck and hotkey control (`docs/stream-deck.md`).
[...]

Issue #588, opened 2026-08-07
[Bug] A Bluetooth headset disconnecting mid-recording freezes the entire app and burns a core


pasrom:
## What happens

If a Bluetooth headset is the input device during a recording and it disconnects, the app freezes completely. Not just the recording: the menu bar stops responding, meeting-end detection stops, no notification is ever shown, and the process settles at about a third of a core with coreaudiod adding roughly another two thirds. It stays that way until the headset reconnects or the app is killed.

Observed in the field on v0.6.0 (macOS 26.6, Apple silicon) for three hours before the headset came back, which is what released it. Then reproduced deliberately on the first attempt.

## Reproduction (deterministic)

1. Connect a Bluetooth headset and make it the system default input device.
2. Start a recording.
3. Switch the headset off, or put it in its case.

Within a second the app is wedged.

Note for anyone trying to reproduce this outside the app: a passive test program is not enough. If the engine was never actually started with a tap installed, a fresh engine simply builds a new default-device aggregate over the healthy built-in devices and returns normally. The failure needs the poisoned aggregate to still be in use by a running IO unit.

## What is going on

The mic restart path handles the device change, discards the engine, creates a fresh `AVAudioEngine`, and reads `inputNode`. That call never returns. It is not blocked, it is executing: an unbounded retry inside AVFAudio, issuing thousands of completed CoreAudio round trips per second.

```
MicCaptureHandler.scheduleRestartRetry closure
  MicCaptureHandler.executeRestart
    MicCaptureHandler.startEngine
      -[AVAudioEngine inputNode]
        AVAudioEngineImpl::UpdateInputNode
          AVAudioIOUnit::GetHWFormat
            AVAudioIOUnit_OSX::_GetHWFormat + 548
              AVAEHalUtil::GetSubDevices
                AudioObjectGetPropertyData -> HAL -> coreaudiod
```

All samples of a 10 second `sample` run sit in that single call, and `spindump` reports 3.3 s of CPU on that one thread in a 10 s window, the only thread in the process consuming anything.

The trigger is that after the headset vanishes, the per-process default-device aggregate keeps dangling sub-device references to it. Measured directly with a small CoreAudio probe right after a disconnect:

```
CADefaultDeviceAggregate-<pid>-0 sub-devices:
  <headset addr>:input   (no such device)
  <headset addr>:output  (no such device)
```

The log stops mid-restart and never resumes:

```
Mic input device: name=MacBook Pro Microphone
[...]

PR #689, opened 2026-09-06
ci: run ASan on every PR, gate TSan on the diff, add sanitizer-gate

### LINKED ISSUE #688: The sanitiser lane is not a default PR check, so a PR can be green and still break main
Observed on 2026-09-06: `main` went red in `Quality & Safety` from a change whose PR had reported thirteen green checks. The sanitiser lane was not among them.

## What happens

`quality-and-safety.yml` runs its jobs only when the PR carries the `run-quality` label. Without it, `quality-baseline` and the `test (tsan)` / `test (asan)` matrix report as `skipping`. `gh pr checks` and the GitHub UI then show an all-green PR, and the branch protection is satisfied, because a skipped job is not a failing one.

The lane still runs on push to `main` and nightly. So the first machine to execute those tests is `main` itself.

## Why it mattered here

Four PRs merged in one session (#680, #681, #682, #684). None carried `run-quality`. The defect that broke `main` was a flaky test that aborts only when a dispatch queue does not drain between two iterations of a poll loop, which needs a slow build: the sanitiser build is slow enough, the ordinary one is not. It had passed twice on `main` before it aborted, once on a merge and once on a nightly, so even running the lane would not have caught it every time.

The fix is #687. This issue is about the gap it went through, not the defect.

## Why this is not simply "remember the label"

The label exists for a good reason: the sanitiser matrix is slow, and running it on every PR would cost a lot of runner time for a rare class of finding. That trade is sound. What is unsound is that its absence is invisible at the moment a PR is judged: nothing distinguishes "the sanitiser passed" from "the sanitiser was never asked", because both render as a green PR.

## Options, none obviously right

1. **Run it on every PR that touches concurrency-adjacent code.** A `paths` filter on `tools/audiotap/**` plus anything with a lock, a queue or `@unchecked Sendable` would have caught this one. Cheap in the common case, and the filter is a guess that will drift.
2. **Make the skip visible.** A required check that fails, rather than skips, when the lane did not run on a PR that touches those paths. Turns a silent gap into a decision.
3. **Run it on the merge queue rather than on the PR.** Costs one run per merge instead of one per push, and still catches this before `main`.
4. **Leave it and rely on nightly.** Defensible if `main` being red in that lane for up to a day is acceptable; it is what happens today, and it is worth saying out loud rather than assuming.

The flakiness angle deserves its own note: a lane that only sometimes reproduces a defect is a weak gate wherever it runs. The deeper fix for that class is to stop writing tests whose correctness depends on scheduling, which is what #687 does for this one.

pasrom:
The sanitizer matrix in `Quality & Safety` was label-gated, so an unlabeled PR reported green whether the sanitizers passed or were never asked. A skipped job and a passed one looked the same at review time. `main` was the first machine to run them and went red from a change whose PR carried thirteen green checks (issue #688).

Four changes, one commit, because they only hold together: a gate is safe to require only once ASan is unconditional and TSan's triggers are stated, and the triggers are only worth stating once a check can bind on them.

## 1. ASan on every same-repo PR

3m12s for both packages when measured on the Mini, cheaper than the required GitHub-hosted test jobs, so it needs no gate at all. Only TSan stays gated.

## 2. TSan on the label, the diff, or a manifest change

TSan runs on the `run-quality` label, when the PR's changed source lines add, remove or move a concurrency primitive (`git diff -G` in the `changes` job, checked out at `refs/pull/N/merge` with depth 2), or when the PR touches `Package.swift` / `Package.resolved` of either sanitized package or this workflow file.

The manifest arm is evidenced by #657: a Dependabot bump of FluidAudio 0.15.5 to 0.15.6 in `Package.resolved` and nothing else. The new version linked a Rust archive whose exception personality was a fourth in the image where Apple's compact unwind format encodes three, and ThreadSanitizer injects one of the three, so the TSan link failed while the regular and ASan builds kept working (#659 has the count). `main` stayed red from 2026-08-26T21:32 until #659 landed on 2026-08-27T18:32, the push run and the nightly both cancelled by the failing leg. No keyword can see a version number. The workflow file is on the list because a change to the TSan job can be exercised before merge only by running it; proving the keyword arm actually started the legs on this PR took a throwaway commit with a keyword line. Left off on purpose: the new composite action and `verify-xcode` (the unconditional ASan leg executes both on every same-repo PR and neither carries anything TSan-specific), `cancel-on-failure` (never runs on a passing test), and the manifests of mt-cli and meeting-simulator (the legs do not build them).

**Two jobs, not a computed matrix.** A job-level `if:` applies to a whole matrix, so one leg unconditional and one gated needs either a matrix computed in `changes` and consumed through `fromJson`, or two jobs. Two jobs, with the steps in one composite action (`.github/actions/sanitizer-tests`) so the legs cannot drift apart in anything but the flag. The computed matrix is worse on three counts: a static `include` entry whose `variant` is missing from the computed list is appended as a new combination (documented include semantics), so the flag mapping would have to be emitted as JSON from a shell step; the label arm would have to be duplicated into that step, so the TSan decision would live in two places; and the gate would see one aggregate result and could not say which leg failed. The check names on a run that executes stay `test (asan)` and `test (tsan)`.

## 3. `sanitizer-gate`, the check to require

Runs on every event under `always()`, needs `changes` and both legs, and fails when:

- a leg reports `failure` or `cancelled`;
- `changes` did not succeed (the legs are then skipped through `needs` and nothing was checked);
- a leg this event should have run reports `skipped`, which means the legs' `if:` and the gate disagree and the workflow is misconfigured.

`cancelled` counts as failure on purpose. The cancel-on-failure step cancels the run when a leg fails, and the run-level cancellation races the culprit's own conclusion, so the failing leg usually ends up marked `cancelled` like its victims (the push run for #684 on 2026-09-06: `test (tsan)` cancelled, `test (asan)` success). A gate that excused `cancelled` would excuse most sanitizer failures. `always()` and not `!cancelled()`: after that cancellation only `always()` still starts the job; a skipped job reports success and would satisfy the required check on exactly the runs it exists for.

The summary is the other half of #688. It names each leg's result and, for a leg that did not run, why:

```
| leg  | result |
| ASan | passed (runs on every same-repo PR and on push, schedule and dispatch) |
| TSan | not run: not asked: no run-quality label, no concurrency primitive on the changed source lines, no package manifest or sanitizer workflow change |
```

so a green gate on a PR where TSan was not asked reads as "ASan passed, TSan not asked", never as "the sanitizers passed".

**After merge:** add `sanitizer-gate` to the required status checks on `main`. With `enforce_admins` off it binds non-admin merges and Dependabot auto-merge; `gh pr merge --admin` bypasses it. No repo setting is changed by this PR.

## 4. Four capture-path keywords

`installTap(` (the AVAudioEngine tap block, called on the render thread), `AudioDeviceCreateIOProcIDWithBlock` (the IOProc block on the real-time I/O thread), `AudioObjectAddPropertyListenerBlock` (the listeners that start capture restarts from a CoreAudio thread, #588) and `CheckedContinuation`, spelled `Checked(Throwing)?Continuation` because the repo writes `withCheckedThrowingContinuation` too. `installTap` is anchored to a call or declaration site: the bare word sits in sixteen comment lines about the installTap NSException recovery and matched two PRs in the sample on prose alone (#639, #645). None of the four added a PR to either sample; they are there for the PR that rewires a tap, an IOProc or a listener without touching a lock line, the shape of the #687 near miss.

## Runs that are no verdict for their commit

Every `pull_request` event starts a run on the PR's head SHA, and `pr-labels.yml` labels each new PR, so an `opened` run is followed within seconds by a `labeled` run on the same SHA (30 of 264 SHAs in the last 300 PR runs of this workflow carried more than one run, median gap 9 s; eight gaps exceeded ten minutes, from labels added later such as `run-e2e`). The legs skip on a `labeled` event for any label but `run-quality`, as before. A skipped job reports success, GitHub lists same-named check runs from both runs in the PR's rollup (this PR's head shows `conventional-title` twice), and a required check named `sanitizer-gate` on the second run could therefore stand in for a red one on the first. On such a run `changes` sets `verdict=false` and the gate runs anyway, under the name `sanitizer-gate [no verdict: label event]` that `changes` hands it, and reports that nothing was evaluated. It runs rather than skips because a skipped job's `name` expression is never evaluated ([actions/runner #1215](https://github.com/actions/runner/issues/1215)): a skipped gate would carry the raw expression text as its check name, which happens not to collide with `sanitizer-gate` today and would start colliding the day that is fixed. The legs keep static names for the same reason, so a TSan leg that was not asked shows as a skipped `test (tsan)` and not as an expression; on a label-event run their skipped rows sit next to the verdict run's rows in the rollup, which is clutter but not a verdict. Re-running the legs on every label event would close the same hole at twice the Mini's sanitizer load on every PR opening.

## Numbers, recounted after all four changes

Over the forty PRs merged before the session that produced #688 (#605 to #613 by merge time; 33 from same-repo branches, 7 from forks, where no PR run happens at all), counting a PR when the arm would have fired on its diff:

| arm | fires | of which same-repo |
|---|---|---|
| path filter `tools/audiotap/**` | 4 | 4 |
| keyword arm, 17 entries (before this PR) | 11 | 9 |
| keyword arm, 21 entries (now) | 11 | 9 |
| manifest arm | 5 | 5 (4 not also keyword hits) |
| TSan, any arm | 15 | 13 of 33 |
| ASan | | 33 of 33 |

Over the five PRs of the session (#680, #681, #682, #684, #687): path filter 4, keyword arm 3, manifest arm 0, TSan 3, ASan 5.

The earlier description and correction said 10 for the keyword arm; both counting methods give 11 over this window. #605, the oldest PR in it, is a hit (an `actor` declaration and two `CheckedContinuation` lines) and is also a fork PR. Every one of the eleven carries a real construct, including the four single-line matches (#651 an `@unchecked Sendable` struct, #637 a `nonisolated(unsafe)` static, #635 an `actor` declaration, #632 a `Task.detached`); ten match a lock, queue, actor or continuation keyword and only #634 matches on pointers alone. The four manifest-only hits are the 0.15.6 bump (#657) and its three repairs (#658, #659, #662); nothing in the sample touched this workflow.

So where the label ran the matrix on none, TSan now runs on about two same-repo PRs in five and ASan on every one. That is a cost, stated as one. What it buys is that the TSan runs are about concurrency or about the link.

Counted from `gh pr diff` and from `git diff -G` over each PR's rebased commit range on `main`, which agree on every PR; the step script itself, run on synthetic two-parent merge commits of all 45 PRs, fires on exactly those 18.

## Verification

- `actionlint` clean, both YAML files parse, `shellcheck` clean on all eight `run` blocks of the workflow and the composite, `./scripts/lint.sh` reports 0 violations in 533 files.
- The gate script, extracted from the workflow, was executed under 67 combinations of event and job results (push, schedule, dispatch with and without `run-sanitizer`, same-repo PR with and without label or diff match, fork PR, `changes` failed or cancelled, label-event runs, every leg result including an unknown string); every exit code and summary line came out as designed.
- Truth table by hand for every event and job: push and schedule run both legs and the gate; dispatch runs the legs iff `run-sanitizer` and the gate always; a same-repo PR runs ASan always, TSan on label or arm, the gate always; a fork PR runs neither leg and a gate that says so; a `labeled` event for any label but `run-quality` skips both legs and runs the gate under the `[no verdict: label event]` name; a failing `changes` skips both legs and turns the gate red.
- `quality-baseline` is byte-identical to `main`.

## What this does not do

Close #688. ASan on every PR says nothing about data races, TSan is asked only when the label, the keywords or the manifest list fire, and the keyword list is a heuristic with false negatives (#687's production change moved a `withLock` call without touching a keyword line). The defect behind #688 passed the sanitizers twice on `main` before it aborted, so even a green TSan leg is evidence, not proof.

Refs #688

### DIFF
--- a/.github/actions/sanitizer-tests/action.yml
+++ b/.github/actions/sanitizer-tests/action.yml
@@ -1,0 +1,72 @@
+name: "Sanitizer tests"
+description: >
+  Runs the app package's and AudioTapLib's test suites under one sanitizer
+  on a self-hosted Mini. Shared by the `test (asan)` and `test (tsan)` jobs
+  of quality-and-safety.yml: the two legs are separate jobs because a
+  job-level `if:` applies to a whole matrix and ASan runs on every same-repo
+  PR while TSan is gated (issue #688), and one composite keeps them from
+  drifting apart in anything but the flag. The caller runs actions/checkout
+  and verify-xcode first, so DEVELOPER_DIR is already exported here.
+
+inputs:
+  sanitizer-flag:
+    description: "The swift test flag: --sanitize=address or --sanitize=thread."
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    # `mt-ci.keychain-db` is pre-staged by setup-self-hosted-runner.sh, but
+    # macOS occasionally drops it from the per-user search list (seen after
+    # a second runner's `config.sh` ran as the same OS user). When that
+    # happens `SecItemAdd` writes to the default keychain while
+    # `SecItemCopyMatching` searches login.keychain only, and the
+    # KeychainHelper tests silent-fail. Re-assert per job; safe when both
+    # legs run at once because every job writes the same value.
+    #
+    # `set-keychain-settings -lut 36000` raises the auto-lock-on-idle
+    # timeout to 10 h. The default ~5 min fired mid-run on the Mini
+    # (sanitizer compile ~4 min + slower tests pushed read calls past
+    # the threshold), causing `SecItemCopyMatching` to silent-fail with
+    # `errSecAuthFailed` between two consecutive reads in
+    # `testOpenAIAPIKeyViaKeychainHelper`. Without `-l` the keychain
+    # also doesn't lock on system sleep, irrelevant inside a CI job
+    # but defensive.
+    - name: Ensure CI keychain is default + in search list + unlocked + no-autolock
+      shell: bash
+      run: |
+        MT_CI_KEYCHAIN="$HOME/Library/Keychains/mt-ci.keychain-db"
+        security default-keychain -s "$MT_CI_KEYCHAIN"
+        "$GITHUB_WORKSPACE/scripts/keychain-prepend.sh" "$MT_CI_KEYCHAIN"
+        security unlock-keychain -p "" "$MT_CI_KEYCHAIN"
+        security set-keychain-settings -lut 36000 "$MT_CI_KEYCHAIN"
+    - name: Swift tests
+      shell: bash
+      env:
+        SANITIZER_FLAG: ${{ inputs.sanitizer-flag }}
+      run: |
+        cd "$GITHUB_WORKSPACE/app/MeetingTranscriber" || exit 1
+        swift test --parallel --skip ModelPreloadTests "$SANITIZER_FLAG"
+    # AudioTapLib is a separate SPM package, so the app's test run never
+    # compiles its test target and the sanitizers never saw it. That is the
+    # wrong way round: it holds the capture-restart machinery, which is the
+    # most lock- and queue-dense code in the repo (an unfair lock behind every
+    # arbiter decision, a serial restart queue, a real-time IOProc block and a
+    # render-thread tap block), and it is the code the ASan job comment names
+    # as the reason ASan is run at all.
+    #
+    # Worth being honest about the cost, because it is not the build. The
+    # package compiles quickly next to the app, but its wedge tests wait out
+    # real `RestartArbiter.attemptTimeout` deadlines (5 s, a static let, not
+    # injectable), and within a class XCTest runs serially. Budget roughly
+    # 1.5 to 2.5 min per sanitizer leg, which still sits well inside the
+    # jobs' timeout. What that buys: those same tests drive real queues and
+    # semaphores, so TSan gets genuine interleavings rather than a
+    # single-threaded walk.
+    - name: Swift tests (AudioTapLib)
+      shell: bash
+      env:
+        SANITIZER_FLAG: ${{ inputs.sanitizer-flag }}
+      run: |
+        cd "$GITHUB_WORKSPACE/tools/audiotap" || exit 1
+        swift test --parallel "$SANITIZER_FLAG"
--- a/.github/workflows/quality-and-safety.yml
+++ b/.github/workflows/quality-and-safety.yml
@@ -1,28 +1,47 @@
 name: Quality & Safety
 
-# Background QA workflow — runs the expensive Swift checks that don't
-# block PRs:
-#   - TSan/ASan sanitizer sweeps (catches races/UB the static checker
-#     can't see; ~10–15 min per leg).
+# Background QA workflow for the expensive Swift checks:
+#   - TSan/ASan sanitizer sweeps (catch the races and memory errors the
+#     static checker cannot see; on the Mini ASan takes about 3 min, TSan
+#     about 15).
 #   - WER/DER quality regression (downloads ~1 GB of model weights and
 #     runs real ASR, then diffs the result against the committed baseline
 #     in Tests/Fixtures/quality/quality-baseline.json; fails on regression).
 #
-# Triggers: push to main, nightly cron, explicit manual dispatch, and
-# label-gated PR runs for SAME-REPO branches only. Fork PR code never runs
-# here: the sanitizer leg executes on the self-hosted Mini right next to an
-# unlocked CI keychain, and fork `pull_request` tokens are read-only, which
-# would also break the cancel-on-failure step. Unlabeled PRs skip every job,
-# so day-to-day review iteration stays fast; applying the `run-quality`
-# label runs the same job set a push to main runs (both the sanitizer
-# matrix and the quality gate), giving pre-merge feedback on demand.
+# Triggers: push to main, nightly cron, explicit manual dispatch, and PR
+# runs for SAME-REPO branches only. Fork PR code never runs here: the
+# sanitizer legs execute on the self-hosted Mini right next to an unlocked
+# CI keychain, and fork `pull_request` tokens are read-only, which would
+# also break the cancel-on-failure step.
+#
+# What a PR gets (issue #688). ASan runs on every same-repo PR: at about
+# three minutes for both packages it is cheaper than the required
+# GitHub-hosted test jobs, so it needs no gate. TSan runs on the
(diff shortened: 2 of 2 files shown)


pasrom:
End-to-end proof on the real merge ref, since the numbers in the description come from synthetic clones and the failure mode of this gate is a silent `false`.

A throwaway commit added a source file with one keyword line, and the branch was force-pushed back to the single commit afterwards, so it is not in the history.

Negative control, the workflow-only commit (the pathspec excludes `.yml`):

```
changes            success -> "No concurrency primitives on the changed source lines"
test (tsan/asan)   skipped
quality-baseline   skipped
```

Positive control, the throwaway commit:

```
changes            success -> "Sanitizer matrix requested: concurrency primitives on
                               changed lines in app/MeetingTranscriber/Tests/A11yIDTests.swift"
test (asan)        success  3m12s
test (tsan)        success 15m13s
quality-baseline   skipped
```

Three things this pins that the description could only assert:

- `HEAD^1` resolves under `fetch-depth: 2` on `refs/pull/N/merge`, in real CI and not just in a synthetic clone.
- The sanitizer legs actually start from the diff arm with no `run-quality` label present, which is the whole point of the change.
- The file that triggered it is in the app package, where a `tools/audiotap/**` path filter would have done nothing, and `quality-baseline` stayed skipped, so the gigabyte of model weights is still label-only.

Both legs were green, which is expected on a base that already carries the fix from #687.

pasrom:
Correction to the numbers in the description, which were wrong when first posted.

It said the path filter would have fired 8 times over the forty PRs merged before this session and this filter 11. Recounted, once from the diffs and once from the files API, it is **4** and **10**.

That reverses the cost half of the argument. The five-PR window has the diff filter running the matrix three times against the path filter's four, which reads like a saving; over forty PRs it is 10 against 4, so the honest statement is that this roughly doubles the number of sanitizer runs against a narrow path filter, to about one PR in four, where the label ran it on none.

The case for it is precision, not cost, and that part holds up. All ten firings carry a genuine construct. The four that match on a single line were checked individually:

```
#651  private struct Rule: @unchecked Sendable {
#637  nonisolated(unsafe) private static var didSweepFactoryDefaults = false
#635  actor OutcomeBox {
#632  return await Task.detached(priority: .utility) { ... }
```

Nine of the ten match a lock, queue or actor keyword; only #634 matches on pointers alone, so the ASan entries are not driving the count. The path filter's four include changes with no concurrency in them and never reach the app package.

The description, the commit message and the comment in the workflow now all carry 4 and 10 and state the cost as a cost.

PR #692, opened 2026-09-08
fix(app): fall back to ANTHROPIC_API_KEY when Claude CLI OAuth session is expired

GB-D:
## What changed and why

`ClaudeCLIProtocolGenerator` shells out to `claude -p` relying solely on the CLI's own OAuth session. If that session expires and can't auto-refresh, the subprocess exits 1 with empty stderr — the real error ("Failed to authenticate: OAuth session expired and could not be refreshed") only appears embedded in the stream-json stdout, which the app never surfaces — so protocol generation silently fails with just "Protocol generation failed — transcript saved" and no way to tell why. This hit 8 real meetings in a row on my machine before I noticed.

`buildEnvironment` now accepts an optional `anthropicAPIKey` and injects it as `ANTHROPIC_API_KEY` when not already set in the base environment, read from the login Keychain's `anthropic-api-key` generic password item. The CLI already accepts API-key auth as an alternative to OAuth, so this gives protocol generation a working fallback instead of a hard dependency on an interactive login session that a background subprocess has no way to refresh itself.

## Checklist

- [x] Rebased on `main`, no merge commits
- [x] `./scripts/lint.sh` is clean
- [x] Tests pass: `cd app/MeetingTranscriber && swift test` (full suite, 3065/3065)

### DIFF
--- a/app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
+++ b/app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
@@ -128,12 +128,7 @@
 
             if process.terminationStatus != 0 {
                 let stderrData = await stderrRead
-                let stderrText = String(data: stderrData, encoding: .utf8)?
-                    .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
-                logger.error(
-                    "claude_cli_failed exit=\(process.terminationStatus, privacy: .public) stderr=\(stderrText, privacy: .public)",
-                )
-                throw Self.makeFailureError(exitCode: process.terminationStatus, stderrText: stderrText)
+                throw Self.handleFailure(exitCode: process.terminationStatus, stderrData: stderrData, text: text)
             }
 
             return try Self.validateGeneratedText(text)
@@ -143,6 +138,29 @@
         /// `ProtocolError.cliFailed`.
         static func makeFailureError(exitCode: Int32, stderrText: String) -> ProtocolError {
             .cliFailed(Int(exitCode), stderrText)
+        }
+
+        /// On a nonzero exit, `text` (accumulated from stream-json on stdout)
+        /// often carries the CLI's actual failure reason as a synthetic
+        /// assistant message (e.g. "Failed to authenticate. API Error: 401
+        /// API key is invalid."), while `stderrText` can be empty or
+        /// uninformative for the same failure. Prefer `text` when it has
+        /// content; fall back to `stderrText` otherwise.
+        static func selectFailureMessage(text: String, stderrText: String) -> String {
+            let trimmedText = text.trimmingCharacters(in: .whitespacesAndNewlines)
+            return trimmedText.isEmpty ? stderrText : trimmedText
+        }
+
+        /// Decodes `stderrData`, logs the raw exit/stderr/text, and builds the
+        /// error `generate()` throws on a nonzero exit — pulled out of
+        /// `generate()` itself to keep that function's body short.
+        static func handleFailure(exitCode: Int32, stderrData: Data, text: String) -> ProtocolError {
+            let stderrText = String(data: stderrData, encoding: .utf8)?
+                .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+            logger.error(
+                "claude_cli_failed exit=\(exitCode, privacy: .public) stderr=\(stderrText, privacy: .public) text=\(text, privacy: .private)",
+            )
+            return makeFailureError(exitCode: exitCode, stderrText: selectFailureMessage(text: text, stderrText: stderrText))
         }
 
         /// Trim whitespace from CLI output. Throws `.emptyProtocol` when
--- a/app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
+++ b/app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
@@ -10,6 +10,20 @@
         let claudeBin: String
         let language: String
 
+        /// User-supplied key from Settings → Protocol Generation (`AppSettings.claudeAPIKey`),
+        /// injected only when the user has explicitly typed one in. Not
+        /// auto-discovered: an implicit switch away from the CLI's own OAuth
+        /// login would silently move a healthy subscription session onto
+        /// metered billing, and a stale/wrong key would silently break a
+        /// previously-working install (both measured — see PR #692 review).
+        let anthropicAPIKey: String?
+
+        init(claudeBin: String, language: String, anthropicAPIKey: String? = nil) {
+            self.claudeBin = claudeBin
+            self.language = language
+            self.anthropicAPIKey = anthropicAPIKey
+        }
+
         static let timeoutSeconds: TimeInterval = 600
 
         /// Search paths for Claude CLI binaries.
@@ -37,6 +51,7 @@
             process.environment = Self.buildEnvironment(
                 baseEnvironment: ProcessInfo.processInfo.environment,
                 searchPaths: Self.searchPaths,
+                anthropicAPIKey: anthropicAPIKey,
             )
 
             let stdinPipe = Pipe()
@@ -279,15 +294,27 @@
 
         /// Strip `CLAUDECODE` (avoid nested-session detection by the child
         /// CLI) and prepend `searchPaths` to `PATH` (app bundles inherit
-        /// a minimal `PATH`).
+        /// a minimal `PATH`). `anthropicAPIKey`, when non-empty and not
+        /// already set in `baseEnvironment`, is injected so the subprocess
+        /// can authenticate via API key instead of the CLI's own OAuth
+        /// session. The key is never auto-discovered — it only arrives here
+        /// when the user has explicitly typed one into Settings → Protocol
+        /// Generation (`AppSettings.claudeAPIKey`), so a healthy OAuth
+        /// session is never silently switched to metered billing and a bad
+        /// key can never silently break a previously-working install.
         static func buildEnvironment(
             baseEnvironment: [String: String],
             searchPaths: [String],
+            anthropicAPIKey: String? = nil,
         ) -> [String: String] {
             var env = baseEnvironment
             env.removeValue(forKey: "CLAUDECODE")
             let extraPaths = searchPaths.joined(separator: ":")
             env["PATH"] = "\(extraPaths):\(env["PATH"] ?? "/usr/bin:/bin")"
+            if env["ANTHROPIC_API_KEY"]?.isEmpty ?? true,
+               let anthropicAPIKey, !anthropicAPIKey.isEmpty {
+                env["ANTHROPIC_API_KEY"] = anthropicAPIKey
+            }
             return env
         }
     }
--- a/app/MeetingTranscriber/Sources/AppSettings.swift
+++ b/app/MeetingTranscriber/Sources/AppSettings.swift
@@ -114,6 +114,13 @@ final class AppSettings {
     /// prompt that blocks the run.
     @ObservationIgnored private let apiKeyAccount: String
 
+    /// Keychain account backing `claudeAPIKey`, same injection reasoning as
+    /// `apiKeyAccount` above. Deliberately its own account name, distinct
+    /// from any Keychain item an external tool on the same machine might
+    /// use for its own Anthropic key — the user opts a key into this app
+    /// specifically, rather than reusing a credential another tool manages.
+    @ObservationIgnored private let claudeAPIKeyAccount: String
+
     // MARK: - Apps to Watch
 
     var watchTeams: Bool {
@@ -471,6 +478,22 @@ final class AppSettings {
         }
     }
 
+    #if !APPSTORE
+        /// API key for the Claude CLI protocol generator, entered by the user
+        /// in Settings → Protocol Generation. Empty means use the CLI's own
+        /// `claude login` session; this is never populated automatically.
+        var claudeAPIKey: String {
+            get { KeychainHelper.read(key: claudeAPIKeyAccount) ?? "" }
+            set {
+                if newValue.isEmpty {
+                    KeychainHelper.delete(key: claudeAPIKeyAccount)
+                } else {
+                    KeychainHelper.save(key: claudeAPIKeyAccount, value: newValue)
+                }
+            }
+        }
+    #endif
+
     // MARK: - Output Directory
 
     /// Security-scoped output-dir bookmark. Stored so `@Observable` can track it.
@@ -521,10 +544,12 @@ final class AppSettings {
     init(
         defaults: UserDefaults = .standard,
         apiKeyAccount: String = "openAIAPIKey",
+        claudeAPIKeyAccount: String = "claudeAPIKey",
         defaultOutputDir: URL = AppPaths.downloadsProtocolsDir,
     ) {
         self.defaults = defaults
         self.apiKeyAccount = apiKeyAccount
+        self.claudeAPIKeyAccount = claudeAPIKeyAccount
         self.defaultOutputDir = defaultOutputDir
 
         watchTeams = defaults.object(forKey: "watchTeams") as? Bool ?? true
--- a/app/MeetingTranscriber/Sources/PipelineController.swift
+++ b/app/MeetingTranscriber/Sources/PipelineController.swift
@@ -193,7 +193,11 @@ final class PipelineController {
         switch settings.protocolProvider {
         #if !APPSTORE
             case .claudeCLI:
-                ClaudeCLIProtocolGenerator(claudeBin: settings.claudeBin, language: settings.protocolLanguage)
+                ClaudeCLIProtocolGenerator(
+                    claudeBin: settings.claudeBin,
+                    language: settings.protocolLanguage,
+                    anthropicAPIKey: settings.claudeAPIKey.isEmpty ? nil : settings.claudeAPIKey,
+                )
         #endif
 
         case .openAICompatible:
--- a/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
+++ b/app/MeetingTranscriber/Sources/PipelineQueue+Stages.swift
@@ -812,7 +812,7 @@ extension PipelineQueue {
             }
             stopElapsedTimer()
         } catch {
-            logger.warning("[\(shortID, privacy: .public)] protocol_generation_failed error=\(error.localizedDescription, privacy: .public)")
+            logger.warning("[\(shortID, privacy: .public)] protocol_generation_failed error=\(error.localizedDescription, privacy: .private)")
             addWarning(id: jobID, "Protocol generation failed — transcript saved")
             stopElapsedTimer()
         }
--- a/app/MeetingTranscriber/Sources/Settings/OutputSettingsView.swift
+++ b/app/MeetingTranscriber/Sources/Settings/OutputSettingsView.swift
@@ -124,6 +124,19 @@ struct OutputSettingsView: View {
                 Text("Binary used for protocol generation")
                     .font(.caption)
                     .foregroundStyle(.secondary)
+
+                HStack {
+                    Text("API Key")
+                    Spacer()
+                    SecureField("", text: $settings.claudeAPIKey)
+                        .frame(width: 200)
+                }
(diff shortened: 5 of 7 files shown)


app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -37,6 +37,7 @@
             process.environment = Self.buildEnvironment(
                 baseEnvironment: ProcessInfo.processInfo.environment,
                 searchPaths: Self.searchPaths,
+                anthropicAPIKey: Self.fetchAnthropicAPIKeyFromKeychain(),
```
pasrom: This line is the entire feature, and nothing pins it. All ten new tests target `buildEnvironment` and `parseKeychainOutput`, both pure, plus two failure paths of the lookup. Delete the argument on this line and every one of them stays green while the feature does nothing.

The rule in `CLAUDE.md` is to assert the effect rather than the attempt, so the gap is worth closing with a seam rather than another pure test: hold the lookup as a property (`var apiKeyLookup: () -> String?` defaulting to `fetchAnthropicAPIKeyFromKeychain`), then a test can inject a known value and assert it arrives in the environment the subprocess is actually given.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -279,17 +280,65 @@
...
         static func buildEnvironment(
             baseEnvironment: [String: String],
             searchPaths: [String],
+            anthropicAPIKey: String? = nil,
         ) -> [String: String] {
             var env = baseEnvironment
             env.removeValue(forKey: "CLAUDECODE")
             let extraPaths = searchPaths.joined(separator: ":")
             env["PATH"] = "\(extraPaths):\(env["PATH"] ?? "/usr/bin:/bin")"
+            if env["ANTHROPIC_API_KEY"]?.isEmpty ?? true,
```
pasrom, line 332: The condition here is "no key in the environment yet", never "the OAuth session failed". Combined with the CLI preferring an explicit key over its own login, a present Keychain item silently decides the authentication method for every single generation.

Two consequences, both measured rather than assumed. A healthy subscription session moves onto metered billing with nothing asking the user. And an item that is stale, rotated, or belongs to another org breaks generation that would otherwise have worked: in my runs the CLI did not fall back to OAuth, it simply produced nothing until I killed it, so in the app it would run to `ProtocolError.timeout` and report the same uninformative failure this PR exists to remove.

Injecting after a failed attempt instead of before the first one keeps everything this change wants and drops both.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -279,17 +280,65 @@
...
+                env["ANTHROPIC_API_KEY"] = anthropicAPIKey
+            }
             return env
         }
+
+        // MARK: - Keychain
+
+        /// Reads the `anthropic-api-key` generic password for the current
+        /// user from the login Keychain (same item `markdown.py` and the
+        /// other GHL pipeline scripts already read via `security
```
pasrom: `markdown.py` and the other GHL pipeline scripts do not exist in this repository. I checked word for word before saying so.

This repo is public, so a reader lands on a pointer they cannot follow, and it carries context from your own environment into ours. The mechanism is worth keeping in the comment. Just describe the item instead of the scripts that read it: a generic password named `anthropic-api-key` for the current user in the login Keychain.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -279,17 +280,65 @@
...
+            let stdoutPipe = Pipe()
+            process.standardOutput = stdoutPipe
+            process.standardError = Pipe()
+            do {
+                try process.run()
+            } catch {
+                return nil
+            }
+            let data = stdoutPipe.fileHandleForReading.readDataToEndOfFile()
+            process.waitUntilExit()
```
pasrom: `generate()` is `async`, and this blocks the thread it is running on. This file was deliberately moved off that pattern: its architecture note records that the main subprocess uses a `terminationHandler` feeding an `AsyncStream` specifically so it never blocks on `waitUntilExit()`. This brings it back, and earlier than the call it was removed from.

Normally `security` returns in milliseconds, so this is not a hot path concern. What worries me is a locked Keychain, or an item whose ACL does not list `/usr/bin/security`: then the child waits on a user prompt while this call holds the thread. A menu bar app generating a protocol in the background is exactly where nobody is looking at the screen to dismiss it.

pasrom (changes requested):
Thanks for this, and sorry it sat. I read it properly now and measured the two claims it rests on instead of taking them on trust.

One thing is settled up front, and nothing below argues against it: reading the key from the login Keychain is the right source. Better than a plaintext setting or an environment variable, and it stays the source in whatever shape this lands.

The problem is the trigger, not the source. The title promises a fallback for an expired OAuth session, but no such condition exists anywhere in the code. `fetchAnthropicAPIKeyFromKeychain()` runs on every generation and the key is injected whenever the item is present. That is an unconditional switch of the authentication method.

I checked which one wins, with a control, using this app's own arguments (`-p - --output-format stream-json --verbose --model sonnet`):

- control, no key in the environment: answers promptly, exit 0
- with an unusable `ANTHROPIC_API_KEY`: no result in two runs of 60 s and 45 s, exit 124, and no fallback to OAuth

The CLI states the precedence itself on stderr: `ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login`.

That changes the risk materially. A stale, rotated or unrelated value under a name as generic as `anthropic-api-key` turns a working installation into one that hangs until the app's own timeout, which is the same silent failure this PR set out to fix, now reachable through the fix itself. Separately it moves a healthy subscription session onto metered billing without anyone being asked.

The shape that keeps the Keychain and matches your own title: run the CLI as it is today, and only when that attempt fails, retry once with the Keychain key injected. A healthy session is then never touched, a bad Keychain item cannot break anything, and the metered path is reached only when the free one is already broken.

One more thing, and it is the cheaper half of your own bug report. You describe eight meetings failing with nothing to go on, because stderr is empty while the real message sits in the stream-json on stdout that the app discards. Surfacing that is a smaller change than this one, it would have told you what was wrong on the first meeting rather than the eighth, and it helps with every other way the CLI can fail. Worth doing on its own merits, whatever happens to the API key path.

Four line comments below.

pasrom:
A second review pass went over both your change and my own review. It confirmed the main finding, corrected one thing I got wrong, and turned up something more useful than anything in my first pass.

## Correction to what I wrote

I said an unusable key makes generation hang until the app's own timeout. That is wrong, and the mistake was mine: my probes used 45 s and 60 s limits, which are shorter than the CLI's retry schedule, so I read "no output yet" as "hangs".

Measured properly, twice, with a limit long enough to see the end: the CLI retries ten times with exponential backoff (`max_retries: 10`, delays growing from about 0.6 s to 38 s, `error_status: 401`) and then exits 1 after 188.0 s and 187.6 s. A control without the key answers in 16.7 s with exit 0.

So in the app the outcome is not `ProtocolError.timeout` after ten minutes. It is `cliFailed(1, ...)` after roughly three, and the message the user gets is whatever stderr held. Which brings me to the part that matters more.

## The reason you were looking for is already parsed, and then thrown away

This is the real answer to your eight meetings, and it is smaller than this PR.

`generate()` computes `let text = try await Self.readStreamJSON(...)` before it checks the exit status. When the status is nonzero it throws `makeFailureError(exitCode:stderrText:)` and discards `text` entirely. But the CLI puts its failure reason into the stream-json on stdout as a synthetic assistant message, and `parseStreamJSONLine` returns it as ordinary text. Fed the captured stdout of a real failing run, the app's own parser yields exactly this, and nothing else:

```
Failed to authenticate. API Error: 401 API key is invalid.
```

Meanwhile stderr, the only thing the user is shown, carried this:

```
⚠ claude.ai connectors are disabled because ANTHROPIC_API_KEY or another auth source is set and takes precedence over your claude.ai login
```

So the app already holds the sentence you needed. Preferring `text` over an empty or unhelpful `stderrText` when the exit is nonzero is a change of a couple of lines, it would have told you what was wrong on the first meeting instead of the eighth, and it improves every other way the CLI can fail. I would take that on its own, whatever happens to the API key path.

## The app already owns the Keychain pattern you want

`KeychainHelper.swift` reads and writes secrets natively through `SecItemCopyMatching`, `AppSettings.openAIAPIKey` is backed by it, and Settings already exposes an API key through a `SecureField`. So there is a shape here that keeps the Keychain, which the maintainer prefers as the source, and drops three of my four objections at once: an item the app creates itself is trusted by default and cannot raise the ACL prompt, there is no subprocess and so no blocking call, and there is no `whoami` naming convention for a user to get wrong. It also makes the billing choice explicit instead of implicit, because the user typed the key in rather than having one discovered.

## Two more things in the diff

Because the lookup sits inside `generate()`, the existing test that drives `generate()` now runs `security find-generic-password` against the real login Keychain of whoever runs `swift test`, and on a machine where the item exists it hands that key to the fake CLI's environment. The suite should not be reading a developer's secrets.

And both `fetchAnthropicAPIKeyFromKeychain` tests still pass with the exit status guard inverted to `guard status != 0`, checked by running that mutant. Neither reaches the success path, which is the one line of the patch with no coverage. The `securityBin` seam you already built makes a positive test easy: a stub that prints a key and exits 0, and one that prints a key and exits 1.

## On the shape I proposed

I presented "retry once after failure with the key injected" as the fix. It is implementable, because the CLI does emit a machine readable auth discriminator on stdout, so a retry can be gated on that rather than on any nonzero exit. But it has costs I did not weigh: it would put those three minutes in front of every protocol until the user runs `claude login`, it couples us to undocumented stream-json fields that a future CLI version may drop, and `readStreamJSON` returns only text today so the error kind needs a structural change to reach the retry.

Treat it as one option rather than the option. My recommendation now is the explicit Settings owned key plus the discarded-reason fix, and no automatic fallback at all.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
pasrom: Correction to this comment: the sentence about running to `ProtocolError.timeout` is wrong. My probes were cut off at 45 s and 60 s, shorter than the CLI's retry schedule, so I read "no output yet" as a hang. Measured with a long enough limit, twice: ten retries with backoff, then exit 1 after 188.0 s and 187.6 s. In the app that is `cliFailed(1, ...)` after about three minutes, not a timeout after ten.

Everything else in this comment stands, and the precedence is confirmed by the CLI's own `apiKeySource: "ANTHROPIC_API_KEY"` and `terminal_reason: "api_error"` with no OAuth attempt. Full correction and what came out of the second pass is in the comment on the conversation tab.

GB-D:
@pasrom could you approve the latest run on this branch? Pushed an empty commit to retrigger CI and check whether the earlier `test (homebrew)` failure (AppPathsTests, trailing slash mismatch) was a one-off flake, but it's sitting at "action required" pending approval.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -130,6 +140,29 @@
...
+        }
+
+        /// Decodes `stderrData`, logs the raw exit/stderr/text, and builds the
+        /// error `generate()` throws on a nonzero exit — pulled out of
+        /// `generate()` itself to keep that function's body short.
+        static func handleFailure(exitCode: Int32, stderrData: Data, text: String) -> ProtocolError {
+            let stderrText = String(data: stderrData, encoding: .utf8)?
+                .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+            logger.error(
+                "claude_cli_failed exit=\(exitCode, privacy: .public) stderr=\(stderrText, privacy: .public) text=\(text, privacy: .public)",
```
pasrom: This has to be `privacy: .private` for `text`, and it is the one thing blocking approval.

`text` is whatever the model streamed back. On an auth failure it is the error sentence, which is harmless. But on a failure that happens after generation has already started, a dropped connection, a rate limit mid-stream, the CLI being killed, `text` holds a partial meeting protocol. This line writes it to the unified log unredacted, where it stays and where anyone reading the log sees it.

The repo already draws this line in exactly this place. `LiveTranscriptionController` logs partial and final transcript text as `.private`, `PipelineQueue+Stages` logs the meeting title as `.private`, and the function right above this one deliberately logs `prompt_bytes` as a count rather than the prompt. The exit code and stderr can stay `.public`; it is only `text` that carries user content.

Worth knowing, and the same root cause, though I would not hold the PR for it: `selectFailureMessage` cannot tell an error message from partial protocol output, so that same mid-stream failure surfaces a truncated protocol to the user as the failure reason. Fine as a trade for what you gained, but if you ever want to tighten it, the CLI does mark its own error messages in the stream-json rather than leaving you to guess from the text.

pasrom (changes requested):
This is a much better change than the one I asked for, and both of my main points are genuinely closed rather than nominally. I read the code for each rather than taking the commit messages for it.

**The unconditional switch is gone.** `fetchAnthropicAPIKeyFromKeychain()` is deleted outright, the key arrives as a constructor value, and `PipelineController` maps an empty setting to `nil`, so a user who never touches the field gets exactly today's behaviour. The Settings copy naming the billing consequence out loud is the right call and more than I asked for.

**The wiring test is the real thing.** `testGeneratePassesAnthropicAPIKeyToSubprocessEnvironment` drives `generate()` with a fake binary that echoes the variable back, so it asserts what the subprocess actually received rather than what the pure function returned. Deleting the injection now turns it red. The negative case using `${ANTHROPIC_API_KEY+x}` to separate unset from empty is the detail that makes it worth having.

The blocking Keychain subprocess went with the auto-discovery, and the pointer into your own environment is gone, replaced by a deliberately app-specific account name with the reasoning written down. `testClaudeAPIKeyIsIndependentOfOpenAIAPIKey` pins that it cannot collide with the existing item.

`selectFailureMessage` plus its end to end test is the part I would have merged on its own. That is the actual answer to your eight meetings.

I ran the full suite locally on this head: 3086 tests, no failures.

One thing has to change before I approve: a log line publishes generated meeting content, through two call sites rather than one. Details in the line comment, including a fix I would prefer over annotating both.

Two smaller notes, neither of them a condition:

The new `SecureField` has no ViewInspector wiring test. The house rule asks for one per control, but the existing OpenAI key field next to it has none either, so you are consistent with the neighbour and I am not going to make you fix somebody else's gap.

`claudeAPIKeyAccount` is declared outside the `#if !APPSTORE` that wraps `claudeAPIKey`, so in the sandboxed build it is a stored property and an init parameter nothing reads. Harmless, and the App Store build is green.

Finally: the branch is behind `main` again, and `main` requires an up to date branch. I can rebase and merge it from here once the log line is fixed, so do not feel you have to chase it.

If you would rather not spend another round on it, say so and I will push the change to your branch myself.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
pasrom: A second opinion went over this comment and found real problems with it. Correcting myself before you act on it.

**"One word" is wrong, and so was separating the two points.** Making `text` `.private` here does not close the leak. `selectFailureMessage` puts the same value into `ProtocolError.cliFailed`, whose description interpolates it (`ProtocolGenerator.swift:280`), and `PipelineQueue+Stages.swift:815` logs `error.localizedDescription` with `privacy: .public`. So the content escapes through a second site regardless of what this line says. The `selectFailureMessage` point is part of the blocker, not the footnote I made it.

**There is a cleaner fix than annotating two log sites.** Take the failure reason from the final `result` event instead of from accumulated assistant text. In a captured failing run that event reads:

```json
{"type":"result","is_error":true,"terminal_reason":"api_error",
 "result":"Failed to authenticate. API Error: 401 API key is invalid."}
```

Note `subtype` is `"success"` even there, so `is_error` is the field to key on. `parseStreamJSONLine` currently drops `result` events entirely. Reading the message from there means the failure reason is never built from protocol content in the first place, both log sites become harmless, and your end to end test keeps passing with the fake binary emitting a result line instead of a text block. That is a better change than the one I asked for.

**Three things I got wrong in the text above, all in your favour:**

"a partial meeting protocol" is inaccurate. The CLI is launched without `--include-partial-messages`, and a real run emits zero `content_block_delta` events, only whole assistant messages. So the payload is not a truncated protocol, it is a complete one followed by a nonzero exit. That makes the leak worse than I described, not better.

"the CLI being killed" does not belong in the list. The app's own timeout calls `process.terminate()` and throws `.timeout` before this path is reached, so only an external kill lands here.

"surfaces a truncated protocol to the user as the failure reason" is simply wrong. The user-facing string is the fixed `"Protocol generation failed — transcript saved"`. The content goes to the log and, through `PersistentDiagnosticLog` and `DiagnosticExporter`, into a file a user may attach to a bug report. Which is the reason to care, but it is not what I wrote.

For the record: `privacy: .public` on `exit` and `stderr` was already there before your change and you kept the existing pattern, and the bug that motivated all of this is real.

pasrom: This is the one thing blocking approval, and it is bigger than a privacy annotation.

`text` is whatever the model streamed back. On an auth failure that is the error sentence, which is harmless. But on any nonzero exit that happens after the assistant message has been emitted, `text` holds the generated meeting protocol. The CLI runs without `--include-partial-messages`, and a real run emits zero `content_block_delta` events, only whole assistant messages, so this is a complete protocol rather than a fragment.

Two places then publish it:

- this line, `text=\(text, privacy: .public)`
- `selectFailureMessage` puts the same value into `ProtocolError.cliFailed`, whose description interpolates it (`ProtocolGenerator.swift:280`), and `PipelineQueue+Stages.swift:815` logs `error.localizedDescription` with `privacy: .public`

So making this line `.private` on its own does not close it. And it reaches further than the unified log: `PersistentDiagnosticLog` streams error-level entries into a file, and `DiagnosticExporter` bundles that into a `.log` a user attaches to a bug report.

The repo already draws this line in exactly this place. `LiveTranscriptionController` logs transcript text as `.private`, `PipelineQueue+Stages` logs the meeting title as `.private`, and the function right above this one deliberately logs `prompt_bytes` as a count rather than the prompt. For the record, `.public` on `exit` and `stderr` was there before your change and you kept the existing pattern. It is only `text` that is new and carries content.

**The fix I would prefer over annotating two call sites:** take the failure reason from the final `result` event instead of from accumulated assistant text. A captured failing run ends with this:

```json
{"type":"result","is_error":true,"terminal_reason":"api_error",
 "result":"Failed to authenticate. API Error: 401 API key is invalid."}
```

`subtype` reads `"success"` even there, so `is_error` is the field to key on. `parseStreamJSONLine` drops `result` events entirely today. Sourcing the message from there means the failure reason is never built out of protocol content in the first place, both log sites become harmless with no privacy annotation needed, and `testGenerateOnFailureThrowsTextOverEmptyStderr` still holds with the fake binary emitting a result line instead of a text block.

Marking both sites `.private` also closes it, if you would rather keep the change small. Either way it needs both.

pasrom (approved):
The log leak is closed at both sites, so this is good to go.

I folded the branch into two commits and rebased it onto current `main`, since `main` requires an up to date branch and the review fixes read better as part of the change they correct than as commits on top. Your authorship is preserved on both. The fold was built from the trees rather than squashed, so the result is bit for bit the same tree as the five commit version, and `git diff` between the two is empty. Gates on the folded head: lint clean across 552 files, and the full suite at 3134 tests with no failures.

One consequence worth recording for whoever reads this later, because it is a real trade and not an oversight. `protocol_generation_failed` in `PipelineQueue+Stages` is the shared failure handler for every protocol provider, so marking it `.private` also redacts reasons that were readable before and carry no meeting content: an OpenAI provider error, `cliNotFound`, `timeout`, `emptyProtocol`. Sourcing the failure message from the stream-json `result` event instead would let both log sites go back to `.public` and keep the diagnostics useful, since that event carries the CLI's own error sentence rather than generated content. Not a condition on this PR, and the leak had to be closed either way.

Thanks for the patience through three rounds, two of which were me correcting my own review rather than your code.

Issue #446, opened 2026-06-22
[Bug] Mystery `Health Check Failed: <private>` Error IN logig


MattyD-Pave:
### App version

0.6.0-rc4 (ce997f5) · Homebrew

### Build variant

Homebrew (beta / RC)

### macOS version

Tahoe 26.5

### What happened?

My transcriber app stopped detecting zoom meetings, and has been signaling and error for a while now. **EDIT:** I can still manually trigger an app recording and the audio comes in correctly, so it seems like this error is around detection. 

<img width="266" height="257" alt="Image" src="https://github.com/user-attachments/assets/cedfcfee-880e-4187-af7a-c8917075a4f7" />

I double checked...
- Microphone ✅ 
- Screen & Audio Sharing ✅ 
- Accessibility ✅ 
- Full Disk Access ✅  (ran into issues with this, possibly just b/c of where i placed the protocols folder.
- I tried updating to the release candidate, still nothing.

After checking all of this, i just see the following logs which don't have much info
```
2026-06-22 15:30:27.522407-0400  localhost MeetingTranscriber[6857]: [com.meetingtranscriber:PermissionHealthCheck] Permission health check failed: <private>
2026-06-22 15:30:36.714905-0400  localhost MeetingTranscriber[6857]: [com.meetingtranscriber:PermissionHealthCheck] Permission health check failed: <private>
```

### Steps to reproduce

On startup

### Diagnostic log

[diagnostics-2026-06-18.log](https://github.com/user-attachments/files/29220535/diagnostics-2026-06-18.log)
[diagnostics-2026-06-22.log](https://github.com/user-attachments/files/29220536/diagnostics-2026-06-22.log)

pasrom:
Thanks for the detailed report, and especially for the screenshot and the two diagnostic logs. That `<private>` in the log was the real blocker: os.Logger redacts non-numeric values by default, so the message hid the one thing you needed, which permission it thought was broken. Fixing that was step one.

Digging into your logs plus the code, we found three separate problems in the permission health check, and they line up with exactly what you saw:

1. **The unreadable log.** The failure cause is now logged in the clear, so it names the actual permission (e.g. `screenRecording=broken` or `microphone=broken`) instead of `<private>`.

[...]

PR #695, opened 2026-09-08
fix(app): stop reporting a working Accessibility grant as broken

egerkuzma:
## The bug

The permission check tells users "Accessibility looks enabled but isn't working —
toggle it off and on in System Settings" while nothing is wrong with the grant.
`/tmp/mt-permission.log`, all in one process, with no user action in between:

```
17:37:50  trusted=true probe=true  -> healthy
17:39:29  trusted=true probe=false -> broken     (notification sent)
17:40:39  trusted=true probe=true  -> healthy
17:47:13  trusted=true probe=false -> broken     (notification sent)
```

A revoked grant does not repair itself in seventy seconds.

## Why it happens

`probeAccessibility()` asks the system-wide AX element for the **frontmost app's**
focused application. That is a cross-process call, so its failure modes belong to
whichever app happens to have focus — but the result was a `Bool` where every
non-success collapsed into "our permission is broken". The check was measuring
how responsive some other application is and announcing the answer as if it were
about us.

## The fix

`AXIsProcessTrusted()` is already the system's own answer about our grant, and
`.broken` exists for exactly one situation: TCC says yes while the API says no.
So the probe only has to catch that contradiction. `kAXErrorAPIDisabled` is
documented as "the accessibility API is disabled" and is the one error that
qualifies; everything else becomes `AccessibilityProbe.inconclusive` and reports
healthy.

`kAXErrorNotImplemented` is deliberately kept out of that bucket even though it
reads like a sibling — Apple documents it as returned when *the process being
asked* does not support the accessibility API, so classifying it as a permission
failure would re-create the same false alarm for every frontmost app with
incomplete AX support. `testClassifyKeepsNotImplementedOutOfTheBrokenSignal` pins that
distinction against `classify` itself, and fails if the two are regrouped.

The raw `AXError` now reaches the diagnostic log. The `Bool` had thrown away the
one fact that explained the failure, which is why this could only be diagnosed by
noticing the timing rather than by reading the code.

**Rejected:** hysteresis over consecutive failures. With the errors classified,
the only remaining route to `.broken` is a non-transient condition, so a counter
would just delay a correct verdict while keeping the incorrect one reachable.

## Notes

The accessibility cases move into their own test file: the four added here push
`PermissionHealthCheckTests` past the 400-line `type_body_length` limit, and the
lint config allows one XCTestCase per file.

## Verification

- `swift test --parallel --skip ModelPreloadTests` — 3077 tests, 0 failures
- SwiftFormat clean, SwiftLint `--strict` exit 0
- `swift build -c release` clean under warnings-as-errors

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -220,7 +220,7 @@ Use the `/git-workflow` skill. Commit proactively after every logical unit of wo
 - `UpdateChecker` checks GitHub releases for newer versions, shows badge on menu bar icon.
 
 **Permission health check:**
-- `PermissionHealthCheck` verifies each TCC permission by combining the system verdict with a live probe. Each resolves to `PermissionStatus` (`.healthy | .denied | .broken | .notDetermined`). `.broken` means TCC says allowed but the probe disagrees — fix is to toggle the permission off and on in System Settings.
+- `PermissionHealthCheck` verifies each TCC permission by combining the system verdict with a live probe. Each resolves to `PermissionStatus` (`.healthy | .denied | .broken | .notDetermined`). `.broken` means TCC says allowed but the probe disagrees — fix is to toggle the permission off and on in System Settings. The Accessibility probe is a cross-process call, so only `kAXErrorAPIDisabled` counts as `.broken`; `.success` and `.noValue` are `AccessibilityProbe.responded`, and every other AXError is `.inconclusive` and reports healthy. Treating those as broken produced a notification telling users to toggle a permission that was working. `kAXErrorNotImplemented` stays inconclusive on purpose: Apple documents it as the asked process lacking AX support. The raw code is appended to `/tmp/mt-permission.log`, which `debugLog` truncates per process.
 - `WatchLoop` runs the check on startup; `AppState` re-runs on app activation.
 - When unhealthy: `MenuBarIcon` composites a red "!" badge over the current icon (non-template, stays red in dark mode). `BadgeKind.compute()` returns `.error` when idle with a problem. A deduped notification is posted via `NotificationManager`.
 
--- a/app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
+++ b/app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
@@ -427,22 +427,83 @@ enum PermissionHealthCheck {
 
     // MARK: - Accessibility (pure, testable)
 
+    /// What the AX probe actually learned. Three cases and not a `Bool`, because
+    /// the probe reaches ACROSS PROCESSES and its failures are therefore not all
+    /// about us: only one of them is evidence about our own grant.
+    enum AccessibilityProbe: Equatable {
+        /// The API answered. Says our grant works.
+        case responded
+        /// The AX API is disabled for us (`kAXErrorAPIDisabled`). The only
+        /// outcome that is evidence our own grant is not working.
+        case apiDisabled
+        /// The call failed for a reason that has not been shown to be about our
+        /// permission. Carries the raw code for the log, so the next occurrence
+        /// can be attributed instead of guessed.
+        case inconclusive(AXError)
+
+        /// Compact, PII-free token for the diagnostic log. The raw `AXError`
+        /// value is spelled out because the name alone ("inconclusive") is what
+        /// the old `Bool` already told us; the number is the part that says
+        /// which failure it was.
+        var logToken: String {
+            switch self {
+            case .responded: "responded"
+            case .apiDisabled: "apiDisabled"
+            case let .inconclusive(err): "inconclusive(AXError \(err.rawValue))"
+            }
+        }
+    }
+
     /// Pure decision function for Accessibility permission state.
     ///
     /// - `trusted`: whether `AXIsProcessTrusted()` returns true.
-    /// - `probeSucceeds`: whether a concrete AX API call (e.g. reading the focused app of
-    ///   `AXUIElementCreateSystemWide`) returns `.success`.
+    /// - `probe`: what a concrete AX API call learned.
+    ///
+    /// **An inconclusive probe reports `.healthy`, not `.broken`.** This used to
+    /// be a `Bool` where every non-success collapsed to "broken", and the
+    /// reported consequence was a false alarm telling users to toggle a
+    /// permission that was fine: one process, no user action, and the verdict
+    /// flipping healthy -> broken -> healthy inside 70 seconds. A revoked grant
+    /// does not repair itself in a minute.
+    ///
+    /// `.broken` exists for exactly one situation, TCC saying yes while the API
+    /// says no, and `AXIsProcessTrusted()` is already the system's own answer
+    /// about our grant. So the probe only has to catch that contradiction, which
+    /// is `.apiDisabled`.
     static func checkAccessibility(
         trusted: Bool,
-        probeSucceeds: Bool,
+        probe: AccessibilityProbe,
     ) -> PermissionStatus {
         if !trusted { return .denied }
-        return probeSucceeds ? .healthy : .broken
+        switch probe {
+        case .responded, .inconclusive: return .healthy
+        case .apiDisabled: return .broken
+        }
+    }
+
+    /// What one `AXError` says about *our* Accessibility access. Kept apart from
+    /// the call that produces it because this mapping is the decision the fix
+    /// turns on, and welding it to an I/O call would leave it untestable.
+    ///
+    /// `.notImplemented` takes the default branch deliberately: Apple documents
+    /// it as "can be returned if a PROCESS does not support the accessibility
+    /// API" — the process being asked, not us.
+    static func classify(_ err: AXError) -> AccessibilityProbe {
+        switch err {
+        // .noValue: nothing focused right now, but the API answered.
+        case .success, .noValue: .responded
+        case .apiDisabled: .apiDisabled
+        default: .inconclusive(err)
+        }
     }
 
     /// Probes the Accessibility API with a lightweight system-wide call.
-    /// Returns true if the call succeeds (and we either get a valid attribute or `noValue`).
-    static func probeAccessibility() -> Bool {
+    ///
+    /// `kAXFocusedApplicationAttribute` on the system-wide element is a question
+    /// put to whichever app currently has focus, so its failure modes belong to
+    /// that app as much as to us — which is what `AccessibilityProbe` separates.
+    /// Everything except the call itself lives in `classify`.
+    static func probeAccessibility() -> AccessibilityProbe {
         let systemWide = AXUIElementCreateSystemWide()
         var value: CFTypeRef?
         let err = AXUIElementCopyAttributeValue(
@@ -450,9 +511,7 @@ enum PermissionHealthCheck {
             kAXFocusedApplicationAttribute as CFString,
             &value,
         )
-        // .success: got the focused app. .noValue: no focused app right now, but the API worked.
-        // Any other error (e.g. .cannotComplete, .apiDisabled) indicates AX isn't actually granted.
-        return err == .success || err == .noValue
+        return classify(err)
     }
 
     static func checkAccessibilityLive() -> PermissionStatus {
@@ -462,8 +521,11 @@ enum PermissionHealthCheck {
             return .denied
         }
         let probe = probeAccessibility()
-        let r = checkAccessibility(trusted: trusted, probeSucceeds: probe)
-        debugLog("checkAccessibilityLive: trusted=true probe=\(probe) → \(r)")
+        let r = checkAccessibility(trusted: trusted, probe: probe)
+        // The raw AXError is logged, not just the verdict. Without it the false
+        // alarm above could only be diagnosed by noticing the timing, because a
+        // `Bool` had already thrown away the one fact that explains it.
+        debugLog("checkAccessibilityLive: trusted=true probe=\(probe.logToken) → \(r)")
         return r
     }
 
--- a/app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
+++ b/app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
@@ -1,0 +1,124 @@
+import ApplicationServices
+@testable import MeetingTranscriber
+import XCTest
+
+/// Accessibility permission verdicts, split out of `PermissionHealthCheckTests`
+/// so neither class outgrows the 400-line body limit. The probe's three-way
+/// outcome carries most of the reasoning in this area, so it earns its own file.
+final class PermissionHealthCheckAccessibilityTests: XCTestCase {
+    func testAccessibilityHealthy() {
+        let result = PermissionHealthCheck.checkAccessibility(trusted: true, probe: .responded)
+        XCTAssertEqual(result, .healthy)
+    }
+
+    func testAccessibilityDenied() {
+        let result = PermissionHealthCheck.checkAccessibility(trusted: false, probe: .apiDisabled)
+        XCTAssertEqual(result, .denied)
+    }
+
+    func testAccessibilityDeniedEvenIfProbeSucceeds() {
+        // Defensive: if the system says no, we never report healthy.
+        let result = PermissionHealthCheck.checkAccessibility(trusted: false, probe: .responded)
+        XCTAssertEqual(result, .denied)
+    }
+
+    func testAccessibilityBroken() {
+        let result = PermissionHealthCheck.checkAccessibility(trusted: true, probe: .apiDisabled)
+        XCTAssertEqual(result, .broken)
+    }
+
+    /// The regression this split exists for. `.cannotComplete` is what a
+    /// frontmost app that does not answer AX produces, and it used to collapse
+    /// into the same "broken" as a genuinely closed API — which is how a healthy
+    /// grant got a notification telling the user to toggle it off and on.
+    func testAccessibilityInconclusiveProbeIsNotReportedAsBroken() {
+        let result = PermissionHealthCheck.checkAccessibility(
+            trusted: true, probe: .inconclusive(.cannotComplete),
+        )
+        XCTAssertEqual(
+            result, .healthy,
+            "An unresponsive frontmost app says nothing about our own grant",
+        )
+    }
+
+    /// Every non-`apiDisabled` error takes the inconclusive path, so the fix
+    /// does not depend on guessing which code a busy app happens to return.
+    func testAccessibilityOtherAXErrorsAreAlsoInconclusive() {
+        for error in [AXError.failure, .invalidUIElement, .illegalArgument, .attributeUnsupported] {
+            XCTAssertEqual(
+                PermissionHealthCheck.checkAccessibility(trusted: true, probe: .inconclusive(error)),
+                .healthy,
+                "AXError \(error.rawValue) describes the element asked, not our permission",
+            )
+        }
+    }
+
+    // MARK: - AXError classification
+
+    /// `.notImplemented` sits next to `.apiDisabled` in the AX header and means
+    /// something entirely different: the *asked* process has incomplete AX
+    /// support. Classifying it as a permission failure would re-create the false
+    /// alarm for every such app, so the classifier must keep them apart. This
+    /// asserts the mapping itself, not its consequence two calls downstream.
+    func testClassifyMapsOnlyApiDisabledToTheBrokenSignal() {
+        XCTAssertEqual(PermissionHealthCheck.classify(.apiDisabled), .apiDisabled)
+    }
+
+    func testClassifyKeepsNotImplementedOutOfTheBrokenSignal() {
+        XCTAssertEqual(
+            PermissionHealthCheck.classify(.notImplemented),
+            .inconclusive(.notImplemented),
+            "notImplemented describes the process being asked, not our grant",
+        )
+    }
+
+    func testClassifyTreatsAnsweredCallsAsResponded() {
+        // `.noValue` is not a failure: nothing has focus right now, but the API
+        // answered, which is the only thing the probe asked.
+        XCTAssertEqual(PermissionHealthCheck.classify(.success), .responded)
+        XCTAssertEqual(PermissionHealthCheck.classify(.noValue), .responded)
+    }
+
+    func testClassifyTreatsCannotCompleteAsInconclusive() {
+        XCTAssertEqual(
+            PermissionHealthCheck.classify(.cannotComplete),
+            .inconclusive(.cannotComplete),
+        )
+    }
+
+    /// Every remaining error describes the element that was asked. Enumerated
+    /// rather than sampled, so a future regrouping of any one of them has to
+    /// come here and say so.
+    func testClassifyTreatsEveryOtherErrorAsInconclusive() {
+        let others: [AXError] = [
+            .failure, .illegalArgument, .invalidUIElement, .invalidUIElementObserver,
+            .cannotComplete, .attributeUnsupported, .actionUnsupported,
+            .notificationUnsupported, .notImplemented, .notificationAlreadyRegistered,
+            .notificationNotRegistered, .parameterizedAttributeUnsupported,
+            .notEnoughPrecision,
+        ]
+        for error in others {
+            XCTAssertEqual(
+                PermissionHealthCheck.classify(error), .inconclusive(error),
+                "AXError \(error.rawValue) is about the element asked, not our permission",
+            )
+        }
+    }
+
+    /// The end-to-end shape of the fix: a classified error carried through to a
+    /// permission verdict.
+    func testInconclusiveClassificationReachesAHealthyVerdict() {
+        let probe = PermissionHealthCheck.classify(.cannotComplete)
+        XCTAssertEqual(PermissionHealthCheck.checkAccessibility(trusted: true, probe: probe), .healthy)
+    }
+
+    /// The verdict alone could not explain the false alarm; the raw code can.
+    func testAccessibilityProbeLogTokenCarriesTheRawErrorCode() {
+        XCTAssertEqual(
+            PermissionHealthCheck.AccessibilityProbe.inconclusive(.cannotComplete).logToken,
+            "inconclusive(AXError \(AXError.cannotComplete.rawValue))",
+        )
+        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.responded.logToken, "responded")
+        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.apiDisabled.logToken, "apiDisabled")
+    }
+}
--- a/app/MeetingTranscriber/Tests/PermissionHealthCheckTests.swift
+++ b/app/MeetingTranscriber/Tests/PermissionHealthCheckTests.swift
@@ -117,29 +117,6 @@ final class PermissionHealthCheckTests: XCTestCase {
         )
     }
 
-    // MARK: - Accessibility
-
-    func testAccessibilityHealthy() {
-        let result = PermissionHealthCheck.checkAccessibility(trusted: true, probeSucceeds: true)
-        XCTAssertEqual(result, .healthy)
-    }
-
-    func testAccessibilityDenied() {
-        let result = PermissionHealthCheck.checkAccessibility(trusted: false, probeSucceeds: false)
-        XCTAssertEqual(result, .denied)
-    }
-
-    func testAccessibilityDeniedEvenIfProbeSucceeds() {
-        // Defensive: if the system says no, we never report healthy.
-        let result = PermissionHealthCheck.checkAccessibility(trusted: false, probeSucceeds: true)
-        XCTAssertEqual(result, .denied)
-    }
-
-    func testAccessibilityBroken() {
-        let result = PermissionHealthCheck.checkAccessibility(trusted: true, probeSucceeds: false)
-        XCTAssertEqual(result, .broken)
-    }
-
     // MARK: - Overall Health
 
     func testOverallHealthy() {


app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
```diff
@@ -427,32 +427,91 @@ enum PermissionHealthCheck {
...
+    /// about us: only one of them is evidence about our own grant.
+    enum AccessibilityProbe: Equatable {
+        /// The API answered. Says our grant works.
+        case responded
+        /// The AX API is disabled for us (`kAXErrorAPIDisabled`). The only
+        /// outcome that is evidence our own grant is not working.
+        case apiDisabled
+        /// The call failed for a reason that is about the element being asked,
+        /// not about our permission — most often the frontmost app being slow
+        /// or not answering AX at all. Carries the raw code for the log.
```
pasrom: "most often the frontmost app being slow or not answering AX at all" is an inference, and I think it is the wrong one.

What I measured on a machine whose grant was fine: `AXIsProcessTrusted()` true, and this exact system-wide query returning `cannotComplete` on every attempt, immediately rather than after a delay, while in the same process app-targeted calls against Finder, the frontmost app and Dock all returned `.success` with real titles and window counts. A slow or unresponsive frontmost app does not explain that shape.

Suggest describing what is known and leaving the mechanism open, something like "the call failed for a reason that is not about our permission; the system-wide query has failure modes of its own". The verdict logic does not depend on the explanation, so a weaker sentence costs nothing.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
```diff
@@ -427,32 +427,91 @@ enum PermissionHealthCheck {
...
+    /// flipping healthy -> broken -> healthy inside 70 seconds (17:37 healthy,
+    /// 17:39 broken + notification, 17:40 healthy again). A revoked grant does
+    /// not repair itself in a minute; an unresponsive frontmost app does.
+    ///
+    /// `.broken` exists for exactly one situation — TCC says yes while the API
+    /// says no — and `AXIsProcessTrusted()` is already the system's own answer
+    /// about our grant. So the probe only has to catch the contradiction, which
+    /// is `.apiDisabled`. Every other AX error describes the element that was
+    /// asked. Reporting those as broken makes the check measure how responsive
+    /// some other application is and announce the answer as if it were about us.
```
pasrom: This paragraph repeats what the CLAUDE.md bullet in this same PR already says, and it pins two things that are weaker than the wording suggests: the wall-clock timestamps are one machine's session and will read as noise later, and "measured consequence" covers the flap (which was measured) but sits next to a cause that was not.

The trim below also buys the five lines the file needs, see the note on `probeAccessibility` below. With it the file is 599 lines and SwiftLint `--strict` passes; the extraction suggestion alone would be 602 and fail.

```suggestion
    /// **An inconclusive probe reports `.healthy`, not `.broken`.** This used to
    /// be a `Bool` where every non-success collapsed to "broken", and the
    /// reported consequence was a false alarm telling users to toggle a
    /// permission that was fine: one process, no user action, and the verdict
    /// flipping healthy -> broken -> healthy inside 70 seconds. A revoked grant
    /// does not repair itself in a minute.
    ///
    /// `.broken` exists for exactly one situation, TCC saying yes while the API
    /// says no, and `AXIsProcessTrusted()` is already the system's own answer
    /// about our grant. So the probe only has to catch that contradiction, which
    /// is `.apiDisabled`.
```

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
```diff
@@ -427,32 +427,91 @@ enum PermissionHealthCheck {
...
+        }
     }
 
     /// Probes the Accessibility API with a lightweight system-wide call.
-    /// Returns true if the call succeeds (and we either get a valid attribute or `noValue`).
-    static func probeAccessibility() -> Bool {
+    ///
+    /// `kAXFocusedApplicationAttribute` on the system-wide element is a question
+    /// put to whichever app currently has focus, so its failure modes belong to
+    /// that app as much as to us — which is what `AccessibilityProbe` separates.
```
pasrom: Follow-up rather than a change request for this PR.

Since `.broken` now hangs entirely on one error code from this one query, it is worth asking whether this query can produce it at all. `ParticipantReader` does not use the system-wide element; it targets an application directly, which is where an untrusted client would actually surface `kAXErrorAPIDisabled`. An app-targeted probe would make `.broken` observable instead of theoretical.

Also relevant: commit 85407512 (issue #446) hit the same false-broken problem in this file for Screen Recording and answered it by dropping the probe from the verdict altogether. Keeping one path here may be the better call, but the PR reads as if the precedent does not exist.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
```diff
@@ -427,32 +427,91 @@ enum PermissionHealthCheck {
...
+        // disabled", which is about our access. `.notImplemented` reads like a
+        // sibling and is not one: its own documentation says it "can be returned
+        // if a PROCESS does not support the accessibility API" — the process
+        // being asked. Putting it here would re-create the false alarm for every
+        // frontmost app with incomplete AX support.
+        case .apiDisabled: return .apiDisabled
+
+        default: return .inconclusive(err)
+        }
     }
```
pasrom: This `switch` is the fix, and nothing exercises it. `probeAccessibility` is referenced only at its definition and its one call site in `checkAccessibilityLive`; every new test builds an `AccessibilityProbe` by hand and passes it to `checkAccessibility(trusted:probe:)`, which never inspects the associated value.

Measured rather than argued: I removed the `.apiDisabled` arm below, so the only remaining route to `.broken` was gone, and `swift test --filter PermissionHealthCheck` still reported 53 of 53 passing. Regrouping `.notImplemented` into that arm is invisible in the same way. That is also why the PR description's "There is a test pinning that distinction" does not hold.

Extracting the mapping puts it on the layer this repo asks for ("pure logic first"):

```suggestion
        return classify(err)
    }

    /// Pure `AXError` -> probe classification, split out of
    /// `probeAccessibility()` so a test can reach it: `checkAccessibility`
    /// never reads the associated value, so nothing else pins this mapping.
    static func classify(_ err: AXError) -> AccessibilityProbe {
        switch err {
        // .success: got the focused app. .noValue: nothing focused right now,
        // but the API answered, which is what we asked.
        case .success, .noValue: .responded

        // ONLY `.apiDisabled`. Apple documents it as "the accessibility API is
        // disabled", which is about our access. `.notImplemented` reads like a
        // sibling and is not one: its own documentation says it "can be returned
        // if a PROCESS does not support the accessibility API" — the process
        // being asked. Putting it here would re-create the false alarm for every
        // frontmost app with incomplete AX support.
        case .apiDisabled: .apiDisabled

        default: .inconclusive(err)
        }
    }
```

Heads up: applied alone this makes the file 602 lines and SwiftLint `--strict` fails with `File Length Violation`. It needs the trim suggested on the doc comment above, which lands at 599.

With both applied plus these four tests, I measured 12 of 12 green, and `case .apiDisabled, .notImplemented: .apiDisabled` then fails `testClassifyKeepsNotImplementedOutOfTheBrokenSignal`:

```swift
    func testClassifyMapsOnlyApiDisabledToTheBrokenSignal() {
        XCTAssertEqual(PermissionHealthCheck.classify(.apiDisabled), .apiDisabled)
    }

    func testClassifyKeepsNotImplementedOutOfTheBrokenSignal() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.notImplemented),
            .inconclusive(.notImplemented),
            "notImplemented describes the process being asked, not our grant",
        )
    }

    func testClassifyTreatsAnsweredCallsAsResponded() {
        XCTAssertEqual(PermissionHealthCheck.classify(.success), .responded)
        XCTAssertEqual(PermissionHealthCheck.classify(.noValue), .responded)
    }

    func testClassifyTreatsCannotCompleteAsInconclusive() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.cannotComplete),
            .inconclusive(.cannotComplete),
        )
    }
```

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
```diff
@@ -0,0 +1,80 @@
...
+        XCTAssertEqual(
+            result, .healthy,
+            "An unresponsive frontmost app says nothing about our own grant",
+        )
+    }
+
+    /// Every non-`apiDisabled` error takes the inconclusive path, so the fix
+    /// does not depend on guessing which code a busy app happens to return.
+    func testAccessibilityOtherAXErrorsAreAlsoInconclusive() {
+        for error in [AXError.failure, .invalidUIElement, .noValue, .attributeUnsupported] {
```
pasrom: `.noValue` cannot reach this assertion: `probeAccessibility` maps it to `.responded`, so `.inconclusive(.noValue)` is a state production never builds. The docstring above ("Every non-`apiDisabled` error takes the inconclusive path") therefore describes the `default:` arm, which this test does not execute.

Since `checkAccessibility` ignores the payload, the loop asserts one code path four times, already covered by the previous test. It becomes meaningful once it runs against `classify`, per the other comment. Minimal fix meanwhile:

```suggestion
        for error in [AXError.failure, .invalidUIElement, .illegalArgument, .attributeUnsupported] {
```

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
```diff
@@ -0,0 +1,80 @@
...
+                "AXError \(error.rawValue) describes the element asked, not our permission",
+            )
+        }
+    }
+
+    /// `.notImplemented` sits next to `.apiDisabled` in the AX header and means
+    /// something entirely different: the *asked* process has incomplete AX
+    /// support. Classifying it as a permission failure would re-create the false
+    /// alarm for every such app, so the probe must keep them apart.
+    func testAccessibilityProbeTreatsNotImplementedAsInconclusive() {
```
pasrom: The name says "Probe treats", but this calls `checkAccessibility`, not the probe. Someone grepping the test class for probe coverage will find this and conclude the classification is pinned, which is exactly the gap flagged above.

The second assertion here is also a verbatim repeat of `testAccessibilityBroken`. Either point this at `classify` (then the name becomes true) or rename it to what it currently checks.

CLAUDE.md
```diff
@@ -220,7 +220,7 @@ Use the `/git-workflow` skill. Commit proactively after every logical unit of wo
...
... combining the system verdict with a live probe. Each resolves to `PermissionStatus` (`.healthy | .denied | .broken | .notDetermined`). `.broken` means TCC says allowed but the probe disagrees — fix is to toggle the permission off and on in System Settings. **A probe that fails must first be shown to be about us.** The Accessibility probe asks the system-wide AX element for the *frontmost app's* focused application, so most of its errors describe that app, not our grant; treating every non-success as `.broken` produced a notification telling users to toggle a working permission, with the verdict flipping healthy → broken → healthy inside 70 seconds in one process with no user action. Only `kAXErrorAPIDisabled` is evidence about our access (`AXIsProcessTrusted()` is already the system's own answer, and `.broken` exists only for the contradiction between the two); everything else is `AccessibilityProbe.inconclusive` and reports healthy. `kAXErrorNotImplemented` is deliberately NOT grouped with it — Apple documents it as the asked *process* lacking AX support. The raw `AXError` goes into the diagnostic log, because the old `Bool` threw away the one fact that explained the false alarm.
```
pasrom: This bullet goes from 316 to 1265 characters and largely restates the doc comments added to `PermissionHealthCheck.swift` in the same PR. Long bullets are this file's habit, so length alone is not the objection; the duplication is, because the two copies will drift.

Two accuracy points regardless of length:

- "everything else is `AccessibilityProbe.inconclusive`" is not right: `.success` and `.noValue` become `.responded`. Only the remaining errors are inconclusive.
- "The raw `AXError` goes into the diagnostic log" is ambiguous in a repo that has `PersistentDiagnosticLog` and `DiagnosticExporter`, neither of which carries this. It goes to the standalone `debugLog` appender at `/tmp/mt-permission.log`, which truncates per process start.

A shorter version carrying the same load-bearing facts:

> `.broken` means TCC says allowed but the probe disagrees. The Accessibility probe is a cross-process call, so only `kAXErrorAPIDisabled` counts as `.broken`; `.success`/`.noValue` are `.responded` and every other AXError is `AccessibilityProbe.inconclusive`, reporting healthy (treating those as broken produced a false "toggle it off and on" notification for a working grant). `kAXErrorNotImplemented` stays inconclusive on purpose: Apple documents it as the asked process lacking AX support. The raw code is appended to `/tmp/mt-permission.log`.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: "most often the frontmost app being slow or not answering AX at all" is an inference, and I think it is the wrong one.

What I measured on a machine whose grant was fine: `AXIsProcessTrusted()` true, and this exact system-wide query returning `cannotComplete` on every attempt, immediately rather than after a delay, while in the same process app-targeted calls against Finder, the frontmost app and Dock all returned `.success` with real titles and window counts. A slow or unresponsive frontmost app does not explain that shape.

Suggest describing what is known and leaving the mechanism open, something like "the call failed for a reason that is not about our permission; the system-wide query has failure modes of its own". The verdict logic does not depend on the explanation, so a weaker sentence costs nothing.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: This paragraph repeats what the CLAUDE.md bullet in this same PR already says, and it pins two things that are weaker than the wording suggests: the wall-clock timestamps are one machine's session and will read as noise later, and "measured consequence" covers the flap (which was measured) but sits next to a cause that was not.

The trim below also buys the five lines the file needs, see the note on `probeAccessibility` below. With it the file is 599 lines and SwiftLint `--strict` passes; the extraction suggestion alone would be 602 and fail.

```suggestion
    /// **An inconclusive probe reports `.healthy`, not `.broken`.** This used to
    /// be a `Bool` where every non-success collapsed to "broken", and the
    /// reported consequence was a false alarm telling users to toggle a
    /// permission that was fine: one process, no user action, and the verdict
    /// flipping healthy -> broken -> healthy inside 70 seconds. A revoked grant
    /// does not repair itself in a minute.
    ///
    /// `.broken` exists for exactly one situation, TCC saying yes while the API
    /// says no, and `AXIsProcessTrusted()` is already the system's own answer
    /// about our grant. So the probe only has to catch that contradiction, which
    /// is `.apiDisabled`.
```

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: Follow-up rather than a change request for this PR.

Since `.broken` now hangs entirely on one error code from this one query, it is worth asking whether this query can produce it at all. `ParticipantReader` does not use the system-wide element; it targets an application directly, which is where an untrusted client would actually surface `kAXErrorAPIDisabled`. An app-targeted probe would make `.broken` observable instead of theoretical.

Also relevant: commit 85407512 (issue #446) hit the same false-broken problem in this file for Screen Recording and answered it by dropping the probe from the verdict altogether. Keeping one path here may be the better call, but the PR reads as if the precedent does not exist.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: This `switch` is the fix, and nothing exercises it. `probeAccessibility` is referenced only at its definition and its one call site in `checkAccessibilityLive`; every new test builds an `AccessibilityProbe` by hand and passes it to `checkAccessibility(trusted:probe:)`, which never inspects the associated value.

Measured rather than argued: I removed the `.apiDisabled` arm below, so the only remaining route to `.broken` was gone, and `swift test --filter PermissionHealthCheck` still reported 53 of 53 passing. Regrouping `.notImplemented` into that arm is invisible in the same way. That is also why the PR description's "There is a test pinning that distinction" does not hold.

Extracting the mapping puts it on the layer this repo asks for ("pure logic first"):

```suggestion
        return classify(err)
    }

    /// Pure `AXError` -> probe classification, split out of
    /// `probeAccessibility()` so a test can reach it: `checkAccessibility`
    /// never reads the associated value, so nothing else pins this mapping.
    static func classify(_ err: AXError) -> AccessibilityProbe {
        switch err {
        // .success: got the focused app. .noValue: nothing focused right now,
        // but the API answered, which is what we asked.
        case .success, .noValue: .responded

        // ONLY `.apiDisabled`. Apple documents it as "the accessibility API is
        // disabled", which is about our access. `.notImplemented` reads like a
        // sibling and is not one: its own documentation says it "can be returned
        // if a PROCESS does not support the accessibility API" — the process
        // being asked. Putting it here would re-create the false alarm for every
        // frontmost app with incomplete AX support.
        case .apiDisabled: .apiDisabled

        default: .inconclusive(err)
        }
    }
```

Heads up: applied alone this makes the file 602 lines and SwiftLint `--strict` fails with `File Length Violation`. It needs the trim suggested on the doc comment above, which lands at 599.

With both applied plus these four tests, I measured 12 of 12 green, and `case .apiDisabled, .notImplemented: .apiDisabled` then fails `testClassifyKeepsNotImplementedOutOfTheBrokenSignal`:

```swift
    func testClassifyMapsOnlyApiDisabledToTheBrokenSignal() {
        XCTAssertEqual(PermissionHealthCheck.classify(.apiDisabled), .apiDisabled)
    }

    func testClassifyKeepsNotImplementedOutOfTheBrokenSignal() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.notImplemented),
            .inconclusive(.notImplemented),
            "notImplemented describes the process being asked, not our grant",
        )
    }

    func testClassifyTreatsAnsweredCallsAsResponded() {
        XCTAssertEqual(PermissionHealthCheck.classify(.success), .responded)
        XCTAssertEqual(PermissionHealthCheck.classify(.noValue), .responded)
    }

    func testClassifyTreatsCannotCompleteAsInconclusive() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.cannotComplete),
            .inconclusive(.cannotComplete),
        )
    }
```

This is also what the Codecov failure is pointing at. In a local full run with coverage, `probeAccessibility()` shows an execution count of 0 across all 3072 tests, and every uncovered added line in this file sits in this function or in `checkAccessibilityLive`. With the extraction and the four tests above, the classifier is executed 5 times and patch coverage on this file goes from 34.4% to 77.1% by my measurement, which clears the 50% target.

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
pasrom: `.noValue` cannot reach this assertion: `probeAccessibility` maps it to `.responded`, so `.inconclusive(.noValue)` is a state production never builds. The docstring above ("Every non-`apiDisabled` error takes the inconclusive path") therefore describes the `default:` arm, which this test does not execute.

Since `checkAccessibility` ignores the payload, the loop asserts one code path four times, already covered by the previous test. It becomes meaningful once it runs against `classify`, per the other comment. Minimal fix meanwhile:

```suggestion
        for error in [AXError.failure, .invalidUIElement, .illegalArgument, .attributeUnsupported] {
```

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
pasrom: The name says "Probe treats", but this calls `checkAccessibility`, not the probe. Someone grepping the test class for probe coverage will find this and conclude the classification is pinned, which is exactly the gap flagged above.

The second assertion here is also a verbatim repeat of `testAccessibilityBroken`. Either point this at `classify` (then the name becomes true) or rename it to what it currently checks.

CLAUDE.md
pasrom: This bullet goes from 316 to 1265 characters and largely restates the doc comments added to `PermissionHealthCheck.swift` in the same PR. Long bullets are this file's habit, so length alone is not the objection; the duplication is, because the two copies will drift.

Two accuracy points regardless of length:

- "everything else is `AccessibilityProbe.inconclusive`" is not right: `.success` and `.noValue` become `.responded`. Only the remaining errors are inconclusive.
- "The raw `AXError` goes into the diagnostic log" is ambiguous in a repo that has `PersistentDiagnosticLog` and `DiagnosticExporter`, neither of which carries this. It goes to the standalone `debugLog` appender at `/tmp/mt-permission.log`, which truncates per process start.

A shorter version carrying the same load-bearing facts:

> `.broken` means TCC says allowed but the probe disagrees. The Accessibility probe is a cross-process call, so only `kAXErrorAPIDisabled` counts as `.broken`; `.success`/`.noValue` are `.responded` and every other AXError is `AccessibilityProbe.inconclusive`, reporting healthy (treating those as broken produced a false "toggle it off and on" notification for a working grant). `kAXErrorNotImplemented` stays inconclusive on purpose: Apple documents it as the asked process lacking AX support. The raw code is appended to `/tmp/mt-permission.log`.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: "most often the frontmost app being slow or not answering AX at all" is an inference, and I think it is the wrong one.

What I measured on a machine whose grant was fine: `AXIsProcessTrusted()` true, and this exact system-wide query returning `cannotComplete` on every attempt, immediately rather than after a delay, while in the same process app-targeted calls against Finder, the frontmost app and Dock all returned `.success` with real titles and window counts. A slow or unresponsive frontmost app does not explain that shape.

Suggest describing what is known and leaving the mechanism open, something like "the call failed for a reason that is not about our permission; the system-wide query has failure modes of its own". The verdict logic does not depend on the explanation, so a weaker sentence costs nothing.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: This paragraph repeats what the CLAUDE.md bullet in this same PR already says, and it pins two things that are weaker than the wording suggests: the wall-clock timestamps are one machine's session and will read as noise later, and "measured consequence" covers the flap (which was measured) but sits next to a cause that was not.

The trim below also buys the five lines the file needs, see the note on `probeAccessibility` below. With it the file is 599 lines and SwiftLint `--strict` passes; the extraction suggestion alone would be 602 and fail.

```suggestion
    /// **An inconclusive probe reports `.healthy`, not `.broken`.** This used to
    /// be a `Bool` where every non-success collapsed to "broken", and the
    /// reported consequence was a false alarm telling users to toggle a
    /// permission that was fine: one process, no user action, and the verdict
    /// flipping healthy -> broken -> healthy inside 70 seconds. A revoked grant
    /// does not repair itself in a minute.
    ///
    /// `.broken` exists for exactly one situation, TCC saying yes while the API
    /// says no, and `AXIsProcessTrusted()` is already the system's own answer
    /// about our grant. So the probe only has to catch that contradiction, which
    /// is `.apiDisabled`.
```

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: Follow-up rather than a change request for this PR.

Since `.broken` now hangs entirely on one error code from this one query, it is worth asking whether this query can produce it at all. `ParticipantReader` does not use the system-wide element; it targets an application directly, which is where an untrusted client would actually surface `kAXErrorAPIDisabled`. An app-targeted probe would make `.broken` observable instead of theoretical.

Also relevant: commit 85407512 (issue #446) hit the same false-broken problem in this file for Screen Recording and answered it by dropping the probe from the verdict altogether. Keeping one path here may be the better call, but the PR reads as if the precedent does not exist.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: This `switch` is the fix, and nothing exercises it. `probeAccessibility` is referenced only at its definition and its one call site in `checkAccessibilityLive`; every new test builds an `AccessibilityProbe` by hand and passes it to `checkAccessibility(trusted:probe:)`, which never inspects the associated value.

Measured rather than argued: I removed the `.apiDisabled` arm below, so the only remaining route to `.broken` was gone, and `swift test --filter PermissionHealthCheck` still reported 53 of 53 passing. Regrouping `.notImplemented` into that arm is invisible in the same way. That is also why the PR description's "There is a test pinning that distinction" does not hold.

Extracting the mapping puts it on the layer this repo asks for ("pure logic first"):

```suggestion
        return classify(err)
    }

    /// Pure `AXError` -> probe classification, split out of
    /// `probeAccessibility()` so a test can reach it: `checkAccessibility`
    /// never reads the associated value, so nothing else pins this mapping.
    static func classify(_ err: AXError) -> AccessibilityProbe {
        switch err {
        // .success: got the focused app. .noValue: nothing focused right now,
        // but the API answered, which is what we asked.
        case .success, .noValue: .responded

        // ONLY `.apiDisabled`. Apple documents it as "the accessibility API is
        // disabled", which is about our access. `.notImplemented` reads like a
        // sibling and is not one: its own documentation says it "can be returned
        // if a PROCESS does not support the accessibility API" — the process
        // being asked. Putting it here would re-create the false alarm for every
        // frontmost app with incomplete AX support.
        case .apiDisabled: .apiDisabled

        default: .inconclusive(err)
        }
    }
```

Heads up: applied alone this makes the file 602 lines and SwiftLint `--strict` fails with `File Length Violation`. It needs the trim suggested on the doc comment above, which lands at 599.

With both applied plus these four tests, I measured 12 of 12 green, and `case .apiDisabled, .notImplemented: .apiDisabled` then fails `testClassifyKeepsNotImplementedOutOfTheBrokenSignal`:

```swift
    func testClassifyMapsOnlyApiDisabledToTheBrokenSignal() {
        XCTAssertEqual(PermissionHealthCheck.classify(.apiDisabled), .apiDisabled)
    }

    func testClassifyKeepsNotImplementedOutOfTheBrokenSignal() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.notImplemented),
            .inconclusive(.notImplemented),
            "notImplemented describes the process being asked, not our grant",
        )
    }

    func testClassifyTreatsAnsweredCallsAsResponded() {
        XCTAssertEqual(PermissionHealthCheck.classify(.success), .responded)
        XCTAssertEqual(PermissionHealthCheck.classify(.noValue), .responded)
    }

    func testClassifyTreatsCannotCompleteAsInconclusive() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.cannotComplete),
            .inconclusive(.cannotComplete),
        )
    }
```

This is also what the Codecov failure is pointing at. In a local full run with coverage, `probeAccessibility()` shows an execution count of 0 across all 3072 tests, and every uncovered added line in this file sits in this function or in `checkAccessibilityLive`. With the extraction and the four tests above, the classifier is executed 5 times and patch coverage on this file goes from 34.4% to 77.1% by my measurement, which clears the 50% target.

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
pasrom: `.noValue` cannot reach this assertion: `probeAccessibility` maps it to `.responded`, so `.inconclusive(.noValue)` is a state production never builds. The docstring above ("Every non-`apiDisabled` error takes the inconclusive path") therefore describes the `default:` arm, which this test does not execute.

Since `checkAccessibility` ignores the payload, the loop asserts one code path four times, already covered by the previous test. It becomes meaningful once it runs against `classify`, per the other comment. Minimal fix meanwhile:

```suggestion
        for error in [AXError.failure, .invalidUIElement, .illegalArgument, .attributeUnsupported] {
```

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
pasrom: The name says "Probe treats", but this calls `checkAccessibility`, not the probe. Someone grepping the test class for probe coverage will find this and conclude the classification is pinned, which is exactly the gap flagged above.

The second assertion here is also a verbatim repeat of `testAccessibilityBroken`. Either point this at `classify` (then the name becomes true) or rename it to what it currently checks.

CLAUDE.md
pasrom: This bullet goes from 316 to 1265 characters and largely restates the doc comments added to `PermissionHealthCheck.swift` in the same PR. Long bullets are this file's habit, so length alone is not the objection; the duplication is, because the two copies will drift.

Two accuracy points regardless of length:

- "everything else is `AccessibilityProbe.inconclusive`" is not right: `.success` and `.noValue` become `.responded`. Only the remaining errors are inconclusive.
- "The raw `AXError` goes into the diagnostic log" is ambiguous in a repo that has `PersistentDiagnosticLog` and `DiagnosticExporter`, neither of which carries this. It goes to the standalone `debugLog` appender at `/tmp/mt-permission.log`, which truncates per process start.

A shorter version carrying the same load-bearing facts:

> `.broken` means TCC says allowed but the probe disagrees. The Accessibility probe is a cross-process call, so only `kAXErrorAPIDisabled` counts as `.broken`; `.success`/`.noValue` are `.responded` and every other AXError is `AccessibilityProbe.inconclusive`, reporting healthy (treating those as broken produced a false "toggle it off and on" notification for a working grant). `kAXErrorNotImplemented` stays inconclusive on purpose: Apple documents it as the asked process lacking AX support. The raw code is appended to `/tmp/mt-permission.log`.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
```diff
@@ -462,8 +521,11 @@ enum PermissionHealthCheck {
             return .denied
         }
         let probe = probeAccessibility()
-        let r = checkAccessibility(trusted: trusted, probeSucceeds: probe)
-        debugLog("checkAccessibilityLive: trusted=true probe=\(probe) → \(r)")
+        let r = checkAccessibility(trusted: trusted, probe: probe)
+        // The raw AXError is logged, not just the verdict. Without it the false
+        // alarm above could only be diagnosed by noticing the timing, because a
+        // `Bool` had already thrown away the one fact that explains it.
+        debugLog("checkAccessibilityLive: trusted=true probe=\(probe.logToken) → \(r)")
```
pasrom: This is the only place the new fact lands, and it does not survive a relaunch.

`DebugLogFile.init` calls `removeItem(atPath:)`, so `/tmp/mt-permission.log` is truncated on first write per process, and `/tmp` is periodically cleared by the OS on top of that. The open question this PR leaves behind, which `AXError` actually flapped, can therefore only be answered by someone who copies the file before the app is next launched.

Sharper version of the same problem: `runLive` emits its `os_log` warning only when `!result.isHealthy`. Before this change an inconclusive probe resolved to `.broken`, so the event was recorded publicly in `os_log` as `accessibility=broken` and survived restarts. Now it resolves to `.healthy`, so `os_log` records nothing. Net effect, the durable trace of this event is removed in the same commit that adds the volatile one. Commit 35f6eecb went the other way deliberately, making the failure reason public in `os_log`.

Not a blocker for the verdict logic, which I think is right, but worth a decision now rather than after the next report arrives with no log attached. Cheapest option is one `logger.info` here alongside the existing `debugLog`, gated the same way, so an inconclusive probe stays visible in Console without re-badging anything for the user.

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
```diff
@@ -0,0 +1,80 @@
...
+        )
+    }
+
+    /// The verdict alone could not explain the false alarm; the raw code can.
+    func testAccessibilityProbeLogTokenCarriesTheRawErrorCode() {
+        XCTAssertEqual(
+            PermissionHealthCheck.AccessibilityProbe.inconclusive(.cannotComplete).logToken,
+            "inconclusive(AXError \(AXError.cannotComplete.rawValue))",
+        )
+        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.responded.logToken, "responded")
```
pasrom: `.apiDisabled` is the one case whose `logToken` nobody asserts, and it is the case that accompanies the only user-visible verdict this PR still produces. Coverage backs that up: `logToken` shows 2 hits in a full local run, one per assertion here.

```suggestion
        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.responded.logToken, "responded")
        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.apiDisabled.logToken, "apiDisabled")
```

app/MeetingTranscriber/Tests/PermissionHealthCheckTests.swift
```diff
@@ -117,29 +117,6 @@ final class PermissionHealthCheckTests: XCTestCase {
...
-        let result = PermissionHealthCheck.checkAccessibility(trusted: false, probeSucceeds: true)
-        XCTAssertEqual(result, .denied)
-    }
-
-    func testAccessibilityBroken() {
-        let result = PermissionHealthCheck.checkAccessibility(trusted: true, probeSucceeds: false)
-        XCTAssertEqual(result, .broken)
-    }
-
     // MARK: - Overall Health
```
pasrom: Minor: the four Accessibility cases left this file and nothing here says where they went. The new file's header points back to this one, but not the other way round, so someone extending the Accessibility verdicts will most likely land here first and conclude there is no coverage.

A one-line `// MARK: - Accessibility: see PermissionHealthCheckAccessibilityTests.swift` where the removed section used to sit would close that, and it costs nothing against the 400-line budget that motivated the split.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: I would leave the mechanism open here rather than name it.

I tried to reproduce your flap and could not, and the attempt is worth reporting because it cuts against my own first guess as much as against this sentence. In a plain command-line process on my machine this call returns `cannotComplete` every time, which looked like a smoking gun until I built the same probe inside a process with an `NSApplication` run loop, which is what the app actually is. There it succeeded 40 out of 40. So I have no evidence that a slow frontmost app was involved, and none that it was not. The old `Bool` discarded the code, so right now nobody can say.

The verdict logic does not depend on the explanation, so this can say less and stay true:

```suggestion
        /// The call failed for a reason that has not been shown to be about our
        /// permission. Carries the raw code for the log, so the next occurrence
        /// can be attributed instead of guessed.
```

Once the enriched log line has caught one real flap, the comment can name the code with something behind it.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: Two details I would drop here: the wall-clock times, which belong to one session, and "measured", which is true of the flap but not of the cause sitting next to it.

There is also a mechanical reason to shorten something in this file: with the extraction suggested below, it reaches 602 lines and `swiftlint --strict` fails on `file_length` (limit 600). With this trim it lands at 599.

Since this is your prose, spelling out everything the suggestion changes beyond those two points: it drops the closing two sentences ("Every other AX error describes the element that was asked..."), drops the clause "; an unresponsive frontmost app does" for the reason in the other comment, and turns two em-dashes into commas. Push back on any of those, they are preference rather than correctness.

```suggestion
    /// **An inconclusive probe reports `.healthy`, not `.broken`.** This used to
    /// be a `Bool` where every non-success collapsed to "broken", and the
    /// reported consequence was a false alarm telling users to toggle a
    /// permission that was fine: one process, no user action, and the verdict
    /// flipping healthy -> broken -> healthy inside 70 seconds. A revoked grant
    /// does not repair itself in a minute.
    ///
    /// `.broken` exists for exactly one situation, TCC saying yes while the API
    /// says no, and `AXIsProcessTrusted()` is already the system's own answer
    /// about our grant. So the probe only has to catch that contradiction, which
    /// is `.apiDisabled`.
```

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom: This switch is the fix, and nothing exercises it. `probeAccessibility` is referenced only at its definition and its one call site in `checkAccessibilityLive`; seven of the eight new tests build an `AccessibilityProbe` by hand and pass it to `checkAccessibility(trusted:probe:)`, which never inspects the associated value (the eighth tests `logToken`).

To check rather than assume, I removed the `.apiDisabled` arm below, so the only remaining route to `.broken` was gone, and the filtered suite still passed 53 of 53. Regrouping `.notImplemented` into that arm is invisible in the same way.

Extracting the mapping puts it on the layer this repo asks for ("pure logic first"):

```suggestion
        return classify(err)
    }

    /// Pure `AXError` -> probe classification, split out of
    /// `probeAccessibility()` so a test can reach it: `checkAccessibility`
    /// never reads the associated value, so nothing else pins this mapping.
    static func classify(_ err: AXError) -> AccessibilityProbe {
        switch err {
        // .success: got the focused app. .noValue: nothing focused right now,
        // but the API answered, which is what we asked.
        case .success, .noValue: .responded

        // ONLY `.apiDisabled`. Apple documents it as "the accessibility API is
        // disabled", which is about our access. `.notImplemented` reads like a
        // sibling and is not one: its own documentation says it "can be returned
        // if a PROCESS does not support the accessibility API" — the process
        // being asked. Putting it here would re-create the false alarm for every
        // frontmost app with incomplete AX support.
        case .apiDisabled: .apiDisabled

        default: .inconclusive(err)
        }
    }
```

**Apply this and the doc-comment suggestion above in the same push.** On its own this one makes the file 602 lines and `swiftlint --strict` fails on `file_length`; together they land at 599.

End state for `PermissionHealthCheckAccessibilityTests.swift`: add these four, and `testAccessibilityProbeTreatsNotImplementedAsInconclusive` can then go, since the second of these replaces it.

```swift
    func testClassifyMapsOnlyApiDisabledToTheBrokenSignal() {
        XCTAssertEqual(PermissionHealthCheck.classify(.apiDisabled), .apiDisabled)
    }

    func testClassifyKeepsNotImplementedOutOfTheBrokenSignal() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.notImplemented),
            .inconclusive(.notImplemented),
            "notImplemented describes the process being asked, not our grant",
        )
    }

    func testClassifyTreatsAnsweredCallsAsResponded() {
        XCTAssertEqual(PermissionHealthCheck.classify(.success), .responded)
        XCTAssertEqual(PermissionHealthCheck.classify(.noValue), .responded)
    }

    func testClassifyTreatsCannotCompleteAsInconclusive() {
        XCTAssertEqual(
            PermissionHealthCheck.classify(.cannotComplete),
            .inconclusive(.cannotComplete),
        )
    }
```

I ran this combination: 12 of 12 green, and regrouping `.notImplemented` in `classify` then fails `testClassifyKeepsNotImplementedOutOfTheBrokenSignal`, which is the point.

app/MeetingTranscriber/Sources/PermissionHealthCheck.swift
pasrom, line 528: Something to decide rather than a change request, and I will most likely handle it myself.

After this change an inconclusive probe resolves to `.healthy`, and `runLive` emits its `os_log` warning only when `!result.isHealthy`, so the event stops appearing in `os_log` at all. That is mostly correct: what used to appear there was the false alarm, and removing it is the point of the PR.

The side effect is that this enriched line becomes the only record, and `DebugLogFile.init` calls `removeItem(atPath:)`, so `/tmp/mt-permission.log` is truncated on first write per process and does not survive a relaunch. That only matters because of the open question in the other comment: catching one real flap then depends on someone copying the file before the next launch. I put the failure reason into `os_log` deliberately in 35f6eecb and would rather not lose the thread, so expect a follow-up from me touching this line.

For the record, in case it saves you an experiment: an `os_log` `.info` here would not survive a restart either. It would be picked up by `PersistentDiagnosticLog` while the app is running, which is a different guarantee from the `.warning` that used to fire.

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
pasrom: `.noValue` cannot reach this assertion: `probeAccessibility` maps it to `.responded`, so `.inconclusive(.noValue)` is a state production never builds.

```suggestion
        for error in [AXError.failure, .invalidUIElement, .illegalArgument, .attributeUnsupported] {
```

app/MeetingTranscriber/Tests/PermissionHealthCheckAccessibilityTests.swift
pasrom, line 121: `.apiDisabled` is the one `logToken` case nobody asserts, and it is the one that accompanies the only user-visible verdict this still produces.

```suggestion
        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.responded.logToken, "responded")
        XCTAssertEqual(PermissionHealthCheck.AccessibilityProbe.apiDisabled.logToken, "apiDisabled")
```

CLAUDE.md
pasrom: Two accuracy points rather than length:

- "everything else is `AccessibilityProbe.inconclusive`" is not quite right: `.success` and `.noValue` become `.responded`, and only the remaining errors are inconclusive.
- "The raw `AXError` goes into the diagnostic log" is ambiguous in a repo that has `PersistentDiagnosticLog` and `DiagnosticExporter`, neither of which carries this. It goes to the `debugLog` appender at `/tmp/mt-permission.log`, which truncates per process.

Both folded in, and shortened where it repeats the doc comment you already added to the source:

```suggestion
- `PermissionHealthCheck` verifies each TCC permission by combining the system verdict with a live probe. Each resolves to `PermissionStatus` (`.healthy | .denied | .broken | .notDetermined`). `.broken` means TCC says allowed but the probe disagrees — fix is to toggle the permission off and on in System Settings. The Accessibility probe is a cross-process call, so only `kAXErrorAPIDisabled` counts as `.broken`; `.success` and `.noValue` are `AccessibilityProbe.responded`, and every other AXError is `.inconclusive` and reports healthy. Treating those as broken produced a notification telling users to toggle a permission that was working. `kAXErrorNotImplemented` stays inconclusive on purpose: Apple documents it as the asked process lacking AX support. The raw code is appended to `/tmp/mt-permission.log`, which `debugLog` truncates per process.
```

pasrom (commented):
Thanks for this. The bug is real, the diagnosis is good, and I want the fix in.

Since most of what follows is detail, first what is right: the three-case enum instead of a `Bool` is the correct shape, keeping `.notImplemented` out of the broken signal is well argued and I agree with the reasoning, rejecting hysteresis is the right call, splitting the test file to stay under the 400-line type body limit is what I would have done too, and diagnosing a permission bug from nothing but the timing of three log lines is good work.

**Two things I would like changed before this lands.**

**1. The mapping this PR changes is not covered.** Every new test goes through `checkAccessibility(trusted:probe:)`, which ignores the associated value, so the classification inside `probeAccessibility` is never executed. To check, I deleted the `.apiDisabled` arm outright and the filtered suite still passed 53 of 53. The comment on that switch has a suggestion that extracts a pure `classify(_:)`, plus the four tests to go with it. This is the same gap the red `codecov/patch` is reporting: on Codecov's own counting the patch goes from 30.8% to 75.0% once the classifier is reachable.

**2. The comments name a cause that was never recorded.** See the comment on the `.inconclusive` case. I tried to reproduce your flap and could not, which is a reason to leave the wording open rather than to argue for some other cause.

**Smaller, your call:** two accuracy points in the CLAUDE.md bullet, the `.noValue` entry in the error loop, and one missing `logToken` assertion. Each is a one-click suggestion.

**Not your job, I will handle it:** the `os_log` question I raise on the `debugLog` line, whether the probe should target an application instead of the system-wide element, and a sentence recording that commit 85407512 (issue #446) hit this same shape for Screen Recording and answered it by dropping the probe from the verdict entirely. You had no way to know that one existed.

**Path to merge:** apply both suggestions on `PermissionHealthCheck.swift` in the same push, because either one alone puts the file over the 600-line `file_length` limit that `--strict` turns into an error. Add the four tests to `PermissionHealthCheckAccessibilityTests.swift`, drop `.noValue` from the loop, and I will re-run and merge. One request on the description: "There is a test pinning that distinction" is not quite right yet, since the tests pin the verdict on a hand-built enum rather than the probe's classification. The suggested tests would make it true.

egerkuzma:
Thanks — the coverage point was right, and demonstrating it by deleting the `.apiDisabled` arm was more convincing than any argument about it. Everything from **Path to merge** is applied, in one push:

- **Both source suggestions together.** `classify(_:)` is extracted and `probeAccessibility()` keeps only the call; the doc comment is trimmed as suggested. The file lands at 595 lines, so `--strict` is happy — thank you for flagging that either one alone would break it.
- **The four tests**, under your names. Mutation-checked rather than assumed: removing the `.apiDisabled` arm now fails `testClassifyMapsOnlyApiDisabledToTheBrokenSignal`, and regrouping `.notImplemented` into it fails `testClassifyKeepsNotImplementedOutOfTheBrokenSignal`. Both passed before.
- **`.noValue` dropped from the loop** — you are right that `classify` maps it to `.responded`, so that state is never built in production.
- **Description fixed.** The sentence now names the test and says what it pins.

Also took the two optional ones: the `logToken` assertion for `.apiDisabled`, and both accuracy points in the CLAUDE.md bullet (your wording).

On the cause: your two experiments are the most useful thing in the review, because they cut against my sentence and your own first guess at once. The comment no longer names a mechanism — it says the cause is not established and points at the enriched log line. Which makes your `debugLog` question the load-bearing one now: with an inconclusive probe resolving to `.healthy`, that line is the only record, and it does not survive a relaunch. Happy to leave that to you as you suggested, or to take a follow-up if you would rather.

I did not know about 85407512 dropping the probe from the Screen Recording verdict entirely. If you would prefer that shape here too, say so and I will rework it rather than defend this one.

pasrom:
Checked rather than taken on trust, and it holds up. Both mutations are now caught: deleting the `.apiDisabled` arm fails `testClassifyMapsOnlyApiDisabledToTheBrokenSignal`, and regrouping `.notImplemented` fails `testClassifyKeepsNotImplementedOutOfTheBrokenSignal` plus the enumeration test. Before this push both were invisible. 13 of 13 green, lint clean across 541 files, and patch coverage lands at 66.7% on Codecov's counting against the 50% target. CI is green including `codecov/patch`.

Two things you did better than what I suggested. Putting `classify` before the call and moving the `.notImplemented` rationale into the doc comment is tighter than my version, and it is why the file came out at 595 rather than the 599 I predicted. And `testClassifyTreatsEveryOtherErrorAsInconclusive` enumerating all thirteen remaining errors instead of sampling is the stronger choice, for exactly the reason your comment gives: a future regrouping of any one of them now has to come to that test and say so.

On the #446 shape: keep what you have, no rework. The reason the Screen Recording fix could drop the probe outright is that its probe had no way to express a contradiction; an empty window list is simply not evidence. Here `AXIsProcessTrusted()` saying yes while the API answers `kAXErrorAPIDisabled` is a real contradiction, it costs one switch arm, and it is now pinned by a test.

One caveat you should have, because it cuts against my own recommendation: nobody has demonstrated that `kAXErrorAPIDisabled` actually occurs in practice. Establishing it needs an untrusted process, which I did not want to arrange, so the branch may be effectively unreachable. If that ever gets shown, removing it becomes a separate cleanup that also has to deal with `PermissionProblem.accessibilityBroken` and the Settings row. That is not this PR's job either way.

The `debugLog` follow-up is yours if you want it, and I would rather you took it since you have the context now. The question worth answering there is whether an inconclusive probe should leave any durable trace at all, not just how to log it.

Nothing blocking from me. Thanks for the careful turnaround, and for pushing back where you thought I was wrong.

pasrom (approved):
Approving. Every point from the review is addressed, and I verified the load-bearing one rather than taking the summary on trust: both mutations are now caught where both were invisible before, 13 of 13 green, lint clean, and CI is fully green including `codecov/patch`.

The remaining open thread on the `debugLog` line is a follow-up, not a condition. Thanks for the thorough turnaround.

PR #710, opened 2026-09-11
fix(app): source failure reason from stream-json result event

GB-D:
## Summary

Follow-up to your 11 Sep review on #692. You flagged that redacting `ClaudeCLIProtocolGenerator`'s and `PipelineQueue+Stages.swift`'s failure logs to `.private` also silenced every failure reason that never carried content in the first place — `cliNotFound`, `timeout`, `emptyProtocol`, an OpenAI provider error — and offered this as your preferred, structural fix over the `.private` stopgap: source the failure message from the stream-json terminal `result` event instead of the accumulated assistant text.

- Verified the event's real shape against the installed `claude` binary itself (its embedded zod schema and construction sites), since the format is undocumented (#24612) and I found conflicting descriptions elsewhere. Three shapes exist: success (`result: <text>`), a structural failure like max-turns or a stalled sandbox (`errors: [<string>...]`, no `result` field), and the in-turn API error that actually caused the original 8-meeting bug — `subtype` stays `"success"`, but `is_error: true` and `result` holds the error sentence instead of content.
- `parseResultEvent`/`resultEventFailureReason` extract that. `handleFailure` now builds the message shown to the user and logged `.public` from the result event (or real stderr, or a fixed placeholder) — never from the accumulated text, which stays `.private` exactly as before. Both log sites go back to `.public`.
- Deliberately does **not** fall back to the accumulated text if the result event is ever absent/malformed — that would reopen the leak your 4th commit on #692 closed. That path falls to a fixed placeholder instead; not yet observed in practice, since a real CLI run always emits a terminal `result` line.
- `drainStreamJSONLines` keeps its existing one-argument signature as a thin wrapper, so no other call site needed to change.
- New tests moved into their own file (`ClaudeCLIProtocolGeneratorResultEventTests.swift`) to stay under the 600-line lint ceiling, following the same split you did in 83b15da.

## Test plan

- [x] `swift test --parallel` — full suite, 3159 tests, all green
- [x] `./scripts/lint.sh` — 0 violations, 553 files
- [x] `./scripts/pre-push.sh` — release build parity, passed
- [x] Rewrote the one existing test whose fake CLI script didn't emit a terminal `result` line (unrealistic — the real CLI always does) and added a second test for the genuine no-result-event fallback

### DIFF
--- a/app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
+++ b/app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
@@ -111,7 +111,7 @@
             }
 
             // Read stream-json output concurrently with stdin write
-            let text = try await Self.readStreamJSON(from: stdoutPipe, process: process)
+            let (text, resultEvent) = try await Self.readStreamJSON(from: stdoutPipe, process: process)
 
             // Ensure stdin write completes (should be done by now)
             _ = await stdinWriteTask.value
@@ -128,7 +128,9 @@
 
             if process.terminationStatus != 0 {
                 let stderrData = await stderrRead
-                throw Self.handleFailure(exitCode: process.terminationStatus, stderrData: stderrData, text: text)
+                throw Self.handleFailure(
+                    exitCode: process.terminationStatus, stderrData: stderrData, text: text, resultEvent: resultEvent,
+                )
             }
 
             return try Self.validateGeneratedText(text)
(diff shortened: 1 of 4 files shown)


app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -215,6 +254,14 @@
         /// for the next call to consume — the caller must keep the buffer
         /// across iterations.
         static func drainStreamJSONLines(buffer: inout Data) -> [String] {
```
pasrom, line 256: This overload now has no production caller. Line 241 uses the two argument version, and this one is reached only from the ten call sites in the test files.

The PR description says keeping it means "no other call site needed to change", but the call sites it spares are all tests, so what it actually buys is not touching ten test lines, at the cost of a production entry point that exists for the tests alone.

Same shape one function up: `selectFailureMessage` no longer has a caller either, and its four tests now pin something nothing runs.

Neither is a defect and I would not hold the PR for either. But a year from now both read as live API. I would either update the ten test call sites and delete this overload, or leave a line saying it is kept deliberately for the existing tests.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -227,6 +274,9 @@
                 if let text = parseStreamJSONLine(line) {
                     fragments.append(text)
                 }
+                if let parsed = parseResultEvent(line) {
```
pasrom, line 277: Every line now gets decoded twice, once here and once in `parseStreamJSONLine` just above, both through `JSONSerialization`.

At the volume this sees it does not matter: a real run emitted around twenty lines. Only worth folding if you are touching this anyway, for instance by having a single decode return the object and letting both extractors read from it. Mentioning it so it is a conscious choice rather than an oversight.

app/MeetingTranscriber/Sources/ClaudeCLIProtocolGenerator.swift
```diff
@@ -261,6 +311,55 @@
...
+            return ResultEventInfo(isError: isError, errors: errors, result: result)
+        }
+
+        /// The CLI's own diagnostic sentence for a failed run, sourced from
+        /// the terminal `result` event rather than the accumulated assistant
+        /// text. Prefers `errors` (joined); falls back to `result` when
+        /// `isError` is set. Returns nil when the event is absent or carries
+        /// nothing usable.
+        static func resultEventFailureReason(_ resultEvent: ResultEventInfo?) -> String? {
+            guard let resultEvent, resultEvent.isError else { return nil }
```
pasrom, line 378: Worth writing one more sentence here, because the safety of this whole change rests on a property of the CLI that the code cannot check.

`result` is a dual purpose field. On a successful run it holds the generated content; on a failed one it holds the error sentence. `is_error` is the only thing that tells them apart, since `subtype` reads `"success"` in both cases. I confirmed both shapes against real runs. So the guarantee "never generated content" is not a property of the field, it is a property of the CLI's current behaviour with an undocumented format, and if a future version ever sets `is_error: true` while `result` still carries content, this function hands it straight to a `.public` log line.

I went looking for a cheap guard and did not find one. The obvious candidate is to reject a `result` that matches the accumulated text, but that fails: in the auth failure case the CLI emits the same sentence both as an assistant message and in `result`, byte for byte, so the accumulated text and `result` are identical and such a guard would reject the legitimate reason and silently degrade every failure to the placeholder.

So I am not asking for a guard. I am asking for the invariant to be stated where the next person changing this will read it: `is_error` is the only discriminator, and the content safety of the `.public` log depends on it.

If you want a belt and braces version anyway, gating the `result` fallback on a known `terminal_reason` (the failing run carried `"api_error"`) fails closed rather than open: an unrecognised value yields the placeholder rather than a leak. Your call, and I would merge this without it.

pasrom (commented):
This is the change I hoped for, and it is better argued than my suggestion was. Both log sites are back to `.public` without the accumulated text ever reaching either of them, and the two end to end tests pin the behaviour rather than the plumbing. `testGenerateOnFailureFallsBackToPlaceholderWhenNoResultEventOrStderr` is the one that matters most: it goes red the moment somebody reinstates a text fallback, which is exactly the mistake this area invites.

Verified rather than taken on trust:

- Your reading of the result event matches two runs I captured against the real CLI earlier. Success: `subtype: "success"`, `is_error: false`, `result: "OK"`. Auth failure: `subtype: "success"`, `is_error: true`, `result: "Failed to authenticate. API Error: 401 API key is invalid."` So `subtype` is useless as a discriminator and `is_error` is the whole signal, precisely as your code assumes.
- The claim in `PipelineQueue+Stages` that every other provider error is content free holds. `OpenAIProtocolGenerator` reads `errorBody` only on a non-2xx status, and it was already logged at `.public` there before this PR, so nothing is newly exposed.
- Gates on this head: lint clean across 553 files, full suite 3149 tests with no failures.

Three line comments, none of them blocking on their own. Plus one thing about the branch:

**It carries a merge commit** (`Merge branch 'main' into fix/protocol-generation-failure-reason`). I ran `scripts/ci/check-merge-commits.sh` against it locally and it reports exactly that. It will not block the merge, since that check is advisory and not a required context, but it does make the branch unreadable commit by commit, which is the reason the check exists. The "Update branch" button creates one by default; "Update with rebase" in its dropdown does not. Do not spend time on it if you would rather not rewrite history: I will fold and rebase before merging, same as last time.

pasrom (approved):
All three points are addressed, and you took the harder option on the one I explicitly said I would merge without.

The `terminal_reason` gate checks out against the two runs I captured from the real CLI: the failing one carries `"api_error"` and passes the gate, the successful one carries `"completed"` and would not, so the guard admits the case this feature exists for and falls closed on everything else. Both directions are tested, including the end to end placeholder path for an unrecognised value. And the invariant is now written down where the next person will actually meet it, which was the part I cared about most: `result` is dual purpose, `subtype` says `"success"` either way, and the content safety of the `.public` log rests on that.

`selectFailureMessage` and the caller-less overload are gone, so nothing is left alive for the tests alone.

I folded the branch into one commit on top of current `main`, which also drops the merge commit that `merge-commit-check` was reporting. One commit rather than three is deliberate: the two follow-ups corrected the same change, and the two deleted functions lost their last production caller to it, so they belong in the same tree. Your authorship is preserved. The fold was built from the tree, so it is bit for bit identical to the three commit version and `git diff` between them is empty.

Gates on the folded head before pushing: lint clean across 553 files, full suite 3147 tests with no failures, release parity passed. CI agrees, `merge-commit-check` included.

Thanks for turning this around so quickly, and for taking the belt and braces route.

Issue #672, opened 2026-09-03
No watchdog on a silent app track: exact zeros never trigger the tap rebuild


pasrom:
## Problem

Nothing in the capture path acts on the samples it records. The level is computed per buffer, and `ChannelHealthController` does notify the user after the configured debounce (default 90 s) when the app track goes silent while the mic still carries audio. But nothing rebuilds on it, and the notification text points at Screen Recording and third-party audio tools, which is the wrong hint for this failure.

The only signal that rebuilds the tap and aggregate is the default output device listener (`installOutputDeviceChangeListener` -> `handleOutputDeviceChanged`). A tap whose data path dies while its IOProc keeps firing therefore produces a full-length file of digital silence, with a notification that sends the user looking in the wrong place.

#671 is a field case: 1193 of 1194 seconds exact zeros, buffers delivered throughout, tap pointed at the right processes. Externally, Apple developer forum thread 825780 reports the same signature for a tap anchored the way ours is (private tap, aggregate main sub-device set to the default output UID): every buffer exactly `0.0f` while the IOProc runs, for 53 s to 16 minutes, during a video call. Full teardown and rebuild of tap plus aggregate is the only recovery reported there.

## Two parts, different confidence

**1. Instrumentation (unconditional).**

Alongside each 5 s RMS line and at `stop()`, log per tapped process object:

- `kAudioProcessPropertyIsRunningOutput`
- `kAudioProcessPropertyDevices`

Be honest about what these buy. `IsRunningOutput` says whether the process had an output stream running at all, which separates a dead tap from a process that rendered nothing. It does **not** separate a muted or silent far end from a dead tap, because a running stream rendering zeros still reports true. That is exactly the objection the forum reporter raised against this property, and it is why part 2 below cannot rely on it alone. `Devices` says whether the process output went to a separate aggregate (a voice-processing one, for instance) or to the output device directly, which is the distinction #671 turns on.

**2. Recovery (gated on the diagnosis in #671).**

When N seconds of *exact zeros* coincide with a tapped process reporting `IsRunningOutput`, run the restart path we already have: `applyAction(.stopAndRetry)` builds a fresh `AppTapSession` (tap, aggregate, IOProc), generation tagged, deadline bounded by `RestartArbiter` (5 s per attempt), budgeted by `CaptureRestartRetryPolicy` (5 attempts), and `TimelineAnchor` turns the gap into silence rather than drift. The machinery is all there; the trigger is what is missing.

Constraints:

[...]

Issue #671, opened 2026-09-03
Process tap records digital silence when Safari renders the far end through VoiceProcessingIO


duncanwood:
**Version:** 0.7.0 (Homebrew) · macOS 26.5.2 (25F84) · Safari 26.x · AirPods Pro as input and output

A 20-minute Microsoft Teams call in Safari recorded my mic fine and left the app track at digital silence. It looks like #524, but the PID fix from #579 is working here: the tap covered the right processes and captured the far end for a moment before going dead. I can't reproduce it outside a real call, so I'm filing what I have rather than a repro.

## The tap was right, and briefly worked

`_app16k_raw.tmp` is exact zeros for 1193 of its 1194 seconds. The only audio is 13.44s to 14.68s, and the transcript says what it was:

```
[00:13] Remote: Hi, Duncan.
```

That's the other participant, so the tap was pointed at a process that does produce the call audio.

```
10:41:46.449 App audio tap: 2 PID(s) [Safari(1017), com.apple.WebKit.GPU(3497)]
```

pid 3497 is the GPU helper macOS holds Safari responsible for. I checked by walking every CoreAudio process object and resolving `responsibility_get_pid_responsible_for_pid` on each: this machine had five `com.apple.WebKit.GPU` processes at the time, attributed to Safari, Raycast, Mail, Google Drive and a Safari web app. The responsibility lookup picked the Safari one.

## Where the audio went

```
10:41:56.537  WebKit.GPU[3497]  setPlayState Started  Input/Output {VPAUAggregateAudioDevice-0x92f054040, 0xa}
10:42:06.222  WebKit.GPU[3497]  setPlayState Started  Output {AC-07-75-E1-29-05:output, 0x14}
10:42:08.178  WebKit.GPU[3497]  setPlayState Stopped  Output {AC-07-75-E1-29-05:output, 0x14}
```

The app track's 13.44s–14.68s is 10:42:06.7 to 10:42:08.0 in wall clock, inside that one window. WebKit opened a VPAU aggregate device for the call ten seconds in, and from then on the sustained call audio ran through it. It also cycled nine short-lived plain output contexts on the AirPods device (0x13 through 0x1b) over the rest of the call, and the tap caught audio during exactly one of them.

## What I ruled out

[...]

Issue #699, opened 2026-09-09
[Bug] App track is digital silence when the meeting app renders to a device other than the system default output


soundman666:
### App version

0.8.0 (fcd8a78) · Homebrew

### Build variant

Homebrew (stable)

### macOS version

26.6.2 (25G83)

### What happened?

Microsoft Teams plays through a USB speakerphone (Jabra SPEAK 410, AudioObjectID 102) while the system default output is a different USB interface (ZOOM UAC-2, AudioObjectID 111). Two recordings (58 s test call and a 56 min meeting): `_app.wav` is exact zeros for the whole duration, `_mic.wav` is fine. Screen & System Audio Recording, System Audio Recording Only and Microphone are all granted; Detect Silent Capture Channel was off at the time.

Expected: the app track carries the Teams audio regardless of which output device Teams renders to, or at least the silent channel is reported at tap-creation time, since the real device is already known.

Cause, as far as I can read the code: `AppAudioCapture.swift` builds the aggregate from `getDefaultOutputDeviceUID()`, and `aggregateDescription()` uses that UID both as `kAudioAggregateDeviceMainSubDeviceKey` and as the only entry of `kAudioAggregateDeviceSubDeviceListKey`. A process rendering to another device is never followed. `kAudioProcessPropertyDevices` is read (#682) and logged as `outputDevices=[102]`, but the value is not used for the tap. Related: #671, #672, #693.

### Steps to reproduce

1. System Settings → Sound → Output: device A (here ZOOM UAC-2).
2. Teams → Settings → Devices → Speaker: a different device B (here Jabra SPEAK 410).
3. Record-only mode on, join a Teams call, talk for a minute, leave.
4. `ffmpeg -i <ts>_app.wav -af volumedetect -f null -` → mean/max −91 dB; raw samples are 0.

Not verified yet: whether making device B the system default restores the app track.

### Diagnostic log

Excerpt from `~/Library/Logs/MeetingTranscriber/diagnostics-2026-09-09.log` for the 56 min recording (verbose logging was off; full export available on request):

```
09:36:23 [com.meetingtranscriber.audiotap:AppAudioCapture] App audio tap: 5 PID(s) [Microsoft Teams WebView Helper(1345), Microsoft Teams WebView Helper(1344), Microsoft Teams ModuleHost(1282), com.microsoft.teams2.notificati(1092), Microsoft Teams WebView(109
[...]

PR #713, opened 2026-09-13
fix(app): recover silent app audio by trying another output device

### LINKED ISSUE #712: [Bug] App-audio tap follows the system default instead of the device Teams is playing through
### App version

Originally reproduced on a source build labeled 0.7.0, based on upstream commit `f3a7b7c` (August 24). Currently using a patched build labeled **0.7.0 (`83bf974`)**, built August 25.

A September 13 source comparison against **v0.8.1 (`672eb1c`)** and current `main` (`f17ae33`) shows that the device-selection/recovery mechanism described below is still absent. I have **not** replaced my patched installation with 0.8.1 to reproduce it again.

### Build variant

Built from source (Homebrew/non-App-Store variant).

### macOS version

Currently macOS **26.6.2**. The exact macOS version at the original August 24–25 incidents was not recorded in the investigation notes, so this is not a claim that those incidents occurred on 26.6.2.

### What happened?

The app-audio track lost the remote participants even though I could still hear them in Microsoft Teams. The tap included the correct Teams processes, including ModuleHost. The distinguishing condition was that **Teams rendered the call to a different output device from the macOS system default**, while the tap's private aggregate remained anchored to the system default.

Two observed forms:

- Teams changed the system default to its own virtual loopback device (`MSLoopbackDriverDevice_UID`) while continuing to render the call to AirPods. In one 68.7-minute recording, the app track contained only zeros across 45.5 minutes associated with that routing condition. The original unified-log investigation showed the default-device transitions and continued rendering to the headphones; the Teams helper was already in the tapped PID list.
- My normal desk default is a Rogue Amoeba Loopback device, **Studio Display Surround**. On August 25, Teams stopped output on that device and started output on AirPods shortly before recording began, without making the headphones the system default. The resulting 44.7-minute app track had only **1.9% nonzero samples**. Occasional nonzero buffers meant that a watchdog requiring an *uninterrupted* run of zeros never fired.

**Virtual transport alone is not a fault.** Earlier recordings with Teams actually rendering into Studio Display Surround captured correctly. A first attempted fix that preemptively avoided virtual devices regressed that working case and was replaced. Likewise, we have not established which Teams feature initiates its default-device changes; ordinary screen sharing is not a reliable trigger.

#### Expected behavior

Capture the remote audio when the meeting app uses a different output device from the system default, or detect the mismatch and attempt bounded recovery without disturbing correctly captured audio. A tap producing full-length, nearly empty audio files should not remain indefinitely attached to a device that carries none of the meeting.

#### Existing local fix and field experience

The revised fix is preserved on my fork in commits `45824f8` and `83bf974`, branch [`alexpfau-fix-virtual-output-device-tap-silence`](https://github.com/alexpfau/meeting-transcriber/tree/alexpfau-fix-virtual-output-device-tap-silence). **I have been using the patched build (`83bf974`) since August 25, 2026, with no recurrence of the reported problem and no other observed issues.** That is field-use experience, not a claim that every recovery path has been observed firing since installation.

Its approach is to:

- Always try the system default first, including valid virtual/aggregate defaults.
- Detect a likely dead anchor when at least **90% of individual samples are exactly zero over a rolling 120-second window**, rather than letting an occasional nonzero sample reset the clock.
- On that evidence, use the existing bounded restart machinery to try another output-device anchor. Rank fallback devices with running I/O first, remember anchors only after sustained substantive delivery, and bound/latch the search rather than retry forever.
- Report the condition through the notification, diagnostic, and automation paths.

The original investigation measured the maximum rolling 120-second zero fraction at **12.2–41.7% across 20 healthy recordings**, versus **98.7–100% across six failed recordings**. The 90% threshold sits between those observed groups. Healthy all-zero stretches reached 49.85 seconds, so aggressively shortening the window would risk interrupting valid recordings. These measurements were taken from the written 16 kHz files, not the raw buffers the watchdog evaluates; they are supporting evidence, not a proof that the signals are identical or the threshold generalizes to every setup.

#### Relationship to #672

Related to **#672**, but this report records a specific **wrong-device anchor** condition rather than assuming a tap's data path has died. Rebuilding against the same wrong output device can reproduce the failure; selecting a different anchor is the relevant additional behavior.

The instrumentation and notification improvements in #682/#681/#698 are valuable, but do not implement that device fallback. The Safari/VoiceProcessingIO investigation in #671 may have a different cause, and this proposal does **not** claim to resolve it. I understand the concerns in #672 about false positives and exposing capture to unnecessary restart failures; those are why the local fix keeps the working default first, uses a conservative measured window, and bounds recovery.

I am porting the existing fix to current upstream before opening a PR, preserving the newer capture lifecycle, sample-rate handling, and diagnostics rather than submitting the old branch unchanged.

### Steps to reproduce

The original failure was observed in real calls, not a fully deterministic automated reproduction. The relevant topology is:

1. Use a virtual device such as Loopback's Studio Display Surround as the system default output.
2. Have the meeting app render its remote audio to a different device, such as AirPods, using its own output selection. Confirm that the system default remains unchanged.
3. Record that app and keep remote audio playing long enough to distinguish a broken capture from a genuinely quiet far end.
4. Compare the app track with the sound actually heard, and inspect the selected output-device anchor. The observed failure is a nearly/all-zero app track while the meeting remains audible through the other device.

A second field-observed route into the same mismatch was Teams making its loopback device the system default mid-call while continuing to play through the headphones. The initiating Teams feature remains unconfirmed.

A necessary regression control is the opposite routing: keep the virtual default and have the meeting app actually play into it. Capture should keep that working anchor rather than reject it based on transport type.

### Diagnostic log

No raw meeting recordings, transcripts, or full private logs are attached. The relevant measurements and routing observations from the original investigation are summarized above without meeting names, participant identities, or hardware addresses. The fork commits preserve the implementation rationale and regression cases.

alexpfau:
## What changed and why

Fixes #712: a meeting app can play through a different output device from the system default while its process tap keeps delivering mostly zeros. Keep the working default first, then attempt bounded recovery at another device when a conservative sample-based detector fires.

Related to #672, but this addresses a reproduced **wrong-device anchor** condition, not an assumption that rebuilding the same tap will fix every silent track. It does not claim to resolve the separate Safari/VoiceProcessingIO investigation in #671.

### Recovery behavior

- Always try the system default first, including virtual devices such as Loopback that legitimately carry the meeting. The earlier transport-based approach was rejected because it broke that working case.
- Count **individual exactly-zero samples** over a rolling **120-second window**, triggering at **90%**. Occasional nonzero samples defeated the old uninterrupted-silence criterion. The original measurements put healthy recordings at 12.2–41.7% and failed ones at 98.7–100%; these were measured after 16 kHz conversion, not on the raw buffers, so they are supporting evidence rather than a universal guarantee.
- Prioritize fallback devices with running I/O while preserving the known-good/built-in/remaining-device ordering within each group. Running I/O is not proof that the meeting app plays there. Credit an anchor as known-good only after a five-second interval with at most 50% zeros.
- Reuse the existing restart coordinator, deadline, retry policy, and timeline-gap handling. Allow **one move per continuous silence episode and at most three per recording**. An unsuccessful fallback stays latched rather than repeatedly rebuilding; a genuine default-device change resets the search to the default.
- Publish the selected anchor only when its session is adopted, and generation-check queued silence verdicts so an old tap cannot restart its replacement.
- Feed the rolling verdict into the current fault-notification path without requiring microphone speech or adding another debounce. Preserve notification deduplication, no-buffer precedence, and terminal give-up escalation. Add `appDigitalSilence` to `/state.channelHealth` without replacing its existing fault, level, or age fields.

This is a port of the field-used commits `45824f8` and `83bf974` onto current upstream `f17ae33`, not a wholesale replacement of the capture code. The newer sample-rate handling and diagnostic probes remain intact. The product does not change the system output device or broaden capture to other processes.

### Field use versus verification of this port

I have been using the original patched build (`83bf974`) since **August 25, 2026**, with no recurrence of the reported problem and no other observed issues. That is field-use experience, not a claim that every recovery path was observed firing throughout that period.

On **September 13**, I installed this port (`e407f7d`) and ran two controlled synthetic-audio checks on macOS **26.6.2**, with microphone capture disabled:

| Check | Observed result |
|---|---|
| System default and producer both use **Studio Display Surround** (Loopback) | **152.82 seconds** of captured audio, no fallback and no false silence verdict. First 90 seconds: 0.62% zeros; final 20 seconds: -54.88 dBFS RMS. |
| System default stays on Loopback; producer plays explicitly to **MacBook Pro Speakers** | First 90 seconds were **100% zeros**. The watchdog fired after **120.06 seconds**, selected candidate 1 (built-in fallback), restarted, and reported signal again at approximately **120.70 seconds**. The 179.47-second recording's final 20 seconds measured **-42.84 dBFS RMS**. |

Sanitized sequence from the second run, relative to the initial anchor selection:

```text
+  0.000s  candidate 0 of 3 (systemDefault)
+120.058s  at least 90% of the last 120 seconds was digital silence
+120.593s  candidate 1 of 3 (builtInFallback)
+120.620s  tap restarted (48000 Hz)
+120.699s  tap is delivering signal again
```

The finished WAVs were measured, not just the RPC flags. Sidecars confirmed there were **no microphone tracks**. Original app preferences and the MacBook speaker output were restored afterward. AirPods were not connected for this new run, so the live evidence above is for the built-in fallback, not a new AirPods verification.

The test harness initially used an externally generated output-folder bookmark that macOS refused; the app's existing Downloads fallback retained the control recording. That synthetic output was identified and measured, and the second check used the default output folder explicitly. No private recordings, transcripts, or full device-identity logs are attached.

### Validation and limits

Passed locally:

- AudioTapLib: `swift test --parallel` (417 scheduled tests).
- App: `swift test --parallel --skip MenuBarIconSnapshotTests` (3,159 scheduled tests; model-dependent/E2E cases retain their normal opt-in gates). Only the documented developer-only icon snapshot class was explicitly excluded.
- Targeted app reporting, escalation, recorder-lifecycle, and RPC coverage (106 tests).
- Targeted AudioTapLib recovery/session-ownership/wedge coverage under ThreadSanitizer (14 tests).
- Checksum-pinned `./scripts/lint.sh` and `./scripts/pre-push.sh --with-appstore`.
- Regression check: substituting the old whole-buffer zero inference makes the raw-buffer recovery integration test fail seven assertions; restoring per-sample counting makes it pass.

A sufficiently long legitimately silent far end can still trigger the heuristic. The window, per-episode latch, and recording-wide budget bound that exposure; the change does not establish a universal distinction between silence and a broken capture path. Unit coverage also pins the exact 90% boundary, healthy-corpus ceiling, no-fallback behavior, sustained silence after an unsuccessful fallback, stop-before-dispatch, and stale verdicts after device replacement.

## Checklist

- [x] Rebased on `main`, no merge commits, since a merge commit pulls commits that are
      already on `main` into this branch and the review reads it commit by commit:
      `git fetch https://github.com/pasrom/meeting-transcriber main && git rebase FETCH_HEAD`
      (the URL spelled out because in a fork clone `origin` is your fork). Base is `f17ae33`; one logical commit ahead.
- [x] `./scripts/lint.sh` is clean
- [x] Tests pass: `cd app/MeetingTranscriber && swift test` — run with `--parallel --skip MenuBarIconSnapshotTests` as detailed above.

The **Update branch** button merges by default; pick **Update with rebase** from its
dropdown instead. Would rather not rewrite history? Leave *Allow edits by maintainers*
checked and a maintainer folds the commits before merging, as
[CONTRIBUTING.md](https://github.com/pasrom/meeting-transcriber/blob/main/CONTRIBUTING.md)
offers.


tools/audiotap/Sources/SilentTapWatchdog.swift
```diff
@@ -0,0 +1,178 @@
...
+    ///   - now: monotonic seconds. Injected rather than read here so the whole
+    ///     transition table is testable without waiting out a real window.
+    mutating func observe(
+        zeroSamples: Int, samples: Int, now: TimeInterval,
+    ) -> Action? {
+        guard samples > 0 else { return nil }
+        window.add(zeroSamples: zeroSamples, totalSamples: samples, now: now)
+        guard let fraction = window.zeroFraction else { return nil }
+
+        if isSilent {
```
pasrom: This is the blocker from the summary, at its source.

While `isSilent` is true, `observe` can only return `.recovered` or `nil`. The single clear is on line 147, behind `fraction <= recoveryFractionThreshold`. `resetWindow()` on line 175 keeps the latch by design, and `AnchorSearch.advance` has exactly one caller, reachable only from `.silenceDetected(mayRestart: true)`. So a fallback that is also dead ends the search for the episode.

Three things in this PR describe the behaviour that would result if it did not:

- `defaultMaxTriggers = 3` and the PR body's "at most three per recording": within one episode the budget cannot be spent at all.
- `OutputDeviceAnchorPolicy`'s "what `SilentTapWatchdog` walks, one step per trip".
- The "all `N` candidate anchors have delivered only digital silence" log line: for `N` greater than 1 it can only be reached after the fallback delivered real audio and later went quiet, at which point the message is not true.

The comment above ("a restart that did not fix the problem is not followed by an identical restart") is right about not re-trying the *same* anchor, but the code also prevents trying the *next* one.

Worth noting separately: the recovery branch has no span requirement, so immediately after `resetWindow()` a single buffer at 50% zeros or less clears the latch. The claim just above it that clearing "takes sustained audio rather than one loud buffer" only holds once the window has refilled.

tools/audiotap/Sources/AppAudioCapture+Restart.swift
```diff
@@ -19,6 +19,13 @@ extension AppAudioCapture {
         let action = deviceChangeCoordinator.handle(.deviceChanged)
         guard action != .ignore else { return }
 
+        // A new default output is a new situation, so the anchor search starts
+        // over rather than staying parked on whatever fallback a previous
+        // stretch escaped to. Load-bearing for the case the search exists for:
+        // when the meeting app hands the default back to real hardware, the tap
+        // has to follow it there immediately.
+        anchorSearch.withLock { $0.resetToDefault() }
```
pasrom: This is the only path that ever returns the cursor to 0, and it fires only on a real `kAudioHardwarePropertyDefaultOutputDevice` change.

Two consequences:

1. **No way back after a false positive.** If a healthy default is abandoned because the far end was legitimately silent for a full window, the tap stays on the fallback for the rest of the recording. `lastKnownGoodUID` cannot rescue it, since credit requires a 5 s interval at 50% zeros or less, which a zero-delivering anchor never produces. The user is not told to change the system output device either, because the message the rolling path selects is not the one that carries that advice.

2. **The latch and the budget are not reset here.** Your own report describes Teams re-pointing the default at its loopback driver four times in one meeting. Each of those periods spends one trigger, so the fourth is stranded with `mayRestart: false`. And if the previous fallback was dead, the latch is still set when the new default arrives, so the search is disarmed on a default it has never evaluated.

Resetting the watchdog's latch (and considering whether the budget should be per episode rather than per recording) alongside the cursor would address both.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
```diff
@@ -0,0 +1,202 @@
...
+        everCarriedSignal: Bool,
+        rollingDigitalSilence: Bool = false,
+    ) -> (title: String, body: String, urgency: NotificationUrgency) {
+        if fault == .gaveUp {
+            return gaveUpAlert(channel: channel)
+        }
+        let suppressible = fault == .digitalSilence && channel == .mic
+        return (
+            "Capture Channel Silent",
+            rollingDigitalSilence && channel == .app && fault == .digitalSilence
```
pasrom: Two things here.

**`everCarriedSignal` is discarded whenever the rolling verdict is set.** That makes the `(.app, .digitalSilence, false)` arm of `faultMessage` unreachable through this path. The case it costs: a tap denied the audio-capture grant delivers zeros from the first buffer, and if the user listens for more than two minutes before speaking, the rolling verdict claims the once-per-recording slot at t=120 with output-device advice, and the later, correctly corroborated age-based verdict is discarded. Previously that user got the Screen Recording and interception advice at first corroboration. The same happens permanently for anyone who sets `asymmetricSilenceWarningSeconds` above 120.

Note it is not a one-line pass-through: the `false` arm's text asserts "while the microphone carries audio", which is not true for the population the rolling path reaches, so that copy needs adjusting too.

**The urgency.** `.timeSensitive` for app-side digital silence is justified in the comment above by "the far end of a call does not go digitally silent while you are talking", which depends on microphone corroboration. The rolling path explicitly removes that corroboration. So the Focus-piercing alert now also fires for a joined-but-silent meeting, which is exactly the shape `applySilentRecording` calls out a few lines away when it chooses `.standard` because "a waiting room looks exactly like this". By this file's own test ("does this have a benign reading"), the uncorroborated rolling verdict looks like a `.standard` case.

app/MeetingTranscriber/Sources/ChannelHealthController.swift
```diff
@@ -302,18 +325,37 @@ final class ChannelHealthController {
...
+    /// purely for length; it has no callers of its own.
+    private func applySilentRecording(
+        _ silentEvent: SilentRecordingEvent?, mic: Double, app: Double,
+    ) {
         switch silentEvent {
         case .started:
             recordingSilentActive = true
+            // The app fault already explains this silent recording. Keep the
+            // level-driven tint, without a second notification about it.
+            guard !appDigitalSilenceActive else { return }
```
pasrom: This guard suppresses the duplicate in only one order, and the default settings produce the other one.

`SilentRecordingMonitor`'s debounce is `asymmetricSilenceWarningSeconds`, default 90. The rolling window is 120. So at t=90 this method runs with `appDigitalSilenceActive` still false and posts "Recording Appears Silent"; at t=120 the rolling verdict posts "Capture Channel Silent" as time-sensitive. `NotificationManager.notify` uses a fresh UUID per call with no title or category de-duplication, and `ChannelFaultMonitor.reportedSilence` is not set for an uncorroborated verdict, so nothing else blocks the second one.

The two monitors are not different populations here: a listening user with a dead tap satisfies both, and the rolling verdict's stated target is precisely "a listening user may never speak".

`testDigitalSilenceIsReportedOncePerEpisode` sets the flag before tick 0, so it exercises only the order that works. A test that flips the flag *after* the debounce would show this.

The guard is also one-directional: the symmetric monitor defers to the rolling flag, but the rolling report never consults `recordingSilentActive`.

tools/audiotap/Sources/AppAudioCapture+Anchor.swift
```diff
@@ -0,0 +1,189 @@
...
+    /// could not name a winner. Logging it on every attempt is what will decide
+    /// whether it is ever trustworthy enough to lead. Device UIDs are already
+    /// logged on the line above, so this adds no new identifying data.
+    private func logRunningIO(_ devices: [OutputDeviceAnchorPolicy.Device]) {
+        let running = devices.filter(\.isRunningIO).map(\.uid).joined(separator: ", ")
+        guard !running.isEmpty else {
+            logger.info("App audio: no output device reports running IO")
+            return
+        }
+        logger.info("App audio: output devices running IO: [\(running, privacy: .public)]")
```
pasrom: This is the only unconditional `.public` device UID in the module, and I do not think it should be.

The justification in the comment above, that the UIDs "are already logged on the line above", does not hold: line 52 interpolates `currentDefault.uid` with no privacy annotation, and os.Logger redacts non-scalar interpolations by default, so that one is `<private>`. Every other UID log in this module is either `[debug]`-gated or un-annotated.

It matters because `PersistentDiagnosticLog` streams `subsystem CONTAINS 'com.meetingtranscriber'` at info level into `~/Library/Logs/MeetingTranscriber/`, which is the file users export and attach to issues. Bluetooth device UIDs embed a MAC address and USB interface UIDs frequently embed a serial number.

It also repeats on every restart attempt, since `resolveOutputAnchor` runs per attempt.

Gating it behind `debugLogging`, or dropping the annotation so it is redacted like line 52, would keep the diagnostic value without that.

tools/audiotap/Sources/RollingZeroFraction.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        // A bin is kept while any part of it can still fall inside the window.
+        let oldestKept = Int(((now - windowSeconds) / binSeconds).rounded(.down))
+        bins.removeAll { $0.index < oldestKept }
+    }
+
+    /// How long has been observed since accumulation began, capped at the
+    /// window. `nil` before anything has been observed at all.
+    func span(now: TimeInterval) -> TimeInterval? {
+        firstObservation.map { min(now - $0, windowSeconds) }
```
pasrom: `firstObservation` is set once and cleared only by `reset()`, while bin eviction is by wall-clock index. So `span` measures time since accumulation began, not the time actually covered by the retained bins, and the two diverge whenever buffers stop arriving without the tap being torn down.

After any gap of 121 s or more, every pre-gap bin is evicted, yet `span` still reports the full window and `SilentTapWatchdog` accepts it as complete. The sample floor does not catch it, because it is computed against the same inflated `span`: 120 x 100 = 12,000 samples is 0.125 s at 48 kHz stereo. So a verdict advertised as covering two minutes can rest on a fraction of a second.

I could not show a field path that reaches it. Sleep is ruled out, since `mach_absolute_time` does not advance while the machine is asleep. It needs an IOProc that stalls in place for minutes and then resumes, which I have not observed here. So this is a latent contract violation rather than a demonstrated bug, and the comment on lines 46 to 48 states the contract it breaks.

The fix is small: derive the span from the oldest retained bin, or require the retained sample count to reflect a realistic rate rather than 100/s.

tools/audiotap/Sources/AnchorSearch.swift
```diff
@@ -0,0 +1,74 @@
...
+        cursor = 0
+    }
+
+    /// Record that audio genuinely arrived at `uid`.
+    ///
+    /// Idempotent per anchor: the caller invokes this from the buffer path, so
+    /// it must stay cheap and must not churn on every buffer of a healthy
+    /// recording.
+    mutating func recordDelivery(at uid: String) {
+        guard lastKnownGoodUID != uid else { return }
```
pasrom: `recordDelivery` overwrites, which means credit can erase a correct memory as well as record one.

The reasoning in `AnchorDeliveryCredit` for the 50% threshold computes roughly 82% zeros, but that is the mean across the 55 energetic windows of that recording, whereas the code tests a per-interval maximum. A mean of 18.5% non-zero is consistent with a distribution that includes one 5 s interval well above 50% non-zero. Around 51 s of non-zero samples exist in that 44.7-minute track, so it takes 2.5 s of it landing in one interval. A ringtone or a lobby loop rendered to the anchored device has that shape; a one-second join chime does not.

The direct consequence I first assumed does not follow, because a last-known-good device that is still the current default is deduplicated out of the candidate list. What does follow is this line: a burst on the dead anchor overwrites the UID of hardware that genuinely carried the meeting earlier, and the search loses that memory for the rest of the recording.

Requiring a bound rather than an average, for example several consecutive qualifying intervals before overwriting an existing known-good UID, would close it.

tools/audiotap/Tests/AppAudioCaptureAnchorRecoveryTests.swift
```diff
@@ -0,0 +1,180 @@
...
+        let capture = makeCapture(attempts)
+        try capture.start()
+        defer { capture.stop() }
+
+        feed(capture, zeros: 980, from: 0, seconds: 1.2)
+        awaitMainQueue()
+        feed(capture, zeros: 980, from: 2, seconds: 10)
+        awaitMainQueue()
+
+        XCTAssertEqual(attempts.count.withLock { $0 }, 2)
```
pasrom: Two notes on this file, since it is the only test that drives the real object.

**It pins the one-step behaviour.** This assertion says that after continued silence at the fallback the attempt count stays at 2. That is the current design, but it is also the behaviour I have flagged as the blocker, so if the search is changed to continue, this test changes with it.

**The cursor-to-device link is not covered.** `performAttempt()` returns `attemptBody()` whenever the seam is set, which bypasses `startCapture()` and therefore `resolveOutputAnchor()` entirely, and `Attempts.run()` supplies the anchor UID itself. I changed `resolveOutputAnchor()` to ignore the cursor and always return candidate 0, and the whole suite stayed green. Nothing under `Tests/` references that function. Since choosing a different device is the entire point of the change, that seam is worth closing, for example by injecting the device enumeration so `resolveOutputAnchor` can be driven with a synthetic device list.

For completeness, these edits also leave the suite green: `resetWindow()` as a no-op, deleting the `resetToDefault()` call, `defaultMaxTriggers` 3 to 1, `defaultRecoveryFractionThreshold` 0.50 to 0.89, deleting the delivery-credit reset at teardown, and hard-coding `AudioCaptureSession.appCaptureDigitallySilent` to `false`. The last one would silently disable the whole app-side feature.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
```diff
@@ -0,0 +1,202 @@
...
+                + "intercepting it."
+
+        default:
+            "Both capture channels have been silent since the recording started. "
+                + "Check the audio routing — the meeting app may have claimed the mic "
+                + "in exclusive mode (e.g. AirPods HFP), or the system input device may be muted."
+        }
+    }
+}
+import AudioTapLib
```
pasrom: This trailing import is what makes `./scripts/lint.sh` fail on the branch: SwiftFormat reports `blankLinesBetweenScopes` at line 201.

The import itself is needed, since `SystemSettingsPaths` lives in AudioTapLib and there is no re-export. Note that `lint.sh --fix` only inserts a blank line and leaves the import at the bottom, so moving it to the top of the file is a manual edit.

tools/audiotap/Sources/OutputDeviceEnumeration.swift
```diff
@@ -0,0 +1,174 @@
...
+        let count = Int(size) / MemoryLayout<AudioObjectID>.size
+        var ids = [AudioObjectID](repeating: AudioObjectID(kAudioObjectUnknown), count: count)
+        let status = AudioObjectGetPropertyData(
+            AudioObjectID(kAudioObjectSystemObject), &address, 0, nil, &size, &ids,
+        )
+        guard status == noErr else {
+            logger.warning("Cannot read fallback devices (status: \(status, privacy: .public))")
+            return []
+        }
+        return ids
```
pasrom: Minor: `ids` is sized from the first call, and the second call can return a smaller `size` if a device disappears in between, leaving trailing `kAudioObjectUnknown` entries. They are harmless downstream, since `describeOutputDevice(0)` fails and is dropped by `compactMap`, but each one logs a "Cannot read output streams" warning.

```suggestion
        return Array(ids.prefix(Int(size) / MemoryLayout<AudioObjectID>.size))
```

tools/audiotap/Sources/SilentTapWatchdog.swift
```diff
@@ -0,0 +1,178 @@
...
+        self.maxTriggers = maxTriggers
+        self.minSamplesPerSecond = minSamplesPerSecond
+        window = RollingZeroFraction(windowSeconds: windowSeconds)
+    }
+
+    /// Feed one captured buffer.
+    ///
+    /// - Parameters:
+    ///   - zeroSamples: how many of that buffer's samples were exactly zero,
+    ///     counted by `exactlyZeroSampleCount` in the capture path's existing
```
pasrom: Small documentation mismatch: `exactlyZeroSampleCount` has no callers. The per-sample counting that actually feeds this type is written out inline in `accumulateDebugRMS`. Either call the function from there or drop it, so the two do not drift.

tools/audiotap/Sources/SilentTapWatchdog.swift
pasrom: This is the blocker from the summary, at its source.

While `isSilent` is true, `observe` can only return `.recovered` or `nil`. The single clear is on line 147, behind `fraction <= recoveryFractionThreshold`. `resetWindow()` on line 175 keeps the latch by design, and `AnchorSearch.advance` has exactly one caller, reachable only from `.silenceDetected(mayRestart: true)`. So a fallback that is also dead ends the search for the episode.

Three things in this PR describe the behaviour that would result if it did not:

- `defaultMaxTriggers = 3` and the PR body's "at most three per recording": within one episode the budget cannot be spent at all.
- `OutputDeviceAnchorPolicy`'s "what `SilentTapWatchdog` walks, one step per trip".
- The "all `N` candidate anchors have delivered only digital silence" log line: for `N` greater than 1 it can only be reached after the fallback delivered real audio and later went quiet, at which point the message is not true.

The comment above ("a restart that did not fix the problem is not followed by an identical restart") is right about not re-trying the *same* anchor, but the code also prevents trying the *next* one.

Worth noting separately: the recovery branch has no span requirement, so immediately after `resetWindow()` a single buffer at 50% zeros or less clears the latch. The claim just above it that clearing "takes sustained audio rather than one loud buffer" only holds once the window has refilled.

tools/audiotap/Sources/AppAudioCapture+Restart.swift
pasrom: This is the only path that ever returns the cursor to 0, and it fires only on a real `kAudioHardwarePropertyDefaultOutputDevice` change.

Two consequences:

1. **No way back after a false positive.** If a healthy default is abandoned because the far end was legitimately silent for a full window, the tap stays on the fallback for the rest of the recording. `lastKnownGoodUID` cannot rescue it, since credit requires a 5 s interval at 50% zeros or less, which a zero-delivering anchor never produces. The user is not told to change the system output device either, because the message the rolling path selects is not the one that carries that advice.

2. **The latch and the budget are not reset here.** Your own report describes Teams re-pointing the default at its loopback driver four times in one meeting. Each of those periods spends one trigger, so the fourth is stranded with `mayRestart: false`. And if the previous fallback was dead, the latch is still set when the new default arrives, so the search is disarmed on a default it has never evaluated.

Resetting the watchdog's latch (and considering whether the budget should be per episode rather than per recording) alongside the cursor would address both.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
pasrom: Two things here.

**`everCarriedSignal` is discarded whenever the rolling verdict is set.** That makes the `(.app, .digitalSilence, false)` arm of `faultMessage` unreachable through this path. The case it costs: a tap denied the audio-capture grant delivers zeros from the first buffer, and if the user listens for more than two minutes before speaking, the rolling verdict claims the once-per-recording slot at t=120 with output-device advice, and the later, correctly corroborated age-based verdict is discarded. Previously that user got the Screen Recording and interception advice at first corroboration. The same happens permanently for anyone who sets `asymmetricSilenceWarningSeconds` above 120.

Note it is not a one-line pass-through: the `false` arm's text asserts "while the microphone carries audio", which is not true for the population the rolling path reaches, so that copy needs adjusting too.

**The urgency.** `.timeSensitive` for app-side digital silence is justified in the comment above by "the far end of a call does not go digitally silent while you are talking", which depends on microphone corroboration. The rolling path explicitly removes that corroboration. So the Focus-piercing alert now also fires for a joined-but-silent meeting, which is exactly the shape `applySilentRecording` calls out a few lines away when it chooses `.standard` because "a waiting room looks exactly like this". By this file's own test ("does this have a benign reading"), the uncorroborated rolling verdict looks like a `.standard` case.

app/MeetingTranscriber/Sources/ChannelHealthController.swift
pasrom: This guard suppresses the duplicate in only one order, and the default settings produce the other one.

`SilentRecordingMonitor`'s debounce is `asymmetricSilenceWarningSeconds`, default 90. The rolling window is 120. So at t=90 this method runs with `appDigitalSilenceActive` still false and posts "Recording Appears Silent"; at t=120 the rolling verdict posts "Capture Channel Silent" as time-sensitive. `NotificationManager.notify` uses a fresh UUID per call with no title or category de-duplication, and `ChannelFaultMonitor.reportedSilence` is not set for an uncorroborated verdict, so nothing else blocks the second one.

The two monitors are not different populations here: a listening user with a dead tap satisfies both, and the rolling verdict's stated target is precisely "a listening user may never speak".

`testDigitalSilenceIsReportedOncePerEpisode` sets the flag before tick 0, so it exercises only the order that works. A test that flips the flag *after* the debounce would show this.

The guard is also one-directional: the symmetric monitor defers to the rolling flag, but the rolling report never consults `recordingSilentActive`.

tools/audiotap/Sources/AppAudioCapture+Anchor.swift
pasrom: This is the only unconditional `.public` device UID in the module, and I do not think it should be.

The justification in the comment above, that the UIDs "are already logged on the line above", does not hold: line 52 interpolates `currentDefault.uid` with no privacy annotation, and os.Logger redacts non-scalar interpolations by default, so that one is `<private>`. Every other UID log in this module is either `[debug]`-gated or un-annotated.

It matters because `PersistentDiagnosticLog` streams `subsystem CONTAINS 'com.meetingtranscriber'` at info level into `~/Library/Logs/MeetingTranscriber/`, which is the file users export and attach to issues. Bluetooth device UIDs embed a MAC address and USB interface UIDs frequently embed a serial number.

It also repeats on every restart attempt, since `resolveOutputAnchor` runs per attempt.

Gating it behind `debugLogging`, or dropping the annotation so it is redacted like line 52, would keep the diagnostic value without that.

tools/audiotap/Sources/RollingZeroFraction.swift
pasrom: `firstObservation` is set once and cleared only by `reset()`, while bin eviction is by wall-clock index. So `span` measures time since accumulation began, not the time actually covered by the retained bins, and the two diverge whenever buffers stop arriving without the tap being torn down.

After any gap of 121 s or more, every pre-gap bin is evicted, yet `span` still reports the full window and `SilentTapWatchdog` accepts it as complete. The sample floor does not catch it, because it is computed against the same inflated `span`: 120 x 100 = 12,000 samples is 0.125 s at 48 kHz stereo. So a verdict advertised as covering two minutes can rest on a fraction of a second.

I could not show a field path that reaches it. Sleep is ruled out, since `mach_absolute_time` does not advance while the machine is asleep. It needs an IOProc that stalls in place for minutes and then resumes, which I have not observed here. So this is a latent contract violation rather than a demonstrated bug, and the comment on lines 46 to 48 states the contract it breaks.

The fix is small: derive the span from the oldest retained bin, or require the retained sample count to reflect a realistic rate rather than 100/s.

tools/audiotap/Sources/AnchorSearch.swift
pasrom: `recordDelivery` overwrites, which means credit can erase a correct memory as well as record one.

The reasoning in `AnchorDeliveryCredit` for the 50% threshold computes roughly 82% zeros, but that is the mean across the 55 energetic windows of that recording, whereas the code tests a per-interval maximum. A mean of 18.5% non-zero is consistent with a distribution that includes one 5 s interval well above 50% non-zero. Around 51 s of non-zero samples exist in that 44.7-minute track, so it takes 2.5 s of it landing in one interval. A ringtone or a lobby loop rendered to the anchored device has that shape; a one-second join chime does not.

The direct consequence I first assumed does not follow, because a last-known-good device that is still the current default is deduplicated out of the candidate list. What does follow is this line: a burst on the dead anchor overwrites the UID of hardware that genuinely carried the meeting earlier, and the search loses that memory for the rest of the recording.

Requiring a bound rather than an average, for example several consecutive qualifying intervals before overwriting an existing known-good UID, would close it.

tools/audiotap/Tests/AppAudioCaptureAnchorRecoveryTests.swift
pasrom: Two notes on this file, since it is the only test that drives the real object.

**It pins the one-step behaviour.** This assertion says that after continued silence at the fallback the attempt count stays at 2. That is the current design, but it is also the behaviour I have flagged as the blocker, so if the search is changed to continue, this test changes with it.

**The cursor-to-device link is not covered.** `performAttempt()` returns `attemptBody()` whenever the seam is set, which bypasses `startCapture()` and therefore `resolveOutputAnchor()` entirely, and `Attempts.run()` supplies the anchor UID itself. I changed `resolveOutputAnchor()` to ignore the cursor and always return candidate 0, and the whole suite stayed green. Nothing under `Tests/` references that function. Since choosing a different device is the entire point of the change, that seam is worth closing, for example by injecting the device enumeration so `resolveOutputAnchor` can be driven with a synthetic device list.

For completeness, these edits also leave the suite green: `resetWindow()` as a no-op, deleting the `resetToDefault()` call, `defaultMaxTriggers` 3 to 1, `defaultRecoveryFractionThreshold` 0.50 to 0.89, deleting the delivery-credit reset at teardown, and hard-coding `AudioCaptureSession.appCaptureDigitallySilent` to `false`. The last one would silently disable the whole app-side feature.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
pasrom: This trailing import is what makes `./scripts/lint.sh` fail on the branch: SwiftFormat reports `blankLinesBetweenScopes` at line 201.

The import itself is needed, since `SystemSettingsPaths` lives in AudioTapLib and there is no re-export. Note that `lint.sh --fix` only inserts a blank line and leaves the import at the bottom, so moving it to the top of the file is a manual edit.

tools/audiotap/Sources/OutputDeviceEnumeration.swift
pasrom: Minor: `ids` is sized from the first call, and the second call can return a smaller `size` if a device disappears in between, leaving trailing `kAudioObjectUnknown` entries. They are harmless downstream, since `describeOutputDevice(0)` fails and is dropped by `compactMap`, but each one logs a "Cannot read output streams" warning.

```suggestion
        return Array(ids.prefix(Int(size) / MemoryLayout<AudioObjectID>.size))
```

tools/audiotap/Sources/SilentTapWatchdog.swift
pasrom: This comment says the zeros are "counted by `exactlyZeroSampleCount`", but that function has no callers: the per-sample counting that actually feeds this type is written out inline in `accumulateDebugRMS`.

`analyze` catches it, along with two more from this change. All three fail the gate as `unused_declaration`:

- `exactlyZeroSampleCount` in `RollingZeroFraction.swift`, never called.
- `restartsRequested` just below in this file, referenced only from tests.
- `getDefaultOutputDeviceTransportType()` in `Helpers.swift`, whose production caller this change replaces with `OutputDeviceEnumeration.transportLabel`, leaving it referenced only from tests and from a doc comment.

Calling the function from `accumulateDebugRMS` would fix the first one and make the comment true at the same time.

tools/audiotap/Sources/OutputDeviceAnchorPolicy.swift
```diff
@@ -0,0 +1,183 @@
+import CoreAudio
+import Foundation
+
+/// The ordered list of devices the process tap's private aggregate may be
+/// anchored to, best guess first.
+///
+/// The aggregate needs one real sub-device to clock against, and the tap
+/// carries what the tapped process renders *to that device*. Pick the wrong
```
pasrom: This sentence is the premise the whole mechanism rests on, and I could not reproduce it.

With a probe built to the same shape as production (private `stereoMixdownOfProcesses` tap, private aggregate, `tapAutoStart` true, anchor as main sub-device), on macOS 26.6.2 build 25G83, a process rendering to the system default was captured at full level with the aggregate anchored to a *different* device: -44.08 dBFS at the producer's own device and -44.08 dBFS at an unrelated USB interface. Anchoring to three different virtual devices also captured the audio; those rows show a non-zero "zero fraction" only because the aggregate adds the anchor device's own output channels alongside the tap's two, and the level drop matches the channel count to 0.01 dB.

I repeated the set with a producer rendering through VoiceProcessingIO, which is the path a meeting app uses for call audio, in case that behaved differently. It did not: -42.68 dBFS at both the producer's device and the unrelated USB anchor.

That is the opposite of "the tap carries what the tapped process renders to that device", and it also matches an earlier probe on the maintainer side that closed the same hypothesis. Your check 2 is the same shape as the virtual-anchor case and measured 100% zeros, so something differs between our setups that neither of us has isolated. Knowing the failing device's transport and channel count, and whether it was the system *default* as well as the anchor in the failing recordings, would help narrow it.

I am not asking you to take my measurement over yours. I am saying the causal claim is not established well enough to build the recovery on it, because if the anchor does not gate capture then moving the anchor is not what fixed your recordings, and the restart alone might be. A run of your check 2 with the cursor pinned at candidate 0 would separate those two directly.

tools/audiotap/Sources/OutputDeviceAnchorPolicy.swift
```diff
@@ -0,0 +1,183 @@
...
+
+        if let lastKnownGoodUID, outputDevices.contains(where: { $0.uid == lastKnownGoodUID }) {
+            append(lastKnownGoodUID, .lastKnownGood)
+        }
+
+        let physical = outputDevices.filter { !isInterposed(transportType: $0.transportType) }
+        if let builtIn = physical.first(where: { $0.transportType == kAudioDeviceTransportTypeBuiltIn }) {
+            append(builtIn.uid, .builtInFallback)
+        }
+        for device in physical {
```
pasrom: Every physical device with output streams becomes a candidate here, including ones that cannot currently start.

Measured on the same probe: anchoring the aggregate to an HDMI display whose panel was asleep produced **zero buffers**, twice, rather than buffers of zeros. That matters specifically for this design, because `SilentTapWatchdog` needs samples to form a verdict. With no samples it can never trip again, the latch from the previous episode stays set, and the tap is parked on a dead anchor for the rest of the recording, with only the separate `noBuffers` notification to show for it.

An HDMI display, a powered-down interface or a disconnected dock all enumerate with output streams while being unable to clock an aggregate, so this is not a corner case. Two things would help: treat a candidate that delivers no buffers within a short deadline as a failed candidate and advance, and prefer candidates that report running I/O more strongly than the current stable partition does.

tools/audiotap/Sources/SilentTapWatchdog.swift
pasrom: This is the blocker from the summary, at its source.

While `isSilent` is true, `observe` can only return `.recovered` or `nil`. The single clear is on line 147, behind `fraction <= recoveryFractionThreshold`. `resetWindow()` on line 175 keeps the latch by design, and `AnchorSearch.advance` has exactly one caller, reachable only from `.silenceDetected(mayRestart: true)`. So a fallback that is also dead ends the search for the episode.

Three things in this PR describe the behaviour that would result if it did not:

- `defaultMaxTriggers = 3` and the PR body's "at most three per recording": within one episode the budget cannot be spent at all.
- `OutputDeviceAnchorPolicy`'s "what `SilentTapWatchdog` walks, one step per trip".
- The "all `N` candidate anchors have delivered only digital silence" log line: for `N` greater than 1 it can only be reached after the fallback delivered real audio and later went quiet, at which point the message is not true.

The comment above ("a restart that did not fix the problem is not followed by an identical restart") is right about not re-trying the *same* anchor, but the code also prevents trying the *next* one.

Worth noting separately: the recovery branch has no span requirement, so immediately after `resetWindow()` a single buffer at 50% zeros or less clears the latch. The claim just above it that clearing "takes sustained audio rather than one loud buffer" only holds once the window has refilled.

tools/audiotap/Sources/AppAudioCapture+Restart.swift
pasrom: This is the only path that ever returns the cursor to 0, and it fires only on a real `kAudioHardwarePropertyDefaultOutputDevice` change.

Two consequences:

1. **No way back after a false positive.** If a healthy default is abandoned because the far end was legitimately silent for a full window, the tap stays on the fallback for the rest of the recording. `lastKnownGoodUID` cannot rescue it, since credit requires a 5 s interval at 50% zeros or less, which a zero-delivering anchor never produces. The user is not told to change the system output device either, because the message the rolling path selects is not the one that carries that advice.

2. **The latch and the budget are not reset here.** Your own report describes Teams re-pointing the default at its loopback driver four times in one meeting. Each of those periods spends one trigger, so the fourth is stranded with `mayRestart: false`. And if the previous fallback was dead, the latch is still set when the new default arrives, so the search is disarmed on a default it has never evaluated.

Resetting the watchdog's latch (and considering whether the budget should be per episode rather than per recording) alongside the cursor would address both.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
pasrom: Two things here.

**`everCarriedSignal` is discarded whenever the rolling verdict is set.** That makes the `(.app, .digitalSilence, false)` arm of `faultMessage` unreachable through this path. The case it costs: a tap denied the audio-capture grant delivers zeros from the first buffer, and if the user listens for more than two minutes before speaking, the rolling verdict claims the once-per-recording slot at t=120 with output-device advice, and the later, correctly corroborated age-based verdict is discarded. Previously that user got the Screen Recording and interception advice at first corroboration. The same happens permanently for anyone who sets `asymmetricSilenceWarningSeconds` above 120.

Note it is not a one-line pass-through: the `false` arm's text asserts "while the microphone carries audio", which is not true for the population the rolling path reaches, so that copy needs adjusting too.

**The urgency.** `.timeSensitive` for app-side digital silence is justified in the comment above by "the far end of a call does not go digitally silent while you are talking", which depends on microphone corroboration. The rolling path explicitly removes that corroboration. So the Focus-piercing alert now also fires for a joined-but-silent meeting, which is exactly the shape `applySilentRecording` calls out a few lines away when it chooses `.standard` because "a waiting room looks exactly like this". By this file's own test ("does this have a benign reading"), the uncorroborated rolling verdict looks like a `.standard` case.

app/MeetingTranscriber/Sources/ChannelHealthController.swift
pasrom: This guard suppresses the duplicate in only one order, and the default settings produce the other one.

`SilentRecordingMonitor`'s debounce is `asymmetricSilenceWarningSeconds`, default 90. The rolling window is 120. So at t=90 this method runs with `appDigitalSilenceActive` still false and posts "Recording Appears Silent"; at t=120 the rolling verdict posts "Capture Channel Silent" as time-sensitive. `NotificationManager.notify` uses a fresh UUID per call with no title or category de-duplication, and `ChannelFaultMonitor.reportedSilence` is not set for an uncorroborated verdict, so nothing else blocks the second one.

The two monitors are not different populations here: a listening user with a dead tap satisfies both, and the rolling verdict's stated target is precisely "a listening user may never speak".

`testDigitalSilenceIsReportedOncePerEpisode` sets the flag before tick 0, so it exercises only the order that works. A test that flips the flag *after* the debounce would show this.

The guard is also one-directional: the symmetric monitor defers to the rolling flag, but the rolling report never consults `recordingSilentActive`.

tools/audiotap/Sources/AppAudioCapture+Anchor.swift
pasrom: This is the only unconditional `.public` device UID in the module, and the file argues against it three lines above this helper.

The comment at line 46 says transport is logged unconditionally because it "is not personal data, unlike the device name/UID", inherited verbatim from the base. Line 52 follows that rule: it interpolates `currentDefault.uid` with no privacy annotation, so os.Logger redacts it. The three other UID logs in this module are all behind `[debug]` (`AppAudioCapture+Anchor.swift:60`, `AppAudioCapture+Restart.swift:34`, `MicCaptureHandler.swift:222`). This line is the one exception, and it runs with `debugLogging` off.

So the justification at lines 95 to 96, that the UIDs "are already logged on the line above", does not hold: the line above is redacted precisely because the file classifies UIDs as personal data.

It matters because `PersistentDiagnosticLog` streams `subsystem CONTAINS 'com.meetingtranscriber'` at info level into `~/Library/Logs/MeetingTranscriber/`, which is the file users export and attach to issues. It also repeats on every restart attempt, since `resolveOutputAnchor` runs per attempt, so a failing candidate writes it up to six times.

Either gate it behind `debugLogging` like the other three, or drop the annotation so it is redacted like line 52. The diagnostic value of knowing *how many* devices were running I/O survives either way.

tools/audiotap/Sources/RollingZeroFraction.swift
pasrom: `firstObservation` is set once and cleared only by `reset()`, while bin eviction is by wall-clock index. So `span` measures time since accumulation began, not the time actually covered by the retained bins, and the two diverge whenever buffers stop arriving without the tap being torn down.

After any gap of 121 s or more, every pre-gap bin is evicted, yet `span` still reports the full window and `SilentTapWatchdog` accepts it as complete. The sample floor does not catch it, because it is computed against the same inflated `span`: 120 x 100 = 12,000 samples is 0.125 s at 48 kHz stereo. So a verdict advertised as covering two minutes can rest on a fraction of a second.

I could not show a field path that reaches it. Sleep is ruled out, since `mach_absolute_time` does not advance while the machine is asleep. It needs an IOProc that stalls in place for minutes and then resumes, which I have not observed here. So this is a latent contract violation rather than a demonstrated bug, and the comment on lines 46 to 48 states the contract it breaks.

The fix is small: derive the span from the oldest retained bin, or require the retained sample count to reflect a realistic rate rather than 100/s.

tools/audiotap/Sources/AnchorSearch.swift
pasrom: `recordDelivery` overwrites, which means credit can erase a correct memory as well as record one.

The reasoning in `AnchorDeliveryCredit` for the 50% threshold computes roughly 82% zeros, but that is the mean across the 55 energetic windows of that recording, whereas the code tests a per-interval maximum. A mean of 18.5% non-zero is consistent with a distribution that includes one 5 s interval well above 50% non-zero. Around 51 s of non-zero samples exist in that 44.7-minute track, so it takes 2.5 s of it landing in one interval. A ringtone or a lobby loop rendered to the anchored device has that shape; a one-second join chime does not.

The direct consequence I first assumed does not follow, because a last-known-good device that is still the current default is deduplicated out of the candidate list. What does follow is this line: a burst on the dead anchor overwrites the UID of hardware that genuinely carried the meeting earlier, and the search loses that memory for the rest of the recording.

Requiring a bound rather than an average, for example several consecutive qualifying intervals before overwriting an existing known-good UID, would close it.

tools/audiotap/Tests/AppAudioCaptureAnchorRecoveryTests.swift
pasrom: Two notes on this file, since it is the only test that drives the real object.

**It pins the one-step behaviour.** This assertion says that after continued silence at the fallback the attempt count stays at 2. That is the current design, but it is also the behaviour I have flagged as the blocker, so if the search is changed to continue, this test changes with it.

**The cursor-to-device link is not covered.** `performAttempt()` returns `attemptBody()` whenever the seam is set, which bypasses `startCapture()` and therefore `resolveOutputAnchor()` entirely, and `Attempts.run()` supplies the anchor UID itself. I changed `resolveOutputAnchor()` to ignore the cursor and always return candidate 0, and the whole suite stayed green. Nothing under `Tests/` references that function. Since choosing a different device is the entire point of the change, that seam is worth closing, for example by injecting the device enumeration so `resolveOutputAnchor` can be driven with a synthetic device list.

For completeness, these edits also leave the suite green: `resetWindow()` as a no-op, deleting the `resetToDefault()` call, `defaultMaxTriggers` 3 to 1, `defaultRecoveryFractionThreshold` 0.50 to 0.89, deleting the delivery-credit reset at teardown, and hard-coding `AudioCaptureSession.appCaptureDigitallySilent` to `false`. The last one would silently disable the whole app-side feature.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
pasrom: This trailing import is what makes `./scripts/lint.sh` fail on the branch: SwiftFormat reports `blankLinesBetweenScopes` at line 201.

The import itself is needed, since `SystemSettingsPaths` lives in AudioTapLib and there is no re-export. Note that `lint.sh --fix` only inserts a blank line and leaves the import at the bottom, so moving it to the top of the file is a manual edit.

tools/audiotap/Sources/OutputDeviceEnumeration.swift
pasrom: Minor: `ids` is sized from the first call, and the second call can return a smaller `size` if a device disappears in between, leaving trailing `kAudioObjectUnknown` entries. They are harmless downstream, since `describeOutputDevice(0)` fails and is dropped by `compactMap`, but each one logs a "Cannot read output streams" warning.

```suggestion
        return Array(ids.prefix(Int(size) / MemoryLayout<AudioObjectID>.size))
```

tools/audiotap/Sources/SilentTapWatchdog.swift
pasrom: This comment says the zeros are "counted by `exactlyZeroSampleCount`", but that function has no callers: the per-sample counting that actually feeds this type is written out inline in `accumulateDebugRMS`.

`analyze` catches it, along with two more from this change. All three fail the gate as `unused_declaration`:

- `exactlyZeroSampleCount` in `RollingZeroFraction.swift`, never called.
- `restartsRequested` just below in this file, referenced only from tests.
- `getDefaultOutputDeviceTransportType()` in `Helpers.swift`, whose production caller this change replaces with `OutputDeviceEnumeration.transportLabel`, leaving it referenced only from tests and from a doc comment.

Calling the function from `accumulateDebugRMS` would fix the first one and make the comment true at the same time.

tools/audiotap/Sources/OutputDeviceAnchorPolicy.swift
pasrom: This sentence is the premise the whole mechanism rests on, and I could not reproduce it.

With a probe built to the same shape as production (private `stereoMixdownOfProcesses` tap, private aggregate, `tapAutoStart` true, anchor as main sub-device), on macOS 26.6.2 build 25G83, a process rendering to the system default was captured at full level with the aggregate anchored to a *different* device: -44.08 dBFS at the producer's own device and -44.08 dBFS at an unrelated USB interface. Anchoring to three different virtual devices also captured the audio; those rows show a non-zero "zero fraction" only because the aggregate adds the anchor device's own output channels alongside the tap's two, and the level drop matches the channel count to 0.01 dB.

I repeated the set with a producer rendering through VoiceProcessingIO, which is the path a meeting app uses for call audio, in case that behaved differently. It did not: -42.68 dBFS at both the producer's device and the unrelated USB anchor.

That is the opposite of "the tap carries what the tapped process renders to that device", and it also matches an earlier probe on the maintainer side that closed the same hypothesis. Your check 2 is the same shape as the virtual-anchor case and measured 100% zeros, so something differs between our setups that neither of us has isolated. Knowing the failing device's transport and channel count, and whether it was the system *default* as well as the anchor in the failing recordings, would help narrow it.

I am not asking you to take my measurement over yours. I am saying the causal claim is not established well enough to build the recovery on it, because if the anchor does not gate capture then moving the anchor is not what fixed your recordings, and the restart alone might be. A run of your check 2 with the cursor pinned at candidate 0 would separate those two directly.

tools/audiotap/Sources/OutputDeviceAnchorPolicy.swift
pasrom: Every physical device with output streams becomes a candidate here, including ones that cannot currently start.

Measured on the same probe: anchoring the aggregate to an HDMI display whose panel was asleep produced **zero buffers**, twice, rather than buffers of zeros. That matters specifically for this design, because `SilentTapWatchdog` needs samples to form a verdict. With no samples it can never trip again, the latch from the previous episode stays set, and the tap is parked on a dead anchor for the rest of the recording, with only the separate `noBuffers` notification to show for it.

An HDMI display, a powered-down interface or a disconnected dock all enumerate with output streams while being unable to clock an aggregate, so this is not a corner case. Two things would help: treat a candidate that delivers no buffers within a short deadline as a failed candidate and advance, and prefer candidates that report running I/O more strongly than the current stable partition does.

tools/audiotap/Sources/SilentTapWatchdog.swift
```diff
@@ -0,0 +1,172 @@
...
+    ///   - now: monotonic seconds. Injected rather than read here so the whole
+    ///     transition table is testable without waiting out a real window.
+    mutating func observe(
+        zeroSamples: Int, samples: Int, now: TimeInterval,
+    ) -> Action? {
+        guard samples > 0 else { return nil }
+        window.add(zeroSamples: zeroSamples, totalSamples: samples, now: now)
+        guard let fraction = window.zeroFraction else { return nil }
+
+        if isSilent {
```
pasrom: This is the blocker from the summary, at its source.

While `isSilent` is true, `observe` can only return `.recovered` or `nil`. The single clear is on line 146, behind `fraction <= recoveryFractionThreshold`. `resetWindow()` on line 169 keeps the latch by design, and `AnchorSearch.advance` has exactly one caller, reachable only from `.silenceDetected(mayRestart: true)`. So a fallback that is also dead ends the search for the episode.

Three things in this PR describe the behaviour that would result if it did not:

- `defaultMaxTriggers = 3` and the PR body's "at most three per recording": within one episode the budget cannot be spent at all.
- `OutputDeviceAnchorPolicy`'s "what `SilentTapWatchdog` walks, one step per trip".
- The "all `N` candidate anchors have delivered only digital silence" log line: for `N` greater than 1 it can only be reached after the fallback delivered real audio and later went quiet, at which point the message is not true.

The comment above ("a restart that did not fix the problem is not followed by an identical restart") is right about not re-trying the *same* anchor, but the code also prevents trying the *next* one.

Worth noting separately: the recovery branch has no span requirement, so immediately after `resetWindow()` a single buffer at 50% zeros or less clears the latch. The claim just above it that clearing "takes sustained audio rather than one loud buffer" only holds once the window has refilled.

tools/audiotap/Sources/AppAudioCapture+Restart.swift
pasrom: This is the only path that ever returns the cursor to 0, and it fires only on a real `kAudioHardwarePropertyDefaultOutputDevice` change.

Two consequences:

1. **No way back after a false positive.** If a healthy default is abandoned because the far end was legitimately silent for a full window, the tap stays on the fallback for the rest of the recording. `lastKnownGoodUID` cannot rescue it, since credit requires a 5 s interval at 50% zeros or less, which a zero-delivering anchor never produces. The user is not told to change the system output device either, because the message the rolling path selects is not the one that carries that advice.

2. **The latch and the budget are not reset here.** Your own report describes Teams re-pointing the default at its loopback driver four times in one meeting. Each of those periods spends one trigger, so the fourth is stranded with `mayRestart: false`. And if the previous fallback was dead, the latch is still set when the new default arrives, so the search is disarmed on a default it has never evaluated.

Resetting the watchdog's latch (and considering whether the budget should be per episode rather than per recording) alongside the cursor would address both.

app/MeetingTranscriber/Sources/ChannelHealthController+Alerts.swift
```diff
@@ -0,0 +1,203 @@
...
+        everCarriedSignal: Bool,
+        rollingDigitalSilence: Bool = false,
+    ) -> (title: String, body: String, urgency: NotificationUrgency) {
+        if fault == .gaveUp {
+            return gaveUpAlert(channel: channel)
+        }
+        let suppressible = fault == .digitalSilence && channel == .mic
+        return (
+            "Capture Channel Silent",
+            rollingDigitalSilence && channel == .app && fault == .digitalSilence
```
pasrom: Two things here.

**`everCarriedSignal` is discarded whenever the rolling verdict is set.** That makes the `(.app, .digitalSilence, false)` arm of `faultMessage` unreachable through this path. The case it costs: a tap denied the audio-capture grant delivers zeros from the first buffer, and if the user listens for more than two minutes before speaking, the rolling verdict claims the once-per-recording slot at t=120 with output-device advice, and the later, correctly corroborated age-based verdict is discarded. Previously that user got the Screen Recording and interception advice at first corroboration. The same happens permanently for anyone who sets `asymmetricSilenceWarningSeconds` above 120.

Note it is not a one-line pass-through: the `false` arm's text asserts "while the microphone carries audio", which is not true for the population the rolling path reaches, so that copy needs adjusting too.

**The urgency.** `.timeSensitive` for app-side digital silence is justified in the comment above by "the far end of a call does not go digitally silent while you are talking", which depends on microphone corroboration. The rolling path explicitly removes that corroboration. So the Focus-piercing alert now also fires for a joined-but-silent meeting, which is exactly the shape `applySilentRecording` calls out a few lines away when it chooses `.standard` because "a waiting room looks exactly like this". By this file's own test ("does this have a benign reading"), the uncorroborated rolling verdict looks like a `.standard` case.

app/MeetingTranscriber/Sources/ChannelHealthController.swift
pasrom: This guard suppresses the duplicate in only one order, and the default settings produce the other one.

`SilentRecordingMonitor`'s debounce is `asymmetricSilenceWarningSeconds`, default 90. The rolling window is 120. So at t=90 this method runs with `appDigitalSilenceActive` still false and posts "Recording Appears Silent"; at t=120 the rolling verdict posts "Capture Channel Silent" as time-sensitive. `NotificationManager.notify` uses a fresh UUID per call with no title or category de-duplication, and `ChannelFaultMonitor.reportedSilence` is not set for an uncorroborated verdict, so nothing else blocks the second one.

The two monitors are not different populations here: a listening user with a dead tap satisfies both, and the rolling verdict's stated target is precisely "a listening user may never speak".

`testDigitalSilenceIsReportedOncePerEpisode` sets the flag before tick 0, so it exercises only the order that works. A test that flips the flag *after* the debounce would show this.

The guard is also one-directional: the symmetric monitor defers to the rolling flag, but the rolling report never consults `recordingSilentActive`.

tools/audiotap/Sources/AppAudioCapture+Anchor.swift
pasrom: This is the only unconditional `.public` device UID in the module, and the file argues against it three lines above this helper.

The comment at line 46 says transport is logged unconditionally because it "is not personal data, unlike the device name/UID", inherited verbatim from the base. Line 52 follows that rule: it interpolates `currentDefault.uid` with no privacy annotation, so os.Logger redacts it. The three other UID logs in this module are all behind `[debug]` (`AppAudioCapture+Anchor.swift:60`, `AppAudioCapture+Restart.swift:34`, `MicCaptureHandler.swift:222`). This line is the one exception, and it runs with `debugLogging` off.

So the justification at lines 95 to 96, that the UIDs "are already logged on the line above", does not hold: the line above is redacted precisely because the file classifies UIDs as personal data.

It matters because `PersistentDiagnosticLog` streams `subsystem CONTAINS 'com.meetingtranscriber'` at info level into `~/Library/Logs/MeetingTranscriber/`, which is the file users export and attach to issues. It also repeats on every restart attempt, since `resolveOutputAnchor` runs per attempt, so a failing candidate writes it up to six times.

Either gate it behind `debugLogging` like the other three, or drop the annotation so it is redacted like line 52. The diagnostic value of knowing *how many* devices were running I/O survives either way.

tools/audiotap/Sources/RollingZeroFraction.swift
```diff
@@ -0,0 +1,77 @@
...
```
pasrom: `firstObservation` is set once and cleared only by `reset()`, while bin eviction is by wall-clock index. So `span` measures time since accumulation began, not the time actually covered by the retained bins, and the two diverge whenever buffers stop arriving without the tap being torn down.

After any gap of 121 s or more, every pre-gap bin is evicted, yet `span` still reports the full window and `SilentTapWatchdog` accepts it as complete. The sample floor does not catch it, because it is computed against the same inflated `span`: 120 x 100 = 12,000 samples is 0.125 s at 48 kHz stereo. So a verdict advertised as covering two minutes can rest on a fraction of a second.

I could not show a field path that reaches it. Sleep is ruled out, since `mach_absolute_time` does not advance while the machine is asleep. It needs an IOProc that stalls in place for minutes and then resumes, which I have not observed here. So this is a latent contract violation rather than a demonstrated bug, and the comment on lines 24 to 26 states the contract it breaks.

The fix is small: derive the span from the oldest retained bin, or require the retained sample count to reflect a realistic rate rather than 100/s.

tools/audiotap/Sources/AnchorSearch.swift
pasrom: `recordDelivery` overwrites, which means credit can erase a correct memory as well as record one.

The reasoning in `AnchorDeliveryCredit` for the 50% threshold computes roughly 82% zeros, but that is the mean across the 55 energetic windows of that recording, whereas the code tests a per-interval maximum. A mean of 18.5% non-zero is consistent with a distribution that includes one 5 s interval well above 50% non-zero. Around 51 s of non-zero samples exist in that 44.7-minute track, so it takes 2.5 s of it landing in one interval. A ringtone or a lobby loop rendered to the anchored device has that shape; a one-second join chime does not.

The direct consequence I first assumed does not follow, because a last-known-good device that is still the current default is deduplicated out of the candidate list. What does follow is this line: a burst on the dead anchor overwrites the UID of hardware that genuinely carried the meeting earlier, and the search loses that memory for the rest of the recording.

Requiring a bound rather than an average, for example several consecutive qualifying intervals before overwriting an existing known-good UID, would close it.

tools/audiotap/Tests/AppAudioCaptureAnchorRecoveryTests.swift
pasrom: Two notes on this file, since it is the only test that drives the real object.

**It pins the one-step behaviour.** This assertion says that after continued silence at the fallback the attempt count stays at 2. That is the current design, but it is also the behaviour I have flagged as the blocker, so if the search is changed to continue, this test changes with it.

**The cursor-to-device link is not covered.** `performAttempt()` returns `attemptBody()` whenever the seam is set, which bypasses `startCapture()` and therefore `resolveOutputAnchor()` entirely, and `Attempts.run()` supplies the anchor UID itself. I changed `resolveOutputAnchor()` to ignore the cursor and always return candidate 0, and the whole suite stayed green. Nothing under `Tests/` references that function. Since choosing a different device is the entire point of the change, that seam is worth closing, for example by injecting the device enumeration so `resolveOutputAnchor` can be driven with a synthetic device list.

For completeness, these edits also leave the suite green: `resetWindow()` as a no-op, deleting the `resetToDefault()` call, `defaultMaxTriggers` 3 to 1, `defaultRecoveryFractionThreshold` 0.50 to 0.89, deleting the delivery-credit reset at teardown, and hard-coding `AudioCaptureSession.appCaptureDigitallySilent` to `false`. The last one would silently disable the whole app-side feature.

tools/audiotap/Sources/OutputDeviceEnumeration.swift
```diff
@@ -0,0 +1,173 @@
...
```
pasrom: Minor: `ids` is sized from the first call, and the second call can return a smaller `size` if a device disappears in between, leaving trailing `kAudioObjectUnknown` entries. They are harmless downstream, since `describeOutputDevice(0)` fails and is dropped by `compactMap`, but each one logs a "Cannot read output streams" warning.

```suggestion
        return Array(ids.prefix(Int(size) / MemoryLayout<AudioObjectID>.size))
```

tools/audiotap/Sources/OutputDeviceAnchorPolicy.swift
pasrom: This sentence is the premise the whole mechanism rests on. My probe could not reproduce it; your second check and #699 did.

With a probe built to the same shape as production (private `stereoMixdownOfProcesses` tap, private aggregate, `tapAutoStart` true, anchor as main sub-device), on macOS 26.6.2 build 25G83, a synthetic producer was captured at full level whichever physical device the aggregate was anchored to, also when it rendered through VoiceProcessingIO. The virtual-anchor rows differ only by channel padding. Through the app's own capture path, with the default as the anchor and the producer on a different device, the same.

Your second check (default on Loopback, producer on the built-in speakers, 100% zeros) and the A/B in #699 (real Teams to a USB speakerphone, captured only when the speakerphone was the system default) both show the anchor deciding what the tap hears, so I no longer ask for the premise to be established before this moves.

One nuance: this sentence speaks of exact digital zeros, and #699 was a low floor rather than zeros; see the comment at the zero counting.

The most useful thing for closing the gap would be the producer in your second check: which API it used, and whether it named the built-in speakers explicitly or followed a device setting. It is the only synthetic run that reproduces the failure, so it is the best lead for what my probe is missing.

tools/audiotap/Sources/OutputDeviceAnchorPolicy.swift
pasrom: Every physical device with output streams becomes a candidate here, including ones that cannot currently start.

Measured on the same probe: anchoring the aggregate to an HDMI display whose panel was asleep produced **zero buffers**, twice, rather than buffers of zeros. That matters specifically for this design, because `SilentTapWatchdog` needs samples to form a verdict. With no samples it can never trip again, the latch from the previous episode stays set, and the tap is parked on a dead anchor for the rest of the recording, with only the separate `noBuffers` notification to show for it.

An HDMI display, a powered-down interface or a disconnected dock all enumerate with output streams while being unable to clock an aggregate, so this is not a corner case. Two things would help: treat a candidate that delivers no buffers within a short deadline as a failed candidate and advance, and prefer candidates that report running I/O more strongly than the current stable partition does.

tools/audiotap/Sources/AppAudioCapture+DebugLogging.swift
```diff
@@ -17,20 +17,27 @@ extension AppAudioCapture {
...
         let count = byteCount / MemoryLayout<Float>.size
         guard count > 0 else { return }
         let buf = UnsafeBufferPointer(
             start: data.assumingMemoryBound(to: Float.self), count: count,
         )
         var sumSq: Double = 0
+        var zeros = 0
         for sample in buf {
             sumSq += Double(sample) * Double(sample)
+            if sample == 0 { zeros += 1 }
```
pasrom: This counts samples that are exactly zero, and I think it will miss one field shape of the same failure.

In #699 the app track was not zeros. In a 5.6-minute Teams call every 5 s window read -98.5 dBFS with no program audio in it, while a parallel recorder had the call at normal level, and the 56-minute call from the same report shows the same flat floor in its per-minute sample counts. After the conversion to 16-bit almost every sample rounds to zero, but in the raw Float32 buffers counted here the samples need not be exactly zero, so whether the rolling fraction reaches 90% on a track like that is open. It is a related blind spot to the one that kept the existing `lastBufferHadEnergy` (`rms > 0`) from reporting that call.

I do not have a threshold to propose, because a floor like that and a genuinely quiet far end have to be told apart. But it would be worth knowing whether any of your failing recordings showed a low floor rather than exact zeros, since that decides whether counting zeros is enough.

PR #717, opened 2026-09-14
feat(app): add caption size presets for the live caption bar

semikashev:
## What

Adds a **Caption size** preset (Small / Medium / Large) to Settings → Transcription, under "Show caption overlay". Each preset pairs the caption font with the panel dimensions that fit it:

| Preset | Font | Panel |
|---|---|---|
| Small | 16 pt | 520 × 140 |
| Medium (default) | 22 pt | 720 × 200 |
| Large | 28 pt | 920 × 260 |

## Why

The caption bar ran at one hard-coded geometry (22 pt, 720 × 200). On a 14" display it covers a good part of the call window; on an external monitor it is on the small side. There was no way to change either the font or the panel without rebuilding.

## How

- `LiveCaptionsSize` couples font and panel size, because `LiveCaptionsWindowController` runs a fixed-size panel (auto-sizing produced the layout feedback loop described in its header). Changing one without the other would clip lines or leave dead space, so Settings only offers the pairs.
- **Medium reproduces the previous geometry and is the default**, so an upgrade changes nothing. An unknown stored value also reads as Medium.
- The preset is pushed through `LiveCaptionsState.setSize` by the window controller, which resizes the panel in the same step, so font and frame never disagree. The overlay is hosted in a panel with no settings object, which is why the view does not read `AppSettings` directly.
- A resize keeps the bar's bottom edge and horizontal centre (`resizedFrame`), and persists the new origin itself so a bar resized while hidden reappears at the same centre on the next `show()`.
- The picker is disabled whenever the overlay toggle is off or disabled, mirroring how that toggle follows the master switch.

## Tests

- `LiveCaptionsSizeTests`: Medium pins the historical metrics, Small pins the compact preset, presets grow monotonically, raw values are pinned, `resizedFrame` keeps bottom-centre.
- `LiveCaptionsSizeSettingTests`: round trip through `UserDefaults`, unknown stored value reads as Medium.
- `LiveCaptionsStateTests`: `size` defaults to Medium and follows `setSize`.
- `SettingsInteractionTests`: the picker writes back to settings and is disabled when the overlay is hidden.
- `AppSettingsTests.testDefaultValues`: default pinned.

Full `swift test --parallel` and `./scripts/lint.sh` pass locally.


app/MeetingTranscriber/Tests/LiveCaptionsSizeTests.swift
```diff
@@ -0,0 +1,53 @@
...
+        XCTAssertEqual(LiveCaptionsSize.medium.rawValue, "medium")
+        XCTAssertEqual(LiveCaptionsSize.large.rawValue, "large")
+        XCTAssertEqual(LiveCaptionsSize.allCases, [.small, .medium, .large])
+    }
+
+    /// Resizing keeps the bar where the user parked it: same bottom edge,
+    /// same horizontal centre. Keeping the bottom-left corner instead would
+    /// walk a shrinking bar sideways on every change.
+    @MainActor
+    func testResizedFrameKeepsBottomCentre() {
```
pasrom, line 75: This is the one change I would like before merging.

The invariant the whole design exists to protect, that the font and the panel can never be observed at different presets, currently has no test. `resizedFrame` is pure and well covered here, but nothing asserts that the controller actually pushes the preset into the state.

I checked how much that hides by applying four mutations to the production code at the same time:

- deleting both `state.setSize(size)` calls (in `init` and in `apply`)
- hard-coding the caption font back to a literal `22` in `LiveCaptionsOverlay`
- deleting the whole `.onChange(of: appState.settings.liveCaptionsSize)` block
- shrinking `.large` to 721x201 while leaving its font at 28 pt

The full suite, 3157 tests, stayed green. In production the first two would mean the panel resizes while the text keeps its old size, which is exactly the failure the doc comments say cannot happen.

Two assertions close it, and they need no panel and no `NSApplication`:

```swift
@MainActor
final class LiveCaptionsWindowControllerSizeTests: XCTestCase {
    func testInitPushesSizeIntoState() {
        let state = LiveCaptionsState()
        _ = LiveCaptionsWindowController(state: state, size: .large)
        XCTAssertEqual(state.size, .large)
    }

    func testApplyPushesSizeIntoState() {
        let state = LiveCaptionsState()
        let controller = LiveCaptionsWindowController(state: state, size: .large)
        controller.apply(size: .small)
        XCTAssertEqual(state.size, .small)
    }
}
```

I ran these in both directions: red against the mutated code (`("medium") is not equal to ("large")` and `("small")`), green against your code as it stands. `apply` returns at `guard let panel` while the panel is nil, so no window is involved. Feel free to put them wherever you think they fit best.

app/MeetingTranscriber/Sources/LiveCaptionsWindowController.swift
```diff
@@ -52,6 +54,34 @@ final class LiveCaptionsWindowController {
...
+        guard let panel else { return }
+        panel.setFrame(Self.resizedFrame(panel.frame, to: size), display: true)
+        persistOrigin(panel.frame.origin)
+    }
+
+    /// The frame a panel at `frame` takes when switched to `size`: same
+    /// bottom edge, same horizontal centre. Anchoring the bottom-left corner
+    /// instead would walk the bar sideways on every preset change, since the
+    /// user parks it by eye at the bottom-centre of a call window.
+    static func resizedFrame(_ frame: NSRect, to size: LiveCaptionsSize) -> NSRect {
```
pasrom: `resizedFrame` keeps the horizontal centre but never clamps to the screen, and AppKit does not constrain a borderless non-activating panel horizontally. I measured this on a panel configured exactly like `ensurePanel()`, across 36 runs covering visible, hidden and never-ordered-front states: `constrainFrameRect(_:to:)` returned the target rect unchanged every time, both immediately after `setFrame` and after `orderFrontRegardless()`.

So a bar parked flush against a side edge grows past it by half the width delta on a preset change: 100 pt for medium to large, 200 pt for small to large. The line below then persists that frame.

The two edges then behave differently, because `savedOrigin()` validates only the top-left corner:

- Right edge: the top-left corner is still on screen, so the overhang is accepted and restored on every later `show()`.
- Left edge: `x` is negative, the saved origin is rejected, and the bar relocates to the bottom-centre default at the next `show()`, which happens at the next recording start.

It is not as bad as it sounds: shrinking never overhangs, the effect is capped at 200 pt and does not accumulate, switching the preset back restores the flush position exactly, and enough of the bar stays on screen to option-drag it. So this is a polish issue rather than a lost bar. Clamping the resized frame into `panel.screen?.visibleFrame` (and, if you want the saved position to survive, validating both corners in `savedOrigin()`) would remove it.

app/MeetingTranscriber/Sources/LiveCaptionsWindowController.swift
```diff
@@ -52,6 +54,34 @@ final class LiveCaptionsWindowController {
...
+    /// step so neither can be observed at the other's old size; a bar that
+    /// already exists keeps its bottom edge and horizontal centre (see
+    /// `resizedFrame`). The origin is persisted here rather than left to the
+    /// move observer so a bar resized while hidden reappears at the same
+    /// centre on the next `show()`, which reads the saved origin back.
+    func apply(size: LiveCaptionsSize) {
+        guard size != self.size else { return }
+        self.size = size
+        state.setSize(size)
+        guard let panel else { return }
```
pasrom: Small correction to the doc comment above: the promise that a bar resized while hidden reappears at the same centre does not hold on the path where no panel exists yet.

This `guard` returns before `persistOrigin`, so the origin saved by an earlier session is kept. That origin is a bottom-LEFT corner stored for the old width, and `positionAtSavedOrDefault` reuses it as-is, so the centre shifts by half the width delta. Measured with a fresh controller and a saved origin of (396, 115): after `apply(.small)` the saved value is unchanged and the next `show()` puts the bar at x 396 with width 520, moving the centre 100 pt left.

This is easy to reach, since the controller is created at launch while `panel` stays nil until the first recording, so any preset change made in Settings beforehand takes this path.

It is a one-time offset, it does not accumulate, and it is correct in the case the comment was presumably written for (panel already exists: `apply` then persists the origin for the new width, and relaunch reproduces the centre). Persisting the centre instead of the left edge, or recording the width alongside the origin, would make the comment true on both paths. Rewording the comment would be fine too.

app/MeetingTranscriber/Tests/LiveCaptionsSizeTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    func testSmallIsTheCompactPreset() {
+        XCTAssertEqual(LiveCaptionsSize.small.fontSize, 16)
+        XCTAssertEqual(LiveCaptionsSize.small.panelSize, CGSize(width: 520, height: 140))
+    }
+
+    /// Every metric grows with the preset. A preset whose panel shrank while
+    /// its font grew would clip captions, so the ordering is pinned.
+    func testPresetsGrowMonotonically() {
```
pasrom, line 53: This test pins the ordering, but the doc comment above promises more than the assertions deliver: a preset "whose panel shrank while its font grew would clip captions". A `.large` of 28 pt inside a 721x201 panel passes this test (I ran it), because only strict ordering is checked, never any relation between font and panel.

`.large` is also the only preset that is not exact-pinned anywhere: `.medium` and `.small` have their own tests, and a grep of the suite finds `.large` only in this array, in the raw-value test and in `allCases`.

Either exact-pin `.large` the way you did for the other two, or encode the fit relation you actually mean (for example: height at least four rows at `fontSize` plus the overlay's 2 x 14 pt vertical padding). Then the comment and the assertion say the same thing.

app/MeetingTranscriber/Sources/LiveCaptionsOverlay.swift
```diff
@@ -43,7 +43,7 @@ struct LiveCaptionsOverlay: View {
         VStack(alignment: .leading, spacing: 4) {
             if let backend = state.activeBackend {
                 Text(backend)
-                    .font(.system(size: 11, weight: .semibold, design: .rounded))
+                    .font(.system(size: state.size.fontSize / 2, weight: .semibold, design: .rounded))
```
pasrom: Optional, and less severe than it first looks.

`fontSize / 2` gives 11 pt at medium, which reproduces the old literal exactly, and 14 pt at large. At small it gives 8 pt, which is below every AppKit reference size (caption2 is 10, mini control 9, small system 11).

I rendered it and measured rather than guessing: the 8 pt label is small but readable, and the contrast against the bar is 3.19:1, against 3.22:1 at medium. In other words what limits legibility here is the pre-existing 0.45 opacity, not the preset, and today's medium label is no better off. So this is not a regression your PR introduces.

If you want the small preset to be comfortable anyway, `max(10, fontSize / 2)` or simply keeping the label at a fixed 11 pt would do it, and neither changes medium.

app/MeetingTranscriber/Sources/LiveCaptionsSize.swift
```diff
@@ -0,0 +1,43 @@
...
+    var fontSize: CGFloat {
+        switch self {
+        case .small: 16
+        case .medium: 22
+        case .large: 28
+        }
+    }
+
+    /// Fixed panel dimensions: wide enough for a sentence at `fontSize`,
+    /// tall enough for four rows plus the overlay's padding. `.medium` is the
```
pasrom: Optional, and only about the word "tall enough".

I measured the overlay's real fitting height with the actual line heights (19 / 26 / 33 pt for the three presets, backend label 10 / 14 / 17, spacing 4, padding 28). Worst case with single-line rows, meaning backend label plus two finals plus two hypotheses: small 130 pt of 140, medium 162 of 200, large 193 of 260. So the comment holds as written.

With one wrapped row the picture separates: small needs 149 and exceeds its preset by 9 pt, medium 188 and large 226 still fit. Two wrapped rows put medium over as well.

Rows do wrap (`.fixedSize(horizontal: false, vertical: true)`), and the characters that fit per line are nearly identical across presets (about 57 / 61 / 63 for a German caption), so wrapping is roughly equally frequent at every preset. Small is simply the one with no headroom for it.

Something like 520x160 would give small the same one-wrap tolerance the other two have. Purely your call.

pasrom (commented):
Thanks for this, and for the unusually careful write-up. I checked the branch against `main`, ran the suite, and measured a few things on a real panel. CI is green on every lane.

What holds up: `.medium` really is byte-identical to the old geometry (720x200, 22 pt rows, 11 pt backend label), so an upgrade is a no-op; the unknown-stored-value fallback is tested; and the reasoning for persisting the origin inside `apply` is right for a reason worth recording: a `setFrame` that changes the size does not post `NSWindow.didMoveNotification` at all (0 of 36 measured runs), so the move observer could never catch it. The explicit `persistOrigin` is load-bearing.

One thing I would like before merge (the test note on `LiveCaptionsSizeTests`), plus smaller notes inline. The two suggestions on `LiveCaptionsOverlay` and `LiveCaptionsSize` are product judgement, so treat them as optional.

A note about our own documentation rather than your PR: `CLAUDE.md` claims that SwiftUI `Picker` and `Stepper` do not surface a ViewInspector-findable `.accessibilityIdentifier`, and that they must be located by label or document order. That is wrong under the ViewInspector version we pin, as your picker test demonstrates. I checked it across picker styles (automatic, menu, segmented, radioGroup, inline, palette, labelsHidden) and several stepper forms: all are findable by identifier and drivable, and removing the identifier turns the lookup red, so the identifier is genuinely doing the work. We will fix the rule. Your approach is the one we want going forward.

Optional and clearly convention rather than requirement: the sibling overlay-toggle change also exposed its key on `/state` (`AppSettings+RPC.swift`, `RPCStateSnapshot.swift`, and the exact-key allowlist in `RPCSettingsStateTests`). Leaving it out breaks no test, but adding it is three one-line edits and gives the automation API a way to observe the preset.

semikashev:
Thanks for the measurements — all of it addressed in the fixup commit (d02ffba); I'll squash once you're happy.

- **Controller pushes the preset into the state**: `LiveCaptionsWindowControllerSizeTests` with your two assertions, plus two for the no-panel path below. Verified red with both `setSize` calls deleted, green as is.
- **Clamp on resize**: `resizedFrame(_:to:within:)` now takes the screen's visible frame and pushes the result back inside it; `apply` passes `panel.screen?.visibleFrame`. `savedOrigin()` validates both top corners at the current size. Test: flush-left and flush-right bars stay inside a 1512-wide screen when going to `.large`.
- **No-panel path**: instead of rewording the comment, `apply` now re-centres the stored origin for the new width when no panel exists yet (a preset change in Settings before the first recording), so the promise holds on both paths. Test: saved x 396 at medium → 496 after `apply(.small)`, y untouched; nothing is written when nothing was saved. The comment also records your `didMoveNotification` finding, since that is what makes the explicit `persistOrigin` load-bearing.
- **`.large` pinned** exactly (28 pt, 920×260), and the fit relation is now a test: label + four rows at ~1.2× font, 4 pt spacing, 2×14 pt padding, plus one wrapped row of headroom, for every preset.
- **Small gets one-wrap headroom**: 520×160, per your measurement.
- **Backend label**: `labelFontSize = max(10, fontSize / 2)` — medium stays at 11, small goes to 10.
- **`/state`**: `recording.liveCaptionsSize` exposed (raw value), allowlist and both snapshot tests updated.

`UserDefaults` is injected into the controller (defaulting to `.standard`) so the origin tests never touch the real saved position.

pasrom (approved):
Thanks for the thorough follow-up. Every point is addressed, including the two I had marked optional, and putting `liveCaptionsSize` on `/state` was not even on the list. Making `UserDefaults` injectable so the tests never touch the real saved panel position is the right call and goes beyond what was asked.

Two notes on what I changed on the branch before merging:

- The `fixup!` commit is folded into the original one. This repository merges by rebase, so a `fixup!` subject would otherwise land on `main` verbatim. Your message and authorship are unchanged, and the folded tree is byte identical to what CI already validated.
- The branch is rebased onto current `main`, which branch protection requires. No conflicts, and the full suite passes on the rebased commit.

One thing I found in the new clamping is ours to fix rather than yours, so I am tracking it separately instead of asking you for another round. The clamp lands the bar with its right edge exactly on `visibleFrame.maxX`, and `CGRect.contains(CGPoint)` reports a point on that maximum edge as outside, so `savedOrigin()` then rejects the very position the clamp just produced and the bar falls back to the default placement on the next panel creation. Both halves arrived in the same commit and neither is wrong on its own. Fixing it properly needs a test seam that does not exist yet, since the screen list is not injectable the way you made `UserDefaults` injectable, so it earns its own change.

PR #719, opened 2026-09-15
feat(app): name recordings after calendar events

semikashev:
## What

Opt-in **calendar titles** (Settings → General → "Name recordings after calendar events"): a recording is named after the calendar event that is running when it starts, and the event's attendees are handed to the protocol as participants. Reads the calendars already on the Mac through EventKit, so a Google/Exchange/iCloud account added in System Settings works with no login of its own.

Before: `20260914_1240_arc_….md` with the H1 the LLM guessed. After: `20260914_1240_weekly_sync_….md`, protocol titled after the invite, attendees pre-filled for speaker naming.

## Why

Window titles are all the app has today, and for browser meetings there is no window title at all — the file is named after the browser ("Arc Call"). Nearly every meeting people record is in their calendar, and the calendar is where the title and the attendee list already live.

## How

- `WatchLoop.enqueueRecording` is the single funnel every trigger passes through (auto, manual, microphone-only, record-only sidecar), so `WatchLoop+CalendarTitle.swift` consults the lookup there, against the start date the recorder captured, rather than at detection. The extension keeps `WatchLoop.swift` under the file-length limit, the same way consent and manual recording are split out.
- `CalendarMeetingMatcher` is pure. The rule that matters: **the event must still be running when the recording starts** (early join allowed up to 15 min). With a looser "recently ended" rule an unscheduled 13:10 call was filed under a 12:00–13:00 weekly — measured on real recordings while developing this. Ties (an external invite plus one's own copy of the same meeting) are broken by a conference link that points at the app in use, then by having attendees, then by the closest start; the rest is logged as ambiguous.
- `CalendarMeetingLookup` protocol with `NoCalendarLookup` (default, injected in tests) and `EventKitMeetingLookup` (production). `isEnabled` is read per call so switching the setting off applies to the next recording. Birthday and subscribed (holiday) calendars and cancelled events are dropped in the adapter.
- A title the user typed in the app picker is never overruled (`calendarMayName`); the app-name default and the microphone-only title are. Teams participants read from the call win over the invite list, which only fills in when nothing was read.
- Off by default: switching it on asks for Calendar access from the toggle itself, and the section reports when macOS refused. `NSCalendarsFullAccessUsageDescription` added to Info.plist; `com.apple.security.personal-information.calendars` to the App Store entitlements.

## Tests

- `CalendarMeetingMatcherTests` (12): running event matches, early join, too-early and after-end rejections, all-day skipped, conference-link tie-break (native app and browser), ambiguity flag, attendees beat a personal block, closest start wins, blank titles skipped.
- `WatchLoopCalendarTitleTests` (5): detected meeting takes title + attendees and the lookup sees the recorder's start date and app name; no match keeps the window title; manual recording with the default title is named, a typed title is kept; microphone-only recording is named.
- `CalendarTitlesSettingTests`, `AppSettingsTests.testDefaultValues`, `SettingsInteractionTests.testCalendarTitlesToggleWritesBackToSettings`.

Full `swift test --parallel` and `./scripts/lint.sh` pass locally; README, `docs/architecture-macos.md` and `CLAUDE.md` updated.


app/MeetingTranscriber/Entitlements/AppStore.entitlements
```diff
@@ -6,6 +6,8 @@
     <true/>
     <key>com.apple.security.device.audio-input</key>
     <true/>
+    <key>com.apple.security.personal-information.calendars</key>
```
pasrom, line 9: This key also has to go into `Homebrew.entitlements`, otherwise the primary distribution channel cannot use the feature at all.

`scripts/build_release.sh` signs the app with `--options runtime`, and `scripts/lib/signing.sh` hands it `Homebrew.entitlements` as the base file. Under hardened runtime, calendar access is gated by `com.apple.security.personal-information.calendars` in exactly the way the microphone is gated by `com.apple.security.device.audio-input`, which is already in that file for this reason. Without the key `requestFullAccessToEvents()` resolves false with no TCC prompt, and the app never appears under System Settings > Privacy & Security > Calendars, so the red caption's advice cannot be followed either.

This will not reproduce during development: `scripts/run_app.sh` signs without `--options runtime`, so the restriction is not in force on the dev bundle. Worth confirming on a Developer ID signed, hardened runtime build.

app/MeetingTranscriber/Sources/CalendarMeetingMatcher.swift
```diff
@@ -0,0 +1,76 @@
...
+        let scored: [(CalendarEventCandidate, Int)] = events.compactMap { event in
+            let title = event.title.trimmingCharacters(in: .whitespacesAndNewlines)
+            guard !event.isAllDay, !title.isEmpty else { return nil }
+            guard event.start <= recordingStart.addingTimeInterval(earlyJoin),
+                  event.end >= recordingStart else { return nil }
+            var score = 0
+            let text = event.conferenceText.lowercased()
+            if domains.contains(where: { text.contains($0) }) { score += 100 }
+            if !event.attendees.isEmpty { score += 10 }
+            score -= Int(abs(event.start.timeIntervalSince(recordingStart)) / 60)
```
pasrom, line 66: This is the one I would fix first. The penalty is unbounded and `abs()` scores an event that has not started yet just as well as the one that is actually running, so a recording gets filed under the *following* meeting.

I compiled this file from your branch and ran it. Weekly 12:00-13:00 and Vendor call 13:00-14:00, both with a Zoom link and attendees, app Zoom:

```
rec 12:02 -> Weekly        (control, correct)
rec 12:44 -> Weekly
rec 12:45 -> Vendor call
rec 12:59 -> Vendor call
```

The threshold is exactly `earlyJoin`. The moment the next event enters the 15 minute window, its start is nearer than the running event's, so it wins outright. Any recording that starts in the last 15 minutes of a meeting that is followed by another meeting gets the wrong title *and* the wrong attendee list, which is the inversion of the rule the doc comment above sets out.

The additive shape also breaks the documented precedence more generally. With the scan window at +/-4h the penalty reaches 240 against a maximum bonus of 110, so a link-carrying event that started 100 minutes ago loses to any unlinked event that started more recently. Measured: an all-hands 09:00-12:00 with a link and attendees loses to an empty `Focus time` block 10:55-11:30 for a recording starting 11:00.

The comment describes a lexicographic order, so comparing tuples would give you what you meant: running events above early-join candidates first, then conference link, then attendees, then distance.

Worth knowing while you fix this: `testClosestStartWinsAmongOverlappingEvents` currently pins the behaviour that causes it, so it will need to change with the rule.

app/MeetingTranscriber/Sources/CalendarMeetingMatcher.swift
```diff
@@ -0,0 +1,76 @@
...
+        return ["zoom.us", "teams.microsoft.com", "teams.live.com", "webex.com", "meet.google.com", "whereby.com", "telemost.yandex"]
+    }
+
+    static func bestMatch(recordingStart: Date, appName: String, among events: [CalendarEventCandidate]) -> Match? {
+        let domains = conferenceDomains(forApp: appName)
+        let scored: [(CalendarEventCandidate, Int)] = events.compactMap { event in
+            let title = event.title.trimmingCharacters(in: .whitespacesAndNewlines)
+            guard !event.isAllDay, !title.isEmpty else { return nil }
+            guard event.start <= recordingStart.addingTimeInterval(earlyJoin),
+                  event.end >= recordingStart else { return nil }
```
pasrom, line 61: `>=` makes the boundary inclusive, so a recording that starts exactly on an event's end second is still treated as inside it. With back-to-back meetings the just-ended one then wins on score, which is the case the doc comment says it is designed against. `>` fixes it.

Minor on its own, but it is the same class as the ranking issue above and the two together make back-to-back meetings the weak spot of the feature.

app/MeetingTranscriber/Sources/CalendarMeetingMatcher.swift
```diff
@@ -0,0 +1,76 @@
...
+            var score = 0
+            let text = event.conferenceText.lowercased()
+            if domains.contains(where: { text.contains($0) }) { score += 100 }
+            if !event.attendees.isEmpty { score += 10 }
+            score -= Int(abs(event.start.timeIntervalSince(recordingStart)) / 60)
+            return (event, score)
+        }
+        guard let top = scored.max(by: { $0.1 < $1.1 }) else { return nil }
```
pasrom, line 69: Two things at this line.

**Ties are resolved by input order.** `max(by:)` keeps the first element among equal scores, and `events(matching:)` has no guaranteed ordering, so which of two equally scored events names the recording is not stable. Measured on your code: `[x, y]` returns "Partner sync", `[y, x]` returns "EXT: partner sync" for two otherwise identical candidates. You already detect the case as `isAmbiguous`, so a deterministic secondary sort (title, or calendar source) would make the outcome reproducible rather than only logged.

**There is no quality floor.** A single event with no attendees and no conference link is returned as the match, so a personal block renames an unrelated call. Measured: `Lunch` 12:00-13:00 with neither, a Zoom call at 12:20, result is `Lunch`. A timed (not all-day) multi-day event is the same shape: an `Offsite` from Monday 09:00 to Wednesday 17:00 names every call for three days and hands its two attendees to every protocol. Requiring either attendees or a link before a calendar event is allowed to overrule a real window title would close both.

app/MeetingTranscriber/Sources/CalendarMeetingMatcher.swift
```diff
@@ -0,0 +1,76 @@
...
+        let meeting: CalendarMeeting
+        /// Another event scored the same, so the choice was arbitrary. Logged,
+        /// not surfaced: the usual cause is one meeting present twice (an
+        /// external invite plus one's own copy), where either title is right.
+        let isAmbiguous: Bool
+    }
+
+    /// Conference domains a native meeting app implies. A browser implies
+    /// nothing in particular, so any conference link counts for it.
+    static func conferenceDomains(forApp appName: String) -> [String] {
```
pasrom, line 46: `contains` is unanchored, so `"TeamSpeak".lowercased().contains("teams")` is true and a TeamSpeak recording is told its conference domains are the Microsoft Teams ones. In a slot that holds both a Teams-linked invite and a Zoom-linked copy that picks the wrong one, where returning nil and falling through to the any-link list would have been right.

More structurally, this is a second table of which app is which. `AppMeetingPattern.forAppName(_:)` in `MeetingPatterns.swift` already answers that question and is used by `MicInputDetector` and `PowerAssertionDetector`. Adding a meeting app now means editing two files with nothing tying them together, so a `conferenceDomains` property on `AppMeetingPattern` would keep the knowledge in one place.

Three of the four native branches also have no test, so a typo in a domain string would silently disable the tie-break rather than fail anything.

app/MeetingTranscriber/Sources/CalendarMeetingLookup.swift
```diff
@@ -0,0 +1,96 @@
...
+            "Calendar event matched: \(match.meeting.title, privacy: .private)\(match.isAmbiguous ? " (another event shares the slot)" : "", privacy: .public)",
+        )
+        return match.meeting
+    }
+
+    /// Birthday and subscribed (holiday) calendars never describe a call, and
+    /// a cancelled event is not a meeting anybody attended.
+    private static func candidate(_ event: EKEvent) -> CalendarEventCandidate? {
+        guard event.status != .canceled,
+              event.calendar.type != .birthday, event.calendar.type != .subscription else { return nil }
```
pasrom, line 80: `event.calendar`, and `event.startDate` / `event.endDate` on lines 89-90, are `null_unspecified` in the EventKit headers, so they import as implicitly unwrapped optionals and a nil traps here. `event.title` got a `?? ""` on line 88, so the nullability was clearly on your mind for that one.

What makes this worth guarding rather than accepting: the crash lands inside `enqueueRecording`, which is after `recorder.stop()` has finished the WAV files and before `pipelineQueue.enqueue(job)`. So the failure mode is not an unnamed recording, it is a recording that never becomes a job at all and only resurfaces through orphan recovery on the next launch, if at all.

A `guard let` over the three of them, returning nil to skip the event, costs nothing.

app/MeetingTranscriber/Sources/CalendarMeetingLookup.swift
```diff
@@ -0,0 +1,96 @@
...
+        logger.info(
+            "Calendar event matched: \(match.meeting.title, privacy: .private)\(match.isAmbiguous ? " (another event shares the slot)" : "", privacy: .public)",
+        )
+        return match.meeting
+    }
+
+    /// Birthday and subscribed (holiday) calendars never describe a call, and
+    /// a cancelled event is not a meeting anybody attended.
+    private static func candidate(_ event: EKEvent) -> CalendarEventCandidate? {
+        guard event.status != .canceled,
```
pasrom, line 79: Only `.canceled` is filtered, so an invitation the user **declined** still names the recording. Declined events stay on the calendar in Google, Exchange and iCloud with status `.confirmed`.

Concretely: a declined 13:00-14:00 meeting with attendees and a Teams link scores well and is the only candidate for an unrelated ad-hoc call at 13:05, so the recording is named after the meeting the user refused, and that invite's attendee list is handed on as participants.

The loop just below already visits the `isCurrentUser` attendee, so reading its `participantStatus` and skipping `.declined` fits where the code already is.

app/MeetingTranscriber/Sources/CalendarMeetingLookup.swift
```diff
@@ -0,0 +1,96 @@
...
+              event.calendar.type != .birthday, event.calendar.type != .subscription else { return nil }
+        let attendees = (event.attendees ?? []).compactMap { attendee -> String? in
+            guard attendee.participantType == .person, !attendee.isCurrentUser else { return nil }
```
pasrom, line 82: `participantType == .person` drops attendees that EventKit reports as `.unknown`, which is what some Exchange and CalDAV servers yield when the ATTENDEE has no CUTYPE. For those events the attendee list comes back empty, so the event silently loses the attendee bonus in the matcher and the protocol gets no participants although the invite had them.

Accepting `.unknown` alongside `.person` and filtering only rooms, resources and groups would be closer to the intent.

app/MeetingTranscriber/Sources/CalendarMeetingLookup.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    func meeting(startingAt start: Date, appName: String) -> CalendarMeeting? {
+        guard isEnabled(), Self.hasAccess else { return nil }
+        let predicate = store.predicateForEvents(
+            withStart: start.addingTimeInterval(-Self.scanWindow),
+            end: start.addingTimeInterval(Self.scanWindow),
+            calendars: nil,
+        )
+        let candidates = store.events(matching: predicate).compactMap(Self.candidate)
```
pasrom, line 65: This is a synchronous EventKit query on the main actor. `WatchLoop` is `@MainActor`, so `enqueueRecording` blocks the main thread here at the end of every recording, over an 8 hour window across every account (`calendars: nil`), with no timeout. The `eventsMatchingPredicate:` header says explicitly to run it somewhere other than the main thread.

On an account with many shared or recurring calendars, or with a cold or wedged CalendarAgent, that is a visible freeze of the menu bar and the whole main actor exactly when the pipeline flips state. The comment on the protocol declaration ("Synchronous on purpose") explains the call site but not the main-thread cost.

Related, on the window itself: `predicateForEvents` matches events that *overlap* the window, not events that start inside it, so a long all-hands is returned even with a much smaller window. The +/-4h rationale in the comment on line 37 is a misreading of the predicate, and everything the extra span pulls in is mapped through `candidate(_:)`, including copying and lowercasing every invite's notes blob, before the guard throws it away.

app/MeetingTranscriber/Sources/CalendarMeetingLookup.swift
```diff
@@ -0,0 +1,96 @@
...
+/// only: `EKEventStore` with full read access is the same store Calendar.app
+/// shows, so a Google or Exchange account added in System Settings works
+/// without a login of its own.
+///
+/// `isEnabled` is read on every call rather than captured once, so switching
+/// the feature off in Settings takes effect for the next recording without
+/// rebuilding the loop.
+final class EventKitMeetingLookup: CalendarMeetingLookup {
+    private let isEnabled: () -> Bool
+    private let store = EKEventStore()
```
pasrom, line 32: An open question rather than a finding, and the one thing in this review I could not settle by reading.

This store is constructed when the `WatchLoop` is built, which is typically before calendar access exists, while `requestAccess()` on line 51 asks on a separate throwaway store. EventKit's header says a store created before access "will not contain any calendars" for that entity type and refreshes via a change notification. Whether this long-lived instance actually picks up the grant without being recreated is what I am unsure about. If it does not, the failure is quiet: `hasAccess` passes, the query returns nothing, and every recording for the rest of the session keeps its window title with only "No calendar event covers the recording start" in the log. Worth one check on a real machine: start watching, then enable the toggle and grant access, then record, without restarting.

Independent of that answer, this is also eager: `EventKitMeetingLookup` is constructed at both `WatchingController` sites, so every watch toggle and every manual recording start opens a CalendarAgent connection even when the feature is off, which is the default for every existing install. Making the store lazy, created on the first call where `isEnabled() && hasAccess` holds, would cover the cost and, if the answer above is the bad one, the staleness too.

app/MeetingTranscriber/Sources/Settings/GeneralSettingsView.swift
```diff
@@ -82,6 +84,40 @@ struct GeneralSettingsView: View {
...
+                .accessibilityIdentifier(A11yID.calendarTitlesToggle)
+                .onChange(of: settings.calendarTitlesEnabled) { _, enabled in
+                    guard enabled, !calendarAccessGranted else { return }
+                    Task { calendarAccessGranted = await EventKitMeetingLookup.requestAccess() }
+                }
+            Text(
+                """
+                Uses the event that is running when a recording starts for the file name and \
+                protocol title, and hands its attendees to the protocol. Calendars come from \
+                System Settings → Internet Accounts; nothing is sent anywhere.
```
pasrom, line 107: "nothing is sent anywhere" is not accurate, and this is the sentence a privacy-conscious user will decide on.

The title itself does stay local, both generators discard the parameter. The attendee names do not: they become `PipelineJob.participants`, go through `resolveSpeakerNames` into the auto-assigned speaker names, and from there into the labelled transcript that is passed to `generator.generate(transcript:...)`. With `protocolProvider == .openAICompatible` that is an outbound HTTP request to whatever host the user configured. Line 84 falls back to the mail address with `mailto:` stripped when an attendee has no display name, so colleagues' addresses can be what leaves the machine.

Suggest saying what is true instead, something like: calendars are read locally, and attendee names are used for speaker naming, so they reach the protocol provider like the rest of the transcript.

(Unrelated and minor, same line region: the `mailto:` strip on line 84 uses an unanchored `replacingOccurrences`, so it removes the substring anywhere in the value rather than only as a prefix.)

app/MeetingTranscriber/Sources/Settings/GeneralSettingsView.swift
```diff
@@ -82,6 +84,40 @@ struct GeneralSettingsView: View {
...
+    @State private var calendarAccessGranted = EventKitMeetingLookup.hasAccess
+
+    /// Opt-in because it triggers a permission prompt. The toggle asks for
+    /// access the moment it is switched on, so a user who sees "no access"
+    /// below knows the answer came from macOS, not from a step they missed.
+    private var calendarSection: some View {
+        Section("Calendar") {
+            Toggle("Name recordings after calendar events", isOn: $settings.calendarTitlesEnabled)
+                .accessibilityIdentifier(A11yID.calendarTitlesToggle)
+                .onChange(of: settings.calendarTitlesEnabled) { _, enabled in
```
pasrom, line 99: Two problems with the access state around this toggle.

The setting is persisted true before the request resolves and nothing resets it when the user denies, so `calendarTitlesEnabled` reads on forever while `hasAccess` stays false and every recording silently keeps its window title. The red caption is the only signal.

And the refresh does not work in the case the comment on lines 87-89 describes. A `@State` initialiser runs once and a re-render does not re-run it, and `.onAppear` does not fire again when the user grants access in System Settings while this window stays open on this tab, so the "not granted" warning stays up after the problem is fixed. The comment states a mechanism SwiftUI does not have, which is worth correcting even if the caption behaviour is accepted.

Also, nothing pins either arm: `Toggle.tap()` in ViewInspector flips the binding without firing `.onChange`, so mutating this closure to never request access keeps the new test green. Injecting the status read and the request as closures would make both sides assertable.

app/MeetingTranscriber/Sources/WatchLoop+CalendarTitle.swift
```diff
@@ -0,0 +1,36 @@
...
+    /// when there is nothing.
+    func calendarNamed(
+        title: String,
+        participants: [String],
+        appName: String,
+        recording: RecordingResult,
+        calendarMayName: Bool,
+    ) -> (title: String, participants: [String]) {
+        guard calendarMayName else { return (title, participants) }
+        let start = recording.recordingStartDate == .distantPast ? nowProvider() : recording.recordingStartDate
```
pasrom, line 29: When the recorder's start is the `.distantPast` sentinel this substitutes `nowProvider()`, but at enqueue time that is the *end* of the recording, not an approximation of its start. So an unknown start does not degrade to "no calendar answer", it degrades to a confident wrong one, matched against whatever is running when the recording stops. For a long recording that is precisely the "inherits the following meeting's name" failure the design is meant to prevent.

`DualSourceRecorder` cannot reach this today, since `stop()` requires the `start()` that sets the date, so this is about the fallback being wrong rather than currently firing. Returning `(title, participants)` unchanged for an unknown start is both simpler and correct, and the comment above would no longer need to claim the clock is a good anchor.

app/MeetingTranscriber/Sources/WatchLoop.swift
```diff
@@ -273,8 +277,12 @@ class WatchLoop {
         var failureMessage: String?
         do {
             let recording = try recorder.stop()
+            // The app picker passes the app name when the title field was left
+            // blank, and the microphone entry point has no field at all; a title
+            // the user typed is a decision the calendar must not overrule.
             enqueueRecording(
                 title: info.title, appName: info.appName, recording: recording, trigger: .manual,
+                calendarMayName: info.title == info.appName || info.title == ManualRecordingInfo.microphoneTitle,
```
pasrom, line 285: This reconstructs "did the user type a title?" by string comparison, from information that was already destroyed upstream: the app picker collapses an empty field into the app name, so the distinction is gone by the time it gets here and this guesses it back.

Two consequences. A user who records Chrome and types "Chrome" as the title has that decision overruled by the calendar, which the comment above says must not happen. And a later change to the picker's default silently turns calendar naming off for every picker recording with every test still green, because the test hard-codes the same equality.

The parameter also defaults to `true` on line 465, so any future call site opts into the override by omission. Carrying the provenance on the value instead, a `titleSource` on `ManualRecordingInfo` or a small enum of user-authored vs fallback, removes the guess and the defaulting question at once.

app/MeetingTranscriber/Sources/WatchLoop.swift
```diff
@@ -454,7 +462,13 @@ class WatchLoop {
         recording: RecordingResult,
         trigger: RecordingSidecar.Trigger,
         participants: [String] = [],
+        calendarMayName: Bool = true,
     ) {
+        let (title, participants) = calendarNamed(
```
pasrom, line 467: Worth a thought for the record-only path specifically. The sidecar's `title` and `participants` have so far meant "observed window title" and "participants read from the call". After this change they can also mean "from a calendar invite", with `RecordingSidecar.currentVersion` unchanged and no field saying which.

That matters for the fleet topology the mode exists for: a consumer that treats `participants` as people who were actually in the call now receives the full invite list of a large meeting for a recording where three people spoke, and cannot tell the two apart or key on the version to find out. Either a `titleSource` / `participantsSource` field with a version bump, or leaving the sidecar on observed values, would keep that contract honest.

app/MeetingTranscriber/Sources/WatchingController.swift
```diff
@@ -251,6 +251,7 @@ final class WatchingController {
                     },
                     notifier: notifier,
                     denyListStore: ConsentDenyListStore(settings: settings),
+                    calendarLookup: EventKitMeetingLookup { [settings] in settings.calendarTitlesEnabled },
```
pasrom, line 254: Nothing asserts this wiring. All the new tests inject `StubCalendarLookup`, and `WatchLoop.init` defaults to `NoCalendarLookup()`, so deleting `calendarLookup:` here or at the second construction site keeps the whole suite green while the feature does nothing in production. The same holds for the adapter itself: removing the `isEnabled()` gate, or the `.canceled` filter, or inverting `!attendee.isCurrentUser` all stay green, because `candidate(_:)` takes an `EKEvent` that only a real store can produce.

Two cheap steps would close most of it. Assert that the constructed loop's lookup is an `EventKitMeetingLookup` after a manual recording start, which prompts nothing since constructing the store does not touch TCC. And have `candidate(_:)` map through a plain input struct (title, isAllDay, status, calendar type, attendee tuples, url/location/notes) that the adapter fills and tests can construct directly.

This matters more than usual here because the fork means CI has run only the title check, so the suite is not even known to compile against these changes on anything but your machine.

app/MeetingTranscriber/Tests/WatchLoopCalendarTitleTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        DetectedMeeting(
+            pattern: .teams,
+            windowTitle: "Test Meeting | Microsoft Teams",
+            ownerName: "Microsoft Teams",
+            windowPID: 9999,
+        )
+    }
+
+    func testDetectedMeetingTakesTitleAndAttendeesFromTheCalendar() async throws {
+        let lookup = StubCalendarLookup(answer: CalendarMeeting(title: "Weekly sync", attendees: ["Anna", "Ben"]))
```
pasrom, line 62: Could you swap these attendee fixtures for anonymous placeholders, `Speaker A` / `Speaker B` style? Same on lines 70, 112 and 119. The convention in this repo is that no real first names go into fixtures, commit messages or PR text, since the history is published.

The PR description needs the same treatment and is the more urgent half, because it is already served publicly: the before/after example is built from two people's real first names. A generic example would make the same point.

pasrom (changes requested):
Thanks for this. The shape is right and it reads like it belongs in this codebase: consulting the calendar at enqueue rather than at detection so every trigger is named by one rule, keeping the matcher pure and testable, and splitting the extension out of `WatchLoop.swift` the way consent and manual recording already are. "The event must still be running" is the correct rule to build on, and the reasoning you give for rejecting the looser one is convincing.

I checked the "single funnel" claim rather than taking it: `enqueueRecording` is private with exactly two call sites, and both production `WatchLoop` constructions do pass the lookup, so it holds for live recordings. It does not hold for crash recovery, which builds its job directly and never consults the calendar, and that is the case where there is no window title and the calendar would be the only naming source available. Either wire recovery through the lookup using the timestamp the file stem already carries, or narrow the claim that this PR adds to `CLAUDE.md` and `docs/architecture-macos.md`.

Three things I would want resolved before this merges. The first two I verified by building against your branch rather than by reading, and the reproductions are in the inline comments.

**1. The Homebrew build can never obtain calendar access.** The entitlement went only into `AppStore.entitlements`. The Homebrew variant is signed with hardened runtime against `Homebrew.entitlements`, which does not carry the calendar key, so access is refused with no prompt. This does not show up in development because the dev bundle is signed without hardened runtime.

**2. The matcher picks the following meeting.** The distance term is unbounded and uses `abs()`, so an event that has not started yet competes on equal footing with the one that is running. Measured threshold: exactly the early-join window, so any recording starting in the last 15 minutes of a meeting followed by another one is filed under the wrong meeting, with the wrong attendee list.

**3. The Settings caption promises "nothing is sent anywhere", which is not accurate.** Attendee names, including raw mail addresses when there is no display name, reach the protocol provider through the transcript.

Below those, the one that is cheapest to fix relative to its cost is the implicit unwrapping of `event.calendar` / `startDate` / `endDate`, because a nil there crashes after the audio is finalised and before the job exists, so the recording is lost rather than merely unnamed.

Two notes that are not about the code.

Because this comes from a fork, the only check that has run is `conventional-title`. The test suite, lint and analyze have not run against these changes at all, so "passes locally" is currently the whole of the evidence. That also raises the bar on the test gap noted inline: every new test injects a stub, and the production wiring, the setting gate and the EventKit mapping are not covered by anything.

And a request: the PR description's before/after example is built from two people's real first names, and the new test fixtures use real first names as attendees. This repository publishes its history, so the convention here is anonymous placeholders. The description is the urgent half since it is already public.

One genuinely open question, noted inline on the store: whether an `EKEventStore` created before authorization returns events after the grant without being recreated. I could not settle that by reading the headers, and it needs a check on a real machine.

semikashev:

PR #730, opened 2026-09-17
fix(app): move the caption panel by hand, ⌥-drag stopped on macOS 27

semikashev:
## What

The caption bar can no longer be repositioned with ⌥-drag on macOS 27. The panel now moves itself from the mouse events it receives instead of relying on `isMovableByWindowBackground`.

## Why

On macOS 27, `isMovableByWindowBackground` no longer moves a window whose content view is an `NSHostingView`. Measured with six otherwise identical borderless panels shown side by side:

| Panel | Difference from the caption panel | Moves? |
|---|---|---|
| A | none (replica) | no |
| B | without `.nonactivatingPanel` | no |
| C | level `.floating` instead of `.statusBar` | no |
| D | without `.stationary` | no |
| E | plain `NSView` content instead of `NSHostingView` | **yes** |
| F | replica, moved by hand from mouse events | **yes** |

`NSHostingView.mouseDownCanMoveWindow` still reports `true` and `hitTest` returns the hosting view itself, so it is not the documented opt-out; the mouse-down, drag and up events all reach the panel (`sendEvent` sees them), only the automatic move never happens.

## How

- `LiveCaptionsPanel` (an `NSPanel` subclass) records where in the frame the mouse went down and keeps that point under the cursor on each drag event. The arithmetic lives in `PanelDragTracker`, a pure value type with tests.
- `isMovableByWindowBackground` is no longer set at all: with the manual move it would only double up on systems where it still works.
- The ⌥ modifier monitor is unchanged; it still only flips `ignoresMouseEvents`. A drag that started keeps following the mouse after ⌥ is released, as before — AppKit routes the rest of a drag to the mouse-down window regardless.

## Tests

`PanelDragTrackerTests`: origin follows the mouse keeping the grab offset; no move without a mouse-down; `end()` stops the drag. Verified by hand on macOS 27.0: the caption bar drags again.

Note: `SettingsInteractionTests` has 5 pre-existing failures on macOS 27 (ViewInspector cannot find the toggles), identical on untouched `main`; not related to this change.


semikashev:

PR #731, opened 2026-09-17
feat(app): toggle the caption overlay from the menu bar

semikashev:
## What

A **Show Captions** checkmark item in the menu bar (⌘L while the menu is open) that hides or brings back the live caption bar in one click. It flips the same setting as Settings → Transcription → "Show caption overlay", so the two never disagree, and live transcription keeps running either way.

## Why

The caption bar can already be hidden live, but mid-call that is a Settings window away, and the bar tends to be in the way exactly then.

## How

- The item only appears while live transcription is on: with it off there is no bar, and the item would only raise the question of what it does. `CaptionOverlayItem` folds the two settings into that three-way state so the menu view stays free of settings logic, the way its other inputs are plain values and callbacks.
- A `Toggle` in a menu renders as a checkmark item; its binding's setter ignores the value and calls back, the setting is the source of truth and the next render reads it back.

## Tests

`MenuBarViewTests`: the item reflects shown/hidden, tapping it calls back, it is absent when live transcription is off; `CaptionOverlayItem` derivation from the two settings.

Note: `SettingsInteractionTests` has 5 pre-existing failures on macOS 27 (ViewInspector cannot find the toggles), identical on untouched `main`; not related to this change.


semikashev:

PR #743, opened 2026-09-20
fix(app): stream long audio imports during resampling

dterenyi:
## What changed and why

Long stereo imports can exceed `AVAudioPCMBuffer` buffer capacity and terminate the app before Swift can catch the error. Ordinary imports now keep the established buffered decoder. When one buffer is unsafe, the app streams decoding and resampling through `AVAssetReader`, then uses the existing fallback chain if streaming fails.

The streaming path explicitly requests little endian Float32, preserves source channels until they are averaged in Swift, reads decoded audio through EOF, and rejects large duration shortfalls instead of concealing them with silence.

Reproduced with a 4:26:43, 48 kHz stereo AAC file containing 768,159,744 decoded frames.

## Verification

* [x] Rebased on current `main`, no merge commits
* [x] Repository pinned SwiftFormat and SwiftLint checks pass
* [x] 91 affected audio, fidelity, streaming, resampling, import, and FFmpeg tests pass with native arm64 FFmpeg
* [x] All five `AudioImportFidelityTests` added on `main` pass after the tiering change
* [x] Debug test build and optimized release build pass
* [x] Conventional Commit, one logical change with an explanatory commit body

An earlier unfiltered local bundle hit UI inspection and snapshot failures that reproduced on untouched `main`, then terminated in an existing model preload test. No affected audio test fails.


app/MeetingTranscriber/Sources/AudioMixer.swift
```diff
@@ -344,6 +344,16 @@ enum AudioMixer {
...
         }
+
+        do {
+            try await streamResampleFile(from: source, to: destination, targetRate: targetRate)
+            return
+        } catch {
+            logger.info(
+                "Streaming resample failed for \(source.lastPathComponent, privacy: .private): \(error.localizedDescription, privacy: .public), trying buffered fallbacks",
+            )
+        }
```
pasrom: I'll start here because most of the other comments follow from this one. With streaming as the first tier, every source whose rate isn't 16 kHz now takes a different decoder, resampler and channel fold, not only the long recordings this fixes. That includes every recording the app makes itself: the recorder runs at 48 kHz, so the byte-copy fast path never applies and `resampleFile` is the production path (`PipelineQueue+Stages.swift` lines 244, 245, 354, 504).

A shape that keeps the proven decoder on the ordinary cases and runs the new code for exactly the case it was written for, reusing the `AVAudioFile` handle the fast path already opens:

```swift
if let audioFile = try? AVAudioFile(forReading: source) {
    if Int(audioFile.processingFormat.sampleRate) == targetRate {
        try FileManager.default.copyItem(at: source, to: destination)
        return
    }
    if !canAllocateSinglePCMBuffer(
        frameCount: audioFile.length, format: audioFile.processingFormat
    ) {
        try await streamResampleFile(from: source, to: destination, targetRate: targetRate)
        return
    }
}
// existing AVAudioFile -> AVAsset -> ffmpeg chain
```

Also worth knowing: the `catch` only fires on a thrown error, so a tier that returns success with wrong audio, or blocks, never hands over to the fallbacks. Several of the comments below are that case in practice.

app/MeetingTranscriber/Sources/AudioMixer+Streaming.swift
```diff
@@ -0,0 +1,157 @@
...
+        let output = AVAssetReaderTrackOutput(
+            track: audioTrack,
+            outputSettings: [
+                AVFormatIDKey: kAudioFormatLinearPCM,
+                AVLinearPCMIsFloatKey: true,
+                AVLinearPCMBitDepthKey: 32,
+                AVLinearPCMIsNonInterleaved: false,
+                AVNumberOfChannelsKey: 1,
+                AVSampleRateKey: targetRate,
+            ],
```
pasrom, line 44: Small one with a big effect: `AVLinearPCMIsBigEndianKey` isn't set, so the reader inherits the source's endianness. For AIFF/AIFC and big-endian CAF it emits big-endian float32, and `writeDecodedAudio` copies those bytes as native floats. On a clean stereo AIFF with a 0.5 amplitude sine, the output peaks at 1.0000 with 16,094 of 32,000 samples clipped and an RMS of -3.0 dBFS, where the current path gives the correct -9.0 dBFS. The reader reports success, so the fallback chain never gets a look in. `.aiff` is in the import picker via `AudioImportTypes.nativeTypes`, so it is reachable from the UI.

Nothing in our fixture set could have shown you this: every audio fixture in the repo is little-endian. One line closes it:

```swift
AVLinearPCMIsBigEndianKey: false,
```

app/MeetingTranscriber/Sources/AudioMixer+Streaming.swift
```diff
@@ -0,0 +1,157 @@
...
+import CoreMedia
+import Foundation
+
+extension AudioMixer {
+    /// Decode and resample incrementally so long imports never need one buffer for
+    /// the entire recording. `AVAudioPCMBuffer` stores byte capacity as `UInt32`;
+    /// a four hour 48 kHz stereo recording exceeds that even though its frame count
+    /// still fits in `AVAudioFrameCount`.
+    static func streamResampleFile(from source: URL, to destination: URL, targetRate: Int) async throws {
+        let asset = AVURLAsset(url: source)
```
pasrom: Without `AVURLAssetPreferPreciseDurationAndTimingKey`, `track.timeRange.duration` is only a bitrate estimate for MP3 without a Xing header and for ADTS AAC, and the clamp on line 108 then trims the decoded audio to that estimate. On the repo's own speech material re-encoded as VBR MP3 without Xing, 720,576 frames are written where the content really holds 796,288: 4.82 seconds (9.7 percent) drop off the end of the recording, silently, with success reported. For a meeting recording that is usually the wrap-up.

```swift
let asset = AVURLAsset(
    url: source,
    options: [AVURLAssetPreferPreciseDurationAndTimingKey: true],
)
```

app/MeetingTranscriber/Sources/AudioMixer+Streaming.swift
```diff
@@ -0,0 +1,157 @@
...
+        let expectedFrameCount = expectedFrameCount(for: timeRange.duration, targetRate: targetRate)
+        let reader = try AVAssetReader(asset: asset)
+        let output = AVAssetReaderTrackOutput(
+            track: audioTrack,
+            outputSettings: [
+                AVFormatIDKey: kAudioFormatLinearPCM,
+                AVLinearPCMIsFloatKey: true,
+                AVLinearPCMBitDepthKey: 32,
+                AVLinearPCMIsNonInterleaved: false,
+                AVNumberOfChannelsKey: 1,
```
pasrom: Two things on `AVNumberOfChannelsKey: 1`. Asking the reader to fold to mono gives a power-preserving fold of about 0.707 per channel, where `readSamplesFromAudioFile` averages. That is a consistent +3.01 dB against today's output, and it clips material the averaging fold leaves alone: a 0.9 amplitude correlated stereo source came out with 13,598 of 32,000 samples at full scale.

Separately, a source tagged `kAudioChannelLayoutTag_DiscreteInOrder` decodes to digital silence (RMS -200 dB) with status `completed` and no error. The mic-pair tags (MidSide, XY, StereoHeadphones, Binaural) and 5.1 all fold correctly, so Discrete is the only affected layout, but it is a silent success.

Neither could show up in the suite: no fixture carries a channel layout, and our only stereo file never goes through this path. Suggestion: request the source's channel count in `outputSettings` and average the channels in Swift, matching what `readSamplesFromAudioFile` already does.

app/MeetingTranscriber/Sources/AudioMixer+Streaming.swift
```diff
@@ -0,0 +1,157 @@
...
+        if reader.status == .failed {
+            throw AudioMixerError.audioExtractionFailed(reader.error?.localizedDescription ?? "Unknown error")
+        }
+        if let expectedFrameCount, writtenFrameCount < expectedFrameCount {
+            try appendSilence(
+                frameCount: expectedFrameCount - writtenFrameCount,
+                format: format,
+                to: destinationFile,
+            )
+        }
```
pasrom: The top-up to the container's declared duration is unconditional, with no threshold, so a decode that stops early without setting `status == .failed` yields a file of the right length that is mostly silence, and nothing downstream can tell it from a correct import. It's not hypothetical: on the repo's `two_speakers_de.opus` the output carries exactly 2,886 trailing zero frames, 180 ms of silence that isn't in the source (true length 5.000 s). Harmless at that size, but a decoder that stops partway through a long recording would pad the rest with silence and report success.

I'd keep the padding, since a few frames of drift is normal, but log written versus expected and treat a large shortfall as a failure, so the buffered fallback still gets its turn:

```swift
logger.info("Streaming resample wrote \(writtenFrameCount) of \(expectedFrameCount) frames")
guard expectedFrameCount - writtenFrameCount <= shortfallTolerance else {
    throw AudioMixerError.audioExtractionFailed(
        "decoded \(writtenFrameCount) of \(expectedFrameCount) frames",
    )
}
```

app/MeetingTranscriber/Sources/AudioMixer+Streaming.swift
```diff
@@ -0,0 +1,157 @@
...
+        to file: AVAudioFile,
+    ) throws -> Int64 {
+        var writtenFrameCount: Int64 = 0
+        while let sampleBuffer = output.copyNextSampleBuffer() {
+            guard let blockBuffer = CMSampleBufferGetDataBuffer(sampleBuffer) else { continue }
+            let byteCount = CMBlockBufferGetDataLength(blockBuffer)
+            let availableFrameCount = byteCount / MemoryLayout<Float>.size
+            let remainingFrameCount = expectedFrameCount.map { max($0 - writtenFrameCount, 0) }
+            let frameCount = min(availableFrameCount, remainingFrameCount.map(Int.init) ?? availableFrameCount)
+            guard frameCount > 0 else { break }
```
pasrom: Two small things on these lines.

The `min(...)` clamp truncates whenever the decoder yields more frames than the container's duration implies, which is the mechanism behind the MP3 loss above. Padding up to the declared length is right, since a container can under-deliver, but truncating down isn't: when the decoder has real samples, it knows better than the header. Even with the precise-timing fix in place, not cutting real samples here is worth having.

`frameCount` also reaches zero for two different reasons: every expected frame has been written, where `break` is correct, or this one sample buffer carried no bytes, where `break` abandons the rest of the file. I didn't see the second case in around 45 real files, so this is hardening rather than a known defect, but separating them is cheap. Roughly, with `remainingFrameCount` being `nil` when the duration is unknown:

```swift
if let remainingFrameCount, remainingFrameCount == 0 { break }
guard availableFrameCount > 0 else { continue }
```

pasrom (commented):
Thank you for this one. The crash is real, the diagnosis is right, and the detail in your report (4:26:43, 48 kHz stereo, 768,159,744 frames) made it straightforward to reproduce and confirm.

`canAllocateSinglePCMBuffer` guards exactly the right thing: the limit really is the total across the channel buffers, which I checked independently, and a pre-allocation check really is the only option, because the overflow terminates the process rather than throwing. One small extra for the guard: the allocator rounds each buffer up to 16 bytes before multiplying, so three frame counts (536,870,909 to 536,870,911 for non-interleaved stereo float32) still pass the check and abort. Rounding the per-buffer byte count up to 16 before the multiply closes that.

What I'm asking for is a rework of the tiering (the comment on `AudioMixer.swift`), not a rejection. Once streaming runs only when the buffered path cannot allocate, the remaining comments become small: mostly one-line changes to `outputSettings` and the asset options.

I also want to be plain about where the fault for the rest lies: with the project, not with you. Every audio fixture in the repo is mono, little-endian and carries no channel layout, and the only stereo file is the one that never goes through this code path. There was no test here you could have run that would have shown any of this. CI hasn't run on this PR either, since fork workflows need maintainer approval, so automation caught nothing as well.

We have since added `AudioImportFidelityTests` on `main` (#746): five guards covering exactly these properties. After a rebase they will go red and show each item concretely. We owed you those; they should have existed before you ever opened this PR. Happy to help with the rebase or anything else, and looking forward to landing this.

pasrom (changes requested):
Marking this as changes requested so the PR status reflects where we are. The message hasn't changed: the detail is in the comments above, and the one thing I'm actually asking for is the tiering rework on `AudioMixer.swift`, so that `streamResampleFile` runs when the buffered path cannot allocate rather than ahead of it. The rest are mostly one-line changes to `outputSettings` and the asset options, and several of them stop mattering once the tiering is right.

To repeat the part that matters most: none of this was findable with the tests this project gave you, and CI never ran on the PR either. The guards on `main` (#746) exist now so the next person gets told by a red test instead of by a review.

Happy to help with the rebase, and glad to land this once the tiering is sorted.

dterenyi:
Thanks for the detailed review. I rebased onto current `main` and addressed the requested tiering in `72357ca`: ordinary imports stay on the established buffered path, and streaming runs only when the allocator check says one buffer is unsafe.

I also rounded each buffer size to 16 bytes before the capacity calculation, requested little endian Float32, preserved source channels until averaging in Swift, enabled precise timing, read through EOF without trimming, and limited silence padding so a large shortfall reaches the existing fallback chain.

The five new `AudioImportFidelityTests` now pass, along with five direct streaming tests. In total, 91 focused tests pass. The pinned formatting and lint checks, debug test build, and optimized release build also pass.

pasrom (approved):
Every point is addressed, and I checked each one against the new streaming code rather than taking it on trust: big-endian AIFF now comes out at -9.0 dBFS with peak 0.5000 and nothing clipped, the stereo fold peaks at 0.6000 which is the arithmetic average exactly, a DiscreteInOrder source decodes instead of going silent, and both the VBR MP3 without Xing and the Opus fixture land within a millisecond of the true length.

One thing I did not expect: the Ogg Vorbis fixture used to block forever inside `copyNextSampleBuffer()`, and it now decodes in 0.20 s. A clean A/B shows that was the missing precise-timing option all along, so your fix for the truncation removed a hang as a side effect. Worth knowing.

Thank you as well for the direct streaming tests. That was the right instinct: once streaming is only the fallback, the guards on `main` stop reaching it, so tests that call `streamResampleFile` directly are the only thing covering that path. The boundary pair on the rounding gap and the asymmetric 0.8/0.4 channels are exactly right.

PR #748, opened 2026-09-20
test(app): stop test defaults suites leaving a plist behind on every run

pasrom:
Every test that needs isolated settings makes a `UserDefaults(suiteName:)`, and the file behind it stays in `~/Library/Preferences` forever. On this machine that had reached 117,903 plists, 113,345 of them named after test classes, taking 500 MB because each 42-byte shell occupies a 4 KB block.

## Why the obvious fix is not the fix

Most of these tests already call `removePersistentDomain(forName:)`. It empties the domain and leaves the file. Deleting the file in the teardown does not help either, and that is the part worth reading before simplifying `DefaultsSuite`, because three shorter versions were tried and measured first.

`cfprefsd` keeps the emptied domain and writes it back out about five seconds later:

| checked at | plain | + `CFPreferencesAppSynchronize` | + `removeSuite` |
| --- | --- | --- | --- |
| immediately | gone | gone | gone |
| 2 s | gone | gone | gone |
| 5 s and beyond | back | back | back |

An `atexit` unlink does not help either, because the rewrite lands roughly five seconds after the writing process has already exited. There is no moment inside a process's life at which the file can be deleted and stay deleted. Relocating `HOME` is also closed: `FileManager.homeDirectoryForCurrentUser` ignores the variable and `cfprefsd` writes to the real location regardless.

That timing is also why this was never visible from a targeted run, and it is the trap to know about: a `swift test --filter OneClass` process is gone long before the rewrite, so a check made right after it exits reads clean and means nothing. Only a full run shows the leak, and then on a varying fraction of the suites created, because each one is a race. Instrumenting the teardown showed it plainly: 888 cleanup calls, and 230 of 238 leftover files had that call made for their own name.

## What this does instead

The cleanup is deferred to a later run, which is what `WatchingControllerFactory` already did for its own files and the reason its prefix was the one that had not piled up: 20 files against 34,863 for a test class with no sweep.

`DefaultsSuite.remove` empties the domain now, because that is what isolates the next test's keys, and records the name in a manifest keyed by pid. A later run unlinks exactly those names. A manifest is only acted on once its process is gone **and** it has sat untouched for a minute: sweeping on a dead pid alone deletes the fresh files of sibling `--parallel` workers before the rewrite lands, and throws away the manifest that was the only record of them. That mistake was measured too, at 404 files across two runs with nearly every manifest consumed inside the run that wrote it.

Names are recorded rather than matched by pattern, and the preferences directory is never listed: it holds a six-figure number of entries on a machine that has run this suite for a while, and the factory sweep's own comment records paying 16 s across the suite for listing it per call.

`ScratchDefaultsHygieneTests` keeps it that way by scanning the test sources, counting suites against cleanups per file rather than asking whether the helper appears somewhere in it. The weaker check let a real gap through, a second suite in `DebugRPCServerIntegrationTests` that nothing cleaned up, which is fixed here as well.

## Measured effect

Two consecutive full runs, counting the directory before and after each:

```
LAUF 1  exit=0  3182 tests  new=+172  removed=-11
LAUF 2  exit=0  3182 tests  new=+172  removed=-184
```

The second run removes the first run's files plus some older backlog, so the total fell rather than growing by 172. Growth is bounded at roughly one run's worth instead of accumulating.

**Not measured:** a third run to confirm that this is the steady state rather than a one-off backlog drain. I stopped short of it deliberately; a reviewer should weigh the claim accordingly.

Separately visible during review: restoring the factory sweep's immediate unlink took its prefix from 79 files back down to 35 in a single run.

## Not included

The existing backlog stays on disk. There is no measured functional impact, the directory holds real preferences next to the test residue, and the sweep will never touch the old files because they have no manifest. If it is ever wanted, it should be a separate step that prints the candidate list for review first, never a pattern.

## Gates

`./scripts/lint.sh` clean over 587 files. `swift test --parallel` green, 3269 tests, exit 0. The guard was proven red against a control file carrying one cleaned and one uncleaned suite, and green without it.


pasrom:
Follow-up after a quality pass over the branch. Two things are worth recording here rather than only in the history.

**What the pass changed.** The dead-pid predicate was a verbatim copy of the one in `XCTestCase+WatchingController`, comment included, and both of them gate deleting files out from under a live `--parallel` sibling; there is now one. The preferences path was spelled in two files that have to agree, and is now `DefaultsSuite.preferencesDirectory`. The once-per-process flag plus lock was a hand-rolled `swift_once` and is now a `static let`. The unlink reports its own outcome instead of a second `stat` per name. And the factory sweep's comment now says what is actually true of it: its own suite goes through the manifest now, so the directory listing only drains files created before this change, which makes it deletable rather than something to maintain.

**A defect this pass introduced and how it was caught,** because the failure mode is instructive. Narrowing the lock, I removed it entirely instead of leaving it around the manifest append, where `seekToEnd` and `write` are two steps on a shared handle reached from teardown blocks on arbitrary threads. The suite then failed with eight to eleven `Timed out while preparing execution worker` errors and not one failed assertion, which reads exactly like a loaded machine. It was not: eight earlier runs the same day had zero such errors, the second failing run had them with the load average down from 17.76 to 5.09, and after restoring the lock the suite came back green with zero of them at load 315.91. A green run under twenty times the load of a red one is what rules the environment out.

Gates on the pushed head: `./scripts/lint.sh` clean over 587 files, `swift test --parallel` exit 0 with 3269 tests and no worker errors.

**Still not measured,** unchanged from the description: the third consecutive run that would show the bounded growth is a steady state rather than a one-off backlog drain.

**One reviewable suggestion deliberately not taken.** The test target already has `makeTempDirectory(prefix:)`, which creates a resource and registers its teardown in one call precisely so a caller cannot forget the second half. This change adds only the cleanup half, so 41 call sites still mint their own name and register cleanup separately, in four spellings and five naming conventions. A `makeScratchDefaults()` on `XCTestCase` would collapse those, drop a `String!` property and its lint suppression from 14 classes, and make most of the guard test unnecessary by making the wrong thing unrepresentable rather than policed. It touches the same 31 files again, so it is left as a follow-up rather than folded in here.

